PR #10, opened 2020-01-10
actually return an error when data.zip is missing

chrboe:
More than two and a half years after messaging @flibitijibibo on Twitter, begging him for the source code, I'm very happy to finally be able to officially lay my eyes on this work of art (well, sort of).
Anyways, I'm also excited to share my first contribution to the game that has cost me countless hours of my life.

## Legal Stuff:

## Changes:

We should return an error code when we can't find data.zip, just letting
the program crash is a little crude.

Note: The message box did not work for me, so I just swapped it for some `puts`s. I'm obviously open for suggestions here.

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -32,7 +32,7 @@ void PLATFORM_getOSDirectory(char* output);
 void PLATFORM_migrateSaveData(char* output);
 void PLATFORM_copyFile(const char *oldLocation, const char *newLocation);
 
-void FILESYSTEM_init(char *argvZero)
+int FILESYSTEM_init(char *argvZero)
 {
 	char output[MAX_PATH];
 	int mkdirResult;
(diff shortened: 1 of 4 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -78,15 +78,13 @@ void FILESYSTEM_init(char *argvZero)
...
 	{
-		SDL_ShowSimpleMessageBox(
-			SDL_MESSAGEBOX_ERROR,
-			"data.zip missing!",
-			"You do not have data.zip!"
-			"\n\nGrab it from your purchased copy of the game,"
-			"\nor get it from the free Make and Play Edition.",
-			NULL
-		);
+		puts("Error: data.zip missing!");
```
flibitijibibo: Let's do the `puts` lines first, then the messagebox afterward (SDL_messagebox is written to be safe anywhere in the program, but for Linux it may require some X ugliness to successfully load).

Other than that, looks good to me! If you want to be the one to write `desktop_version/CONTRIBUTORS.txt` with your name in it, that's fine as well.

chrboe: Sounds like a plan

flibitijibibo:
Beautiful, thanks! (Also grats on being the first public contributor!)

PR #17, opened 2020-01-10
Introduce `seconds()` utility

chrboe:
## Legal Stuff:

## Changes:

See commit messages.

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -257,67 +257,67 @@ void entityclass::generateswnwave( Game& game, UtilityClass& help, int t )
                 game.swnstate = 1;
                 game.swndelay = 5;
 
-                if (game.swntimer <= 150)   //less than 5 seconds
+                if (game.swntimer <= seconds(5))
                 {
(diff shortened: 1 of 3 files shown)


flibitijibibo:
For the most part this is a no-op, ignoring the added overhead from the function, so this doesn’t seem necessary. Plus, this is only partially true because of the (buggy) FPS cap, which is a whole other can of worms... this would need to be fixed alongside a “real” framerate fix throughout the whole source.

Thanks anyway though!

chrboe:
No hurt feelings, these are just basically RFCs anyways.

However, could you elaborate on the FPS cap point? Surely if the framerate is not consistent this would also imply that the current numbers are only partially correct, right? I don't see how adding a wrapper around it would change that fact

flibitijibibo:
> Surely if the framerate is not consistent this would also imply that the current numbers are only partially correct, right?

Bingo. The timestep logic is super wrong and thus the frame times in the game logic are technically _only_ frame counts, not measurements of time. So technically, for the moment, the code is right and the comments are wrong!

PR #21, opened 2020-01-10
Remove hardcoded macOS SDK path

ePirat:
## Legal Stuff:

## Changes:

The macOS SDK was hardcoded in the CMake file, which is not needed as CMake will automatically pick an installed SDK, per their documentation.
Hardcoding it makes it quite troublesome to build the project easily and it is absolutely fine to use a SDK newer than the deployment target. Some care must be taken to not use APIs newer than the deployment target, therefore I've added the partial availability error which will error out if this is the case so that its easily noticeable.

For release builds that are to be distributed, it might still be a good idea to use the exact same SDK version as the deployment target when building 3rd party libraries as those might not correctly deal with macOS SDK versions and use functions newer than the deployment target else.
This can still be easily done by calling CMake like this so:

```
cmake -DCMAKE_OSX_SYSROOT=/path/to/sdk ..
```

Another possibly controversial change I made was to remove the not so friendly comments about Apple, as IMO its not really professional to have them and not particularly useful especially as some of these complain about things not really related to Apple but more about decisions taken by 3rd party dependencies like PhysFS decision to write some part in Objective-C even though it seems they could just have used plain C using CoreFoundation for that.
But if it's preferred to keep those, I can remove that change from the PR.


flibitijibibo:
Generic builds probably won’t care too much about the SDK, but for official builds I get very picky about the versions since I have frequently tried building old games with new SDKs only to have the SDK arbitrarily break the code at compile, link, and even _run_ time for some of my games. It seems extreme, but that’s just what building for macOS is like when you’re in the business of preserving games.

Also, I’m keeping my Apple comments. If a trillion-dollar corporation’s feelings are hurt they can let me know directly. (Agree that the PhysFS file should be C though.)

ePirat:
@flibitijibibo Ok, to clarify are you ok with any of the other changes made? (If yes, which? So I can update the PR to remove the stuff you don't want to change)

If none, feel free to close.

flibitijibibo:
The -Werror=partial-availability line is good, but sadly I have to keep the SDK line, at least for now. When I move to osx-cross I may re-evaluate that line since it’d be incorrect at that point... (Either way, thanks for looking at it!)

ePirat:
@flibitijibibo Would you accept a CMake option for the user to override the SDK path? Currently it seems impossible to build without editing the CMake file. (Or is there a way I am overlooking?)

flibitijibibo:
I'd be okay with that, but I thought CMake already allowed for that already (or am I just thinking of editing with ccmake?).

ePirat:
@flibitijibibo I thought so too but as CMAKE_OSX_SYSROOT is one of these magic CMake variables it appears that your set always overrides it and its impossible to check if the user had set it on the command line and only override conditionally…

The only working solution I found is to do this:

```diff
-       SET(CMAKE_OSX_SYSROOT /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk)
+       set(CUSTOM_OSX_SDK_PATH
+               /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk
+               CACHE PATH "Override the macOS SDK path")
+       SET(CMAKE_OSX_SYSROOT ${CUSTOM_OSX_SDK_PATH})
```

Maybe someone more familiar with CMake has a more elegant idea to do this?

flibitijibibo:
Possible workaround for now:

```
IF(IS_DIRECTORY /my/long/sdk/path/apple/why/)
    SET(CMAKE_OSX_SYSROOT /blah/blah/blah)
ELSE()
    "warn"("10.9 SDK not found, expect badness!");
ENDIF()
```

flibitijibibo:
New fix looks good, thanks (sorry for all the trouble!).

ePirat:
@flibitijibibo Note that ideally we would not hardcode the path but use `xcrun --sdk macosx10.14 --show-sdk-path` that way it will work even if Xcode.app has a different name. But it seems CMake has no support for that?

flibitijibibo:
You might be able to run a shell command directly `` style, but this might break attempts to build osxcross anyway. Will have to check when migrating over to that!

PR #22, opened 2020-01-10
Add vsync option

tim241:
## Legal Stuff:

## Changes:

I had screen tearing on linux, so I added a vsync option (using [SDL_GL_SetSwapInterval](https://wiki.libsdl.org/SDL_GL_SetSwapInterval)) which seems to work nicely


flibitijibibo:
This actually won’t work, both because GL_SetSwapInterval won’t work for all renderers and because the timestep breaks when vblank interrupts the frame. Before VSync can be integrated, someone will need to fix the update/render loop to spin updates when the game falls behind, rather than the really bad sleep() -> update() -> sleep() behavior.

Agree that vsync should be integrated after that, though!

PR #19, opened 2020-01-10
Update PhysFS APIs

ePirat:
## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV (for
  example, a 2.3 update on Steam for Windows/macOS/Linux)
- [x] I will be credited in a `CONTRIBUTORS` file for all of said releases, but
  will NOT be compensated for these changes

## Changes:

Some deprecated PhysFS APIs were used, which were trivial to update to the
non-deprecated versions.

The `PHYSFS_getUserDir` is still used for Linux as I could not test the change there
yet but looking at the implementation of the function it should work to replace it too
at a later point.

PR #23, opened 2020-01-10
FS: Use PHYSFS_getPrefDir on Linux/FreeBSD too

akien-mga:
## Legal Stuff:

## Changes:

Follow-up to #19 which did the change for macOS.
It appears to work as expected on Linux too.

Tested on a distro where XDG_DATA_HOME is undefined by default,
and `PHYSFS_getPrefDir` also resolves to `~/.local/share/VVVVVV/`.
The first organization parameter is unused on Linux and macOS.

Added myself to `CONTRIBUTORS` as requested though this patch is trivial enough not to require attribute IMO - but I'll contribute some more hopefully to make up for it.

I suggest ordering contributors alphabetically to make it easier for new ones to know where to add themselves - alternatively the instructions could be to add oneself at the bottom of the list. If so I can amend this commit to change that.

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -1,7 +1,10 @@
 Contributors
 ------------
 
+(Ordered alphabetically by last name.)
+
 * Christoph Böhmwalder (@chrboe)
+* Marvin Scholz (@ePirat)
 * Elijah Stone
 * Emmanuel Vadot (@evadot)
-* Marvin Scholz (@ePirat)
+* Rémi Verschelde (@akien-mga)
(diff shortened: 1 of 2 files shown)


desktop_version/CONTRIBUTORS.txt
```diff
@@ -1,7 +1,10 @@
 Contributors
 ------------
 
+(Ordered alphabetically by last name.)
```
akien-mga: Actually I see now that prior to this commit contributors were ordered alphabetically by first name -- maybe that's better and we should keep it this way?

flibitijibibo:
I was organizing by first name, but last name’s fine too!

PR #26, opened 2020-01-11
Enable and fix warnings

akien-mga:
## Legal Stuff:

## Changes:

- CMake: Add ENABLE_WARNINGS and ENABLE_WERROR build options

ENABLE_WARNINGS is enabled by default, with `-Wall` on GCC/Clang and
`/W4` on MSVC.

ENABLE_WERROR is disabled by default, treats warnings as errors.

- Fix warnings raised by GCC 8

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -3,6 +3,10 @@
 
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
 
+# CMake Options
+OPTION(ENABLE_WARNINGS "Enable compilation warnings" ON)
+OPTION(ENABLE_WERROR "Treat compilation warnings as errors" OFF)
+
 # Architecture Flags
 IF(APPLE)
 	# Wow, Apple is a huge jerk these days huh?
@@ -101,6 +105,17 @@ IF(WIN32)
 	ADD_EXECUTABLE(vvvvvv WIN32 ${VVV_SRC})
 ELSE()
 	ADD_EXECUTABLE(vvvvvv ${VVV_SRC})
+ENDIF()
+
+# Build options
+IF(ENABLE_WARNINGS)
+	# The weird syntax is due to CMake generator expressions.
+	# Saves quite a few lines and boilerplate at the price of readability.
+	TARGET_COMPILE_OPTIONS(vvvvvv PRIVATE
+		$<$<OR:$<CXX_COMPILER_ID:GNU>,$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+			-Wall $<$<BOOL:ENABLE_WERROR>:-Werror>>
+		$<$<CXX_COMPILER_ID:MSVC>:
+			/W4 $<$<BOOL:ENABLE_WERROR>:/WX>>)
 ENDIF()
 
 # Library information
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -108,7 +108,7 @@ bool binaryBlob::unPackBinary(const char* name)
 	}
 	PHYSFS_close(handle);
 
-	printf("The complete reloaded file size: %li\n", size);
+	printf("The complete reloaded file size: %lli\n", size);
 
 	for (int i = 0; i < 128; i += 1)
 	{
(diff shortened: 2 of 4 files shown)


desktop_version/CMakeLists.txt
```diff
@@ -101,6 +105,17 @@ ELSE()
...
 
+# Build options
+IF(ENABLE_WARNINGS)
+	# The weird syntax is due to CMake generator expressions.
+	# Saves quite a few lines and boilerplate at the price of readability.
+	TARGET_COMPILE_OPTIONS(vvvvvv PRIVATE
+		$<$<OR:$<CXX_COMPILER_ID:GNU>,$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+			-Wall $<$<BOOL:ENABLE_WERROR>:-Werror>>
+		$<$<CXX_COMPILER_ID:MSVC>:
+			/W4 $<$<BOOL:ENABLE_WERROR>:/WX>>)
```
akien-mga: This weird magic comes from https://foonathan.net/2018/10/cmake-warnings/

It confirms my feeling that CMake has one of the most unreadable syntaxes, but at least it seems to work well :)

Only tested on Linux / GCC 8 though.

desktop_version/src/BinaryBlob.cpp
```diff
@@ -107,7 +107,7 @@ bool binaryBlob::unPackBinary(const char* name)
 	}
 	PHYSFS_close(handle);
 
-	printf("The complete reloaded file size: %li\n", size);
+	printf("The complete reloaded file size: %li\n", (long int)size);
```
flibitijibibo: This might warn on type size issues... should we change the printf type instead?

desktop_version/CMakeLists.txt
```diff
@@ -101,6 +105,17 @@ ELSE()
 	ADD_EXECUTABLE(vvvvvv ${VVV_SRC})
 ENDIF()
 
+# Build options
+IF(ENABLE_WARNINGS)
+	# The weird syntax is due to CMake generator expressions.
+	# Saves quite a few lines and boilerplate at the price of readability.
+	TARGET_COMPILE_OPTIONS(vvvvvv PRIVATE
```
flibitijibibo: Whew, weird is right... but if this works on CentOS 7's GCC/CMake then whatever.

desktop_version/src/BinaryBlob.cpp
akien-mga: Right, I'll change it to `%lli`.

akien-mga: Good point, I changed it to `%lli`.

desktop_version/CMakeLists.txt
akien-mga: That would be worth double checking indeed. I can test on Ubuntu 14.04 if you want :)

flibitijibibo: Ha, yeah that'd be a good test! If it works there I'll merge.

flibitijibibo:
Just realized that we should probably use `%zu` for the size_t, my mistake... I dunno why I didn't pick up on what that variable was.

akien-mga:
> Just realized that we should probably use `%zu` for the size_t, my mistake... I dunno why I didn't pick up on what that variable was.

`PHYSFS_sint64` seems to always be signed 64-bit: https://icculus.org/physfs/docs/html/physfs_8h.html#a88c5f1d7b088888c729b22644651c9f5

```
[ 36%] Building CXX object CMakeFiles/vvvvvv.dir/src/BinaryBlob.cpp.o
/home/akien/tmp/VVVVVV/desktop_version/src/BinaryBlob.cpp: In member function ‘bool binaryBlob::unPackBinary(const char*)’:
/home/akien/tmp/VVVVVV/desktop_version/src/BinaryBlob.cpp:110:9: error: format ‘%zu’ expects argument of type ‘size_t’, but argument 2 has type ‘PHYSFS_sint64’ {aka ‘long long int’} [-Werror=format=]
  printf("The complete reloaded file size: %zu\n", size);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~
```

flibitijibibo:
What I get for not reading it more carefully... nevermind!

desktop_version/CMakeLists.txt
akien-mga: Finally managed to test it in Ubuntu 14.04 i386. It only provides SDL2 2.0.2 and CMake 2.8, so I had to hack a bit for proper detection of the libraries, and tweak the code to let it build even though it uses one function from later SDL versions. Seems to build without error.

Issue #16, opened 2020-01-10
[CMake] Linking to .lib files


ThatWeirdAndrew:
I was going through the build instructions and I was confused by the cmake command. The instructions state to run `-DSDL2_LIBRARIES="../../SDL2-2.0.10/lib/x86/SDL2;../../SDL2-2.0.10/lib/x86/SDL2main;../../SDL2_mixer-2.0.4/lib/x86/SDL2_mixer"`. Aren't you supposed to link directly to the libraries like this: `-DSDL2_LIBRARIES="../../SDL2-2.0.10/lib/x86/SDL2.lib;../../SDL2-2.0.10/lib/x86/SDL2main.lib;../../SDL2_mixer-2.0.4/lib/x86/SDL2_mixer.lib"`. If you don't link to the libraries, the build fails with the error `SDL2.obj not found`.

flibitijibibo:
That's what I thought too, but apparently CMake does the file extensions for you - when I tried to use .lib in my original Windows build it went "SDL.lib.lib". Thanks, CMake...

ThatWeirdAndrew:
Huh, for some reason it didn't do that for me.

flibitijibibo:
Yeah, I’m not sure what the rules are - seems like it does this to be like `-l` on other compilers. If we can find documented behavior for this it should be possible to confirm/fix our docs.

notviri:
Definitely needs the .lib suffix on my end added to the path.

flibitijibibo:
This is going to sound nuts, but... does it work specifically for VS2010 and the “10 2010” generator?

flibitijibibo:
Quick output from the flibitWindows builder... did this with Git bash, other possible factor(!?):

```
Ethan Lee@WIN-1Q73UBN293E MINGW64 ~/Programming/VVVVVV/desktop_version/flibitBuild (master)
$ cmake -G "Visual Studio 10 2010" ..  -DSDL2_INCLUDE_DIRS="C:\Users\Ethan Lee\Programming\SDL2-2.0.10\include;C:\Users\Ethan Lee\Programming\SDL2_mixer-2.0.4\include" -DSDL2_LIBRARIES="../../SDL2-2.0.10/lib/x86/SDL2;../../SDL2-2.0.10/lib/x86/SDL2main;../../SDL2_mixer-2.0.4/lib/x86/SDL2_mixer"
-- The C compiler identification is MSVC 16.0.40219.1
[...]

PR #30, opened 2020-01-11
Add CI

ePirat:
## Legal Stuff:

## Changes:

Add very basic GitHub Actions based CI that checks if the code builds.
Currently Ubuntu 18.04 only but can be expanded to more OSes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,66 @@
+name: CI
+
+# Trigger this workflow on push or pull request
+on: [push, pull_request]
+
+env:
+  SRC_DIR_PATH: desktop_version
+
+jobs:
+  build:
+    name: Build
+
+    runs-on: ${{ matrix.os }}
+
+    env:
+      CXXFLAGS: -I/usr/local/include/SDL2
+      LDFLAGS: -L/usr/local/lib
+
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-latest]
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Install dependencies (Linux)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ninja-build libsdl2-dev libsdl2-mixer-dev
+
+    - name: Install dependencies (macOS)
+      if: startsWith(matrix.os, 'macos')
+      run: brew install ninja sdl2 sdl2_mixer
+
+    - name: CMake configure
+      run: |
+        mkdir ${SRC_DIR_PATH}/build && cd ${SRC_DIR_PATH}/build
+        cmake -GNinja ..
+    - name: Build
+      run: ninja -C ${SRC_DIR_PATH}/build
+
+  build-win:
+    name: Build (windows-latest)
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Install dependencies (Windows)
+      run: |
+        vcpkg install sdl2 sdl2-mixer
+
+    - name: CMake configure
+      run: |
+        cd $env:SRC_DIR_PATH
+        $env:LDFLAGS =  "/LIBPATH:$env:VCPKG_INSTALLATION_ROOT\installed\x86-windows\lib "
+        $env:LDFLAGS += "/LIBPATH:$env:VCPKG_INSTALLATION_ROOT\installed\x86-windows\lib\manual-link"
+        cmake -G "Visual Studio 16 2019" -A Win32 `
+              -DSDL2_INCLUDE_DIRS="$env:VCPKG_INSTALLATION_ROOT\installed\x86-windows\include\SDL2" `
+              -DSDL2_LIBRARIES="SDL2;SDL2main;SDL2_mixer" .
+    - name: Build
+      run: |
+        cd $env:SRC_DIR_PATH
+        cmake --build .


flibitijibibo:
Dang, should have put a bet on when this would happen!

But, I suppose this would be good for community contributions... does this seem like a sane checklist?

- [ ] Visual Studio x86
- [ ] macOS Clang (targeting 10.9, doesn't have to be the right SDK I guess...)
- [ ] CentOS 7

I'd also be good with Ubuntu 14.04 over CentOS, but something tells me Cent7 is still a bit more supported... the main thing is that I want to run builds on a system with glibc 2.19 or older.

ePirat:
@flibitijibibo VS on Windows should be easy, macOS too, CentOS or old Ubuntu a bit trickier (because Github Actions has no machines for that, oldest it has is Ubuntu 16.04) but possibly doable by using Docker images.

flibitijibibo:
Going to CC @cybik on this... we're actually dealing with a similar issue for our ANGLE prebuilds, so it seems like a good idea to collaborate on this even though the projects are different.

cybik:
I'll be blunt: anything older than Ubuntu 16.04 should be considered "out of support" as far as user targets are concerned. By and large, kernels and core libraries on user systems should at the very _*least*_ be from 2016 onwards.

I am also fully aware that I just painted a target on my back.

flibitijibibo:
Dang, very well... 16.04 it is I guess. I did check and 14.04 went EOL last year, somehow didn't notice that. (OT: Why are we not using that for ANGLE-Builds then?)

cybik:
In the interest of full disclosure: there IS a way to chain a build inside a docker container, in GHA. It involves pulling ubuntu and then telling GHA to use a subcontainer through it.

https://stackoverflow.com/a/58944205/1920739

(For ANGLE: because I'm lazy and went for the obvious, why else)

charliebruce:
https://github.com/charliebruce/VVVVVV/tree/ci-build - I've been working on this too ;)

cybik:
If I may recommend, for Linux: all required libraries should be custom-built with the required configuration for VVVVVV in mind, so that all libraries can be bundled (license permitting) in a deliverable package. This way, one can "skirt" the reliance on system-provided payloads, and can "customise" the deliverables for self-contained DRM-Free packages and GameStore-tuned packages that may (at some point) provide a "known" base runtime (looking at you, @kisak-valve ;D)

That's how I do it on two specific game build configs I maintain myself.

flibitijibibo:
I think we’re just using the Actions as a test - official builds will still be on flibitBuild :P

ePirat:
@cybik As @flibitijibibo already said, the intention of this is just to verify the commits here do not break the build, not at all that binaries built by the CI should be distributed / uploaded anywhere, as that requires proper packaging which is far more complex. (i.e. Something build all deps ourselves with exactly the right configuration to not pull in license-incompatible deps, ensure no system-wide installed libraries were accidentally used, ensure for macOS that install_name of the libraries is fine for creating an .app bundle… For Linux mostly the same things, when using self-contained bundle approaches like AppImage or Snapcraft. So I believe this is not worth the hassle, for people who just want the game they should just buy it in the stores and not download a binary from Github?)

ePirat:
Added a build matrix to build on macOS and Ubuntu. Tried adding Ubuntu 16.04 but it fails due to some obscure CMake error…

ePirat:
Update the PR, mentioned @charliebruce as Co-Author as I took most of the Windows stuff from his PR, as discusses there.

There were some issues when using the CMake SDL2 targets instead of the variable, that I solved too which was needed to make the Windows builds work.
(In @charliebruce PR it used the variables instead, so these changes were not needed, but IMO using the targets is a nicer approach here as they are anyway provided so why not make use of them)

This covers now Linux (Ubuntu 18.04), Windows and macOS.
Ubuntu 16.04 is still missing but maybe we can merge this and someone add that later?

desktop_version/CMakeLists.txt
```diff
@@ -145,9 +145,10 @@ if (DEFINED SDL2_INCLUDE_DIRS AND DEFINED SDL2_LIBRARIES)
 else()
 	# Only try to autodetect if both SDL2 variables aren't explicitly set
 	find_package(SDL2 CONFIG)
+	find_package(sdl2-mixer CONFIG)
```
flibitijibibo: Not a huge deal, but is there a reason it’s this and not SDL2_mixer? (Also, does CMake care about case sensitivity at all here? Never actually checked that before)

ePirat: This was what the vcpkg package manager printed out as the way to find this 😅 
I am actually not sure what the "correct" thing here to put is, this seems at least to work fine for the vcpkg package but maybe it varies on other platforms…

flibitijibibo: Looks like it’s a known bug in vcpkg:

https://github.com/Microsoft/vcpkg/issues/4933

I’ll test this on my setups real quick...

flibitijibibo: This breaks on my CentOS and macOS builds - not too surprising since stock SDL_mixer doesn't have any CMake files. I don't know if I want to depend on vcpkg for this; is it possible to define SDL2_INCLUDE_DIRS and SDL2_LIBRARIES on Windows, even if we use vcpkg to install everything? That's technically what Windows users will do anyway, so _technically_ it'd be a more realistic test... (maybe it'll light up #16 somehow)

desktop_version/CMakeLists.txt
```diff
@@ -145,9 +145,10 @@ if (DEFINED SDL2_INCLUDE_DIRS AND DEFINED SDL2_LIBRARIES)
...
 	if (TARGET SDL2::SDL2)
 		message(STATUS "Using TARGET SDL2::SDL2")
-		target_link_libraries(VVVVVV SDL2::SDL2 SDL2_mixer)
+		target_link_libraries(VVVVVV SDL2::SDL2 SDL2::SDL2main SDL2::SDL2_mixer)
```
flibitijibibo: Because of the above issue, the SDL2_mixer change here will probably have to be reverted, but the SDL2main part is definitely a bug, so good catch there.

ePirat: Do you know if thats always required or vcpkg specific too?

ePirat:
Update the PR, removed the CMake changes, those are no longer needed as I set the variables now and do not rely on the brittle CMake magic

flibitijibibo:
This all looks fine now, we can figure out 16.04 later... @charliebruce Anything else?

charliebruce:
Looks good to me :+1:

PR #32, opened 2020-01-11
Menu fixes and a typo

ddm999:
## Legal Stuff:

## Changes:

Fixes some UI things in the menu.
- Unlock Play Modes > Return goes up one menu to Game Options, instead of up two menus to the main menu
- Game Pad Options > Return plays a menu sound instead of being silent
- When disabling Screen Effects, a menu sound is played instead of being silent (enabling Screen Effects is unchanged)
- Changing the following Graphics Options > Fullscreen, Scaling (Letterbox), Filter plays a menu sound before the change, and not both before and after the change
  - as these changes are instant or near-instant, these would just overlap and make an excessively loud noise

Also fixes a typo from Victoria in the Secret Lab ("scentists" -> "scientists").

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -428,7 +428,6 @@ SDL_assert(0 && "Remove open level dir");
                   if (game.currentmenuoption == 0){
                     music.playef(11, 10);
                     dwgfx.screenbuffer->toggleFullScreen();
-                    music.playef(11, 10);
                     game.fullscreen = !game.fullscreen;
                     updategraphicsmode(game, dwgfx);
                     game.savestats(map, dwgfx);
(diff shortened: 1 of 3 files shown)


desktop_version/src/Input.cpp
```diff
@@ -1332,6 +1331,7 @@ SDL_assert(0 && "Remove open level dir");
 
 					if (game.currentmenuoption == 4)
 					{
+                        music.playef(11, 10);
```
flibitijibibo: Weird tab here, other than that LGTM!

ddm999: Most of this file is spaces, but a few places are tabs: this is an area that uses tabs (which I didn't notice). Fixed now.

flibitijibibo:
Perfect, thanks!

PR #33, opened 2020-01-11
CI build

charliebruce:
## Legal Stuff:

## Changes:

Introduces builds using Github Actions for macOS, Windows (x86) and Linux. Binaries for each platform are also uploaded as build artifacts.

So far, I've only tested the produced binaries on macOS 10.15 (to the main menu).

~~Note that this also brings in https://github.com/tcbrindle/sdl2-cmake-scripts as a submodule. It's licensed as 2-clause BSD, so could pretty easily bring it within this repo for tidiness/simplicity if you prefer (or could remove the need for it entirely with some CMake work, but I'm not too familiar with CMake so might take a little while).~~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,78 @@
+on: push
+name: Build
+jobs:
+  build_macos:
+    name: Build (macOS)
+    runs-on: macos-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Check out submodules
+      shell: bash
+      run: |
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+    - name: Install requirements
+      run: brew install sdl2 sdl2_mixer
+    - name: Build
+      run: cd desktop_version/ && mkdir build && cd build && cmake .. && make
+    - name: Upload artifacts (MacOS)
+      uses: actions/upload-artifact@master
+      with:
+        name: mac
+        path: "desktop_version/build/vvvvvv.osx"
+  build_linux:
+    name: Build (Linux)
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Check out submodules
+      shell: bash
+      run: |
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+    - name: Install requirements
+      run: sudo apt-get install libsdl2-dev libsdl2-mixer-dev
+    - name: Build
+      run: cd desktop_version/ && mkdir build && cd build && cmake .. && make
+    - name: Upload artifacts (Linux)
+      uses: actions/upload-artifact@master
+      with:
+        name: linux
+        path: "desktop_version/build/vvvvvv.x86_64"
+  build_windows:
+    name: Build (Windows)
+    runs-on: windows-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Check out submodules
+      shell: bash
+      run: |
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+    - name: Install requirements
+      run: |
+        mkdir lib
+        cd lib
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
+on: push
```
charliebruce: May want to do this on pull request too, as in #30 ?

flibitijibibo:
All of these commits seem good except d7371e88 and 8fdd751f. Can we do builds without changing the SDL2 work at all? Linux and macOS should work with their respective SDL packages, and Windows should grab the VC development packages and set the paths in the cmake command as devs would do normally. You can submit these separately if you want! I would also sync with the other dev for setting up CI.

charliebruce:
Re SDL2 changes - it's fine on Linux, but I had issues on OS X when I was setting it up. Can spend a bit more time looking into it.

Is #30 waiting for any changes, or is it ready to go? I can rebase my changes on top of that if so.

flibitijibibo:
I think the only barrier is getting that checklist I posted all in one place. I don’t do CI at all so I don’t know if there’s any deeper comparisons to be made or if the two should be merged with one another before getting upstreamed.

charliebruce:
With the original method of finding SDL, Linux and Windows work fine, but macOS build doesn't work:

https://github.com/charliebruce/VVVVVV/runs/384546561 

It does identify the correct directory for SDL2 though:
```
-- No TARGET SDL2::SDL2, or SDL2, looking for variables...
-- Found SDL2_INCLUDE_DIRS at /usr/local/Cellar/sdl2/2.0.10/include/SDL2
-- Found SDL2_LIBRARIES at -L/usr/local/Cellar/sdl2/2.0.10/lib  -lSDL2
```

Think this might be a problem with the `sdl2_mixer` Brew package not including a `.cmake` file (`sdl2` has `lib/cmake/SDL2/sdl2-config.cmake`, `sdl2_mixer` does not) but I'm not too familiar with CMake so I'm unsure.

flibitijibibo:
That would explain it, on self-built setups both dylibs end up in /usr/local/lib. I don’t think there exists a way to get the satellite libs separately, since the includes and libs are all meant to be in the location specified by `sdl2-config —cflags —libs`.

charliebruce:
There's a fix - now the macOS runner builds SDL2 from source rather than installing from Brew. Takes ~1.5 min longer but seems like the simplest solution.

ePirat:
What exactly was the issue with using SDL2 from brew? I have it installed using macports and no issue compiling with it at all.

ePirat:
And I believe we should not upload artifacts, they would anyway not be very useful without properly running packaging scripts that ensure library paths are correct and all dependencies are bundled, etc.

.github/workflows/build.yml
```diff
@@ -0,0 +1,97 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Check out submodules
+      shell: bash
+      run: |
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
```
ePirat: Why are you doing these steps? Seems unnecessary?

charliebruce: Ah yes, not necessary any more. I was pulling in some CMake scripts for SDL from a submodule but removed that.

charliebruce:
@ePirat Build log with SDL2/SDL2_mixer installed via the latest Brew is linked above. With Brew, SDL2_mixer is installed in a different directory to SDL2, and this isn't auto-detected by the CMakeList. This ends up failing due to `fatal error: 'SDL_mixer.h' file not found`. Looks like you don't hit this issue because you're using Ninja not Make - not sure why that helps though?

Not too bothered about Artifacts, just did it so that I could check the builds actually worked. Agreed it makes most sense with a packaging system.

Do you want to lead the process of getting this merged, as you got the PR open first? Happy for you to bring my Windows approach over to your branch if you want (unless there's a more elegant way to do it?  My implementation definitely isn't pretty).

ePirat:
Using Ninja vs Make does not make a difference at all (unless CMake is even crappier than I thought).

I think you miss:

```
env:
   CXXFLAGS: -I/usr/local/include/SDL2
   LDFLAGS: -L/usr/local/lib
```

to make it look in the homebrew paths for this.

ePirat:
Sure I can move over your windows parts to my PR if you are ok with that.
I can probably be simplified a bit maybe as the Windows CI machines iirc have some windows package manager so we could use that to fetch SDL2 and SDL2_mixer eventually.

charliebruce:
Re Ninja/Make - ah, that would make more sense. Maybe I installed system SDL2 on my laptop in between tests so that was confusing.

Yes, happy for you to take whatever you like from this PR. Windows CI machines have `vcpkg` which might be useful for installing SDL2? The only other package managers seem to be `choco` and `npm`.

charliebruce:
Superseded by #30

PR #36, opened 2020-01-11
Fix hardreset() not resetting all 50 slots in (custom)collect

InfoTeddy:
## Legal Stuff:

## Changes:

* **Fix `hardreset()` not resetting all 50 slots in `obj.collect` and `obj.customcollect`**

   This means you can actually use more than 20 trinkets/coins (slots in `obj.collect`) and 20 crewmates (slots in `obj.customcollect`) in a custom level, since the only problems with doing so previously was that they wouldn't get reset if you started the level from the beginning.

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -10,6 +10,7 @@ Contributors
 * Elliott Saltar (@eboyblue3)
 * Marvin Scholz (@ePirat)
 * Elijah Stone
+* Info Teddy (@InfoTeddy)
 * Emmanuel Vadot (@evadot)
 * Rémi Verschelde (@akien-mga)
 * viri (viri.me)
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Took a quick look since I hadn't seen this component before, aren't there 100 slots in each vector? (See the 100 loop higher up in the file)

InfoTeddy:
My mistake. I could've sworn that it was 50 slots for both, but after checking the save file (has 100 slots) and checking in GDB (also has 100 slots) it appears that I was wrong.

Thank you, will fix accordingly.

flibitijibibo:
It surprised me too - the size isn’t documented so I just grepped around until I saw the initialization routine. Thanks for the fix!

InfoTeddy:
Ah, I think I know why I thought there was 50. I was confusing it with the fact that UtilityClass::number() only goes up to Fifty, and thus thought that the next reasonable trinkets/crewmates limit after 20 is 50 because any more than that and the numbers always keep saying "Lots".

PR #37, opened 2020-01-11
Do not hardcode default macOS SDK path

ePirat:
## Legal Stuff:

## Changes:

Doing this is not necessary as CMake already looks up the default one
correctly and in fact breaks whenever you have multiple Xcode versions
installed so that Xcode is not called Xcode.app.
(And it does not respect the default Xcode set using xcode-select)

If for some reason this is really needed (maybe older CMake versions did
not look up the default SDK?) let me know and I can change it to use
xcrun to find the right path instead of using a hardcoded one.


ePirat:
cc @charliebruce as this affects a change you made so maybe you can clarify why it was needed for you?

ajmas:
Should this not be a conditional, rather than removing it completely?

Something like:
If ! CMAKE_OSX_SYSROOT then SET(‘...’)

ePirat:
Oh indeed moving it before the PROJECT call apparently fixed the issue that I could not check that before… (As it would always be filled out by CMake at that point already)

I've updated the PR:

The behavior is now to respect `CMAKE_OSX_SYSROOT` if given by the user explicitly (when calling CMake), if its not set, try to use the 10.9 SDK. If the 10.9 SDK is not set, default to whatever default SDK CMake will detect and warn about the fact that the 10.9 SDK is not used.

flibitijibibo:
Something something only happens with Apple something...

ePirat:
It's sad they no longer have the Developer directory outside the Xcode.app so that one can easily have old SDKs around without needing to mess with the Xcode bundles and different Xcode versions.

ajmas:
@ePirat totally agree. At the very least they could provide an easy way to download them.

PR #40, opened 2020-01-11
Look for data.zip in the OSDirectory

evadot:
For FreeBSD user it mean that they can just pkg install VVVVVV
once the package will be available and copy the data.zip in
~/.local/share/VVVVVV as the executable will be in
/usr/local/bin/ and users do not have the permission to write files
in this directory.

Signed-off-by: Emmanuel Vadot <manu@freebsd.org>

## The Basics

Thanks for making a pull request! Before making a pull request, we have some
things for you to read through first:

- We generally do not accept patches for formatting fixes, unless the formatting
  fixes are part of a functional patch (for example, when fixing a bug in a
  function you can fix up the lines surrounding it if needed).
- Patches that break compatibility with the original game data or save data will
  not be accepted.
- New features and user interface changes will most likely not be accepted
  unless they are for improving user accessibility.
- New platforms are acceptable if they use SDL + PhysicsFS and don't mess with
  the game source too much.
    - (No homebrew console targets, sorry! Maybe do the work in SDL instead?)
- Pull requests that do not fill out the Legal Stuff will be closed
  automatically.

If you understand these notes, you can delete the text in this section. Pull
requests that still have this text will be closed automatically.

## Legal Stuff:

## Changes:

Describe your patch here!

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -73,6 +73,12 @@ int FILESYSTEM_init(char *argvZero)
 #ifdef _WIN32
 	strcpy(output, PHYSFS_getBaseDir());
 	strcat(output, "data.zip");
+#elif defined(__FreeBSD__)
+	PLATFORM_getOSDirectory(output);
+	if (strlcat(output, "data.zip", sizeof(output)) >= sizeof(output)) {
+		puts("Cannot find location for data.zip\n");
+		return 0;
+	}
 #else
 	strcpy(output, "data.zip");
 #endif


evadot:
@ibara do you want me to add openbsd ifdef too ?
Will openbsd package do the same ?

evadot:
@flibitijibibo if you don't want that kind of patches in the repository I'll keep it local for FreeBSD packages. I don't mind at all.

flibitijibibo:
Would searching through XDG_DATA_DIRS be better? A possible issue is that we use DATA_HOME as a mounted filesystem, so this kind of recurses in a weird way. This was a request for Linux as well, so if we can do that instead we’ll solve both problems at once.

evadot:
XDG_DATA_DIR default to /usr/local/share which is also not user-writable.
Using PLATFORM_getOSDirectory will try XDG_DATA_HOME and fallback to ~/.local/share/VVVVVV which is easier for a user to put files in.
And yes I think that my patch would be better for Unix based OS.

flibitijibibo:
The recursion is enough for me to think it’s a bad idea. If an administrator is installing the game to the root filesystem they can presumably put other data there as well. A user that cannot do the latter cannot do the former in the first place.

evadot:
What recursion are you talking about ? The fact that we "mount" both ~/.local/share/VVVVVV/ and ~/.local/share/VVVVVV/data.zip ?

flibitijibibo:
Yes. Not that I expect anyone to try and mount files inside mounted directories, but mixing static data with save/mod data doesn’t seem right at all (plus, need I remind the readers that this project technically isn’t accommodating the needs of repositories).

evadot:
That's fine with me, I'll keep a local patch for FreeBSD needs.

PR #45, opened 2020-01-12
Fix Werror always being enabled

xesxen:
## Legal Stuff:

## Changes:

CMake always enabled -Werror due to an error in evaluating ENABLE_WERROR. This patch fixes this problem by fetching the value instead of using the name as a boolean.


flibitijibibo:
Looks good, you can add yourself to `desktop_version/CONTRIBUTORS.txt` and I’ll merge this in.

PR #47, opened 2020-01-12
Unicode rendering

leo60228:
## Legal Stuff:

## Changes:

If font.png is more than 128px tall, it is assumed to be the .png version of the Space Station font. Compatibility with the original font is preserved.

All characters outside ASCII have a width of 8.

The Space Station font (font.png) was created by Matt 'Stelpjo' Aaldenberg.

utfcpp is by @nemtrif. It does not require attribution in binary releases.


desktop_version/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ IF(NOT WIN32)
 ENDIF()
 
 # Include Directories
-INCLUDE_DIRECTORIES(src tinyxml physfs lodepng)
+INCLUDE_DIRECTORIES(src tinyxml physfs lodepng utfcpp/source)
```
flibitijibibo: PhysicsFS has a unicode system in it, can we use that instead of adding a very large dependency?

desktop_version/src/main.cpp
```diff
@@ -1,6 +1,8 @@
 #include <SDL.h>
 #include "SoundSystem.h"
 
+#include "tinyxml.h"
```
flibitijibibo: Why?

leo60228: This was from development and I forgot to remove it, sorry.

desktop_version/CMakeLists.txt
leo60228: PhysicsFS doesn't have a way to get the length of a UTF-8 string in codepoints.

desktop_version/tinyxml/tinyxml.h
```diff
@@ -167,7 +167,7 @@ enum TiXmlEncoding
 	TIXML_ENCODING_LEGACY
 };
 
-const TiXmlEncoding TIXML_DEFAULT_ENCODING = TIXML_ENCODING_UNKNOWN;
+const TiXmlEncoding TIXML_DEFAULT_ENCODING = TIXML_ENCODING_UTF8;
```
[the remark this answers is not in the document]
leo60228: This seems to be what the tinyxml docs recommend to do.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -37,6 +37,7 @@ int FILESYSTEM_init(char *argvZero)
 	char output[MAX_PATH];
 	int mkdirResult;
 
+        PHYSFS_permitSymbolicLinks(1);
```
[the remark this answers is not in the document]
leo60228: This made testing much easier and doesn't cause any harm, should I remove it?

leo60228:
For what it's worth, over 2/3 of tinyxml is documentation and tests.

leo60228:
utfcpp I mean, not tinyxml...

flibitijibibo:
If it’s not needed to build we can remove it (aside from the license). This still seems like a very large dependency for what is essentially a single function.

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
flibitijibibo: I was asked a long time ago about supporting something like this, I would accept it as a separate patch.

desktop_version/tinyxml/tinyxml.h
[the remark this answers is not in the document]
flibitijibibo: Is it absolutely required to make it work or can we fix this on the application side?

flibitijibibo:
Still have some dot files, but WOW that repo must be pretty bloated. 3000+ lines to almost less than 500 is quite dramatic, no?

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
leo60228: #48

desktop_version/tinyxml/tinyxml.h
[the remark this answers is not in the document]
leo60228: I think we can do it on the application side.

flibitijibibo:
Right, so that covers the dependency issue... I am definitely more okay with a dependency that’s ~400 lines, rather than 10 times that(!!!).

I do still think the changes need to be designed differently, however - try to aim for the absolute minimum C++ standard possible (if you can get away with it, write it as basically C). We have to support some very old compilers, so there’s no room to be overly fancy.

The char list in particular is pretty extreme - if there’s a way to determine the font information at load time instead of every lookup, and with a... slightly less hardcoded charset, that will make maintaining the font a lot easier in the long run.

flibitijibibo:
Possible route towards fixing the mega array: should we have a font.txt next to the PNG to provide the list of supported chars? This would allow for a default charset with the current font.png (when no txt is found) and allow for custom fonts, including the one you’ve made and other font updates in the future.

(Sidenote: We’ll probably not include the custom font in the repo, that can be a content mod until Terry decides to update the font officially.)

leo60228:
That sounds good to me.

desktop_version/src/Graphics.cpp
```diff
@@ -129,6 +130,20 @@ Graphics::Graphics()
     warprect.x = 0, warprect.y = 0, warprect.w = 0, warprect.h = 0;
 }
 
+int Graphics::font_idx(char32_t ch) {
+    if (grphx.im_bfont->h > 128) {
+        auto font = UR"( !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~ ¡¢£¤¥¦§¨©«­®°²³»¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿĀāĂăĄąĆćĈĉĊċČčĎďĐđĒēĔĕĖėĘęĚěĜĝĞğĠġĢģĤĥĦħĨĩĪīĬĭĮįİıĲĳĴĵĶķĸĹĺĻļĽľĿŀŁłŃńŅņŇňŉŊŋŌōŎŏŐőŒœŔŕŖŗŘřŚśŜŝŞşŠšŢţŤťŦŧŨũŪūŬŭŮůŰűŲųŴŵŶŷŸŹźŻżŽžſΆΈΉΊΌΎΏΐΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩΪΫάέήίΰαβγδεζηθικλμνξοπρςστυφχψωϊϋόύώЀЁЂЃЄЅІЇЈЉЊЋЌЍЎЏАБВГДЕЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯабвгдежзийклмнопрстуфхцчшщъыьэюяѐёђѓєѕіїјљњћќѝўџאבגדהוזחטיךכלםמןנסעףפץצקרשתװױײ׳״—‘’“”€←↑→↓↔↕↖↗↘↙↰↱↲↳↶↷↺↻↼↽↾↿⇀⇁⇂⇃⇅⇆─━│┃┄┅┆┇┈┉┊┋┌┍┎┏┐┑┒┓└┕┖┗┘┙┚┛├┝┞┟┠┡┢┣┤┥┦┧┨┩┪┫┬┭┮┯┰┱┲┳┴┵┶┷┸┹┺┻┼┽┾┿╀╁╂╃╄╅╆╇╈╉╊╋╌╍╎╏═║╒╓╔╕╖╗╘╙╚╛╜╝╞╟╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╯╰╱╲╳╴╵╶╷╸╹╺╻╼╽╾╿▀▁▂▃▄▅▆▇█▉▊▋▌▍▎▏▐░▒▓▔▕▖▗▘▙▚▛▜▝▞▟■□▢▲△▶▷▼▽◀◁、。〃〈〉《》「」『』【】〒〓〔〕〖〗〘〙〚〛〜〝〞ぁあぃいぅうぇえぉおかがきぎくぐけげこごさざしじすずせぜそぞただちぢっつづてでとどなにぬねのはばぱひびぴふぶぷへべぺほぼぽまみむめもゃやゅゆょよらりるれろゎわゐゑをんゔゕゖ゠ァアィイゥウェエォオカガキギクグケゲコゴサザシジスズセゼソゾタダチヂッツヅテデトドナニヌネノハバパヒビピフブプヘベペホボポマミムメモャヤュユョヨラリルレロヮワヰヱヲンヴヵヶヷヸヹヺ・ーＵｇｈｉｐｒｔ)";
```
JayFoxRox: I can easily see this breaking in some editors (or being modified by accident). Such data should be loaded from another file, or at least `#include`'d (although I'm not even sure if all supported C++ compilers must support such charsets).

JayFoxRox: The name `font` feels like a misnomer; isn't this supposed to be `charset` or `symbols` or something?

JayFoxRox: This line is very long. I know that even in 2020 some editors suck with long lines. gedit for example becomes horribly slow or suddenly stops drawing correctly. I'm not sure if this is a GTK or gedit problem.
You should split symbols into different lines (if they are arranged in a texture, then maybe follow that scheme).

flibitijibibo: We’re looking at replacing this with a text file, so this should vanish eventually...

desktop_version/src/Graphics.cpp
```diff
@@ -129,6 +130,20 @@ Graphics::Graphics()
...
+        auto ptr = std::char_traits<char32_t>::find(font, std::char_traits<char32_t>::length(font), ch);
```
JayFoxRox: This looks like it will raise the C++ standard a lot.

*(Another thing: I'm currently porting VVVVVV to a platform without unicode and already struggle with the required unicode/utf8 support in the physfs backend. With VVVVVV being software render-able at 320x240, very few dependencies and very little requirements for input devices, I can see it being ported to many niche platforms with sub-par toolchains, maintained by a small number of people - I'm not sure how widespread unicode support will be; so it might make sense to keep a non-unicode path for minimalistic builds for retro platforms / compilers)*

desktop_version/src/Graphics.cpp
JayFoxRox: This must be the mentioned "mega array" then - I wasn't sure (as there was no code-comment with context).

desktop_version/src/Graphics.cpp
```diff
@@ -146,33 +161,49 @@ void Graphics::drawspritesetcol(int x, int y, int t, int c, UtilityClass& help)
...
-			SDL_FreeSurface(temp);
-
         }
     }
 
-    //Ok, now we work out the lengths (this data string cortesy of a program I wrote!)
-    for (int i = 0; i < 256; i++)
-    {
-        bfontlen.push_back(6);
+    PHYSFS_File* charmap_file = PHYSFS_openRead("graphics/font.txt");
```
flibitijibibo: FILESYSTEM_loadFileToMemory should do most of this work for you.

leo60228:
My keyboard broke midway through this commit and my only spare is super difficult to type on. I'm gonna work on this some more tomorrow.

desktop_version/src/Graphics.cpp
```diff
@@ -129,6 +131,19 @@ Graphics::Graphics()
     warprect.x = 0, warprect.y = 0, warprect.w = 0, warprect.h = 0;
 }
 
+int Graphics::font_idx(char32_t ch) {
+    if (grphx.im_bfont->h > 128) {
```
flibitijibibo: This part is still a hack - if we have the map in the first place, that’s enough of an indicator to use the special path.

desktop_version/src/Graphics.cpp
```diff
@@ -3,6 +3,8 @@
 #include "Entity.h"
 #include "Map.h"
 #include "Screen.h"
+#include <utf8/checked.h>
+#include <physfs.h>
```
flibitijibibo: Should be removable after the FILESYSTEM call is introduced...

desktop_version/font.txt
```diff
@@ -0,0 +1 @@
+ !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~ ¡¢£¤¥¦§¨©«­®°²³»¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿĀāĂăĄąĆćĈĉĊċČčĎďĐđĒēĔĕĖėĘęĚěĜĝĞğĠġĢģĤĥĦħĨĩĪīĬĭĮįİıĲĳĴĵĶķĸĹĺĻļĽľĿŀŁłŃńŅņŇňŉŊŋŌōŎŏŐőŒœŔŕŖŗŘřŚśŜŝŞşŠšŢţŤťŦŧŨũŪūŬŭŮůŰűŲųŴŵŶŷŸŹźŻżŽžſΆΈΉΊΌΎΏΐΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩΪΫάέήίΰαβγδεζηθικλμνξοπρςστυφχψωϊϋόύώЀЁЂЃЄЅІЇЈЉЊЋЌЍЎЏАБВГДЕЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯабвгдежзийклмнопрстуфхцчшщъыьэюяѐёђѓєѕіїјљњћќѝўџאבגדהוזחטיךכלםמןנסעףפץצקרשתװױײ׳״—‘’“”€←↑→↓↔↕↖↗↘↙↰↱↲↳↶↷↺↻↼↽↾↿⇀⇁⇂⇃⇅⇆─━│┃┄┅┆┇┈┉┊┋┌┍┎┏┐┑┒┓└┕┖┗┘┙┚┛├┝┞┟┠┡┢┣┤┥┦┧┨┩┪┫┬┭┮┯┰┱┲┳┴┵┶┷┸┹┺┻┼┽┾┿╀╁╂╃╄╅╆╇╈╉╊╋╌╍╎╏═║╒╓╔╕╖╗╘╙╚╛╜╝╞╟╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╯╰╱╲╳╴╵╶╷╸╹╺╻╼╽╾╿▀▁▂▃▄▅▆▇█▉▊▋▌▍▎▏▐░▒▓▔▕▖▗▘▙▚▛▜▝▞▟■□▢▲△▶▷▼▽◀◁、。〃〈〉《》「」『』【】〒〓〔〕〖〗〘〙〚〛〜〝〞ぁあぃいぅうぇえぉおかがきぎくぐけげこごさざしじすずせぜそぞただちぢっつづてでとどなにぬねのはばぱひびぴふぶぷへべぺほぼぽまみむめもゃやゅゆょよらりるれろゎわゐゑをんゔゕゖ゠ァアィイゥウェエォオカガキギクグケゲコゴサザシジスズセゼソゾタダチヂッツヅテデトドナニヌネノハバパヒビピフブプヘベペホボポマミムメモャヤュユョヨラリルレロヮワヰヱヲンヴヵヶヷヸヹヺ・ーＵｇｈｉｐｒｔ
```
flibitijibibo: These files won’t be in the repo - they’ll be a community mod at first, with a possible official integration directly into data.zip later down the line.

JayFoxRox:
> @JayFoxRox does this at least compile and not crash when the mod is absent? If not, we’ll need to figure out a safe way to support this (maybe a text interface with UTF8 and ASCII backends?).

Not sure which mod you mean (mod[ification] = changes in this PR?).

Before commenting earlier I didn't try compiling this PR. I actually did compile it now and found *another* issue (with utfcpp): dependency on C++ exceptions.

Our toolchain does not currently support those:

```
VVVVVV/desktop_version/utfcpp/source/utf8/checked.h:81:17: error: cannot use 'throw' with exceptions disabled
                throw not_enough_room();
```

There doesn't seem to be a way to disable them with utfcpp?

<details>
<summary>This explanation about my ongoing port is offtopic, so I'll fold this (click here for more information)</summary><br>

My port (to original Xbox) is currently blocked by issues in our SDL2 port and a lot of stupid changes required for PhysFS. It still compiles though. But even the VVVVVV master revision does not work correctly on original Xbox (crashes in sound initialization), yet. As we don't have a debugger in our toolchain it will take some time to understand the crashes I'm getting. See https://github.com/JayFoxRox/VVVVVV/pull/1.

However, I know that our original Xbox toolchain doesn't (currently?) support unicode in winapi (selfmade winapi abstraction for Xbox NT kernel) and also doesn't support `wchar` in our libc (pdclib). We use LLVM libcxx for C++, but it was only upstreamed a week ago and I wasn't involved with it (not sure if that supports UTF / unicode in our port).
Even if compiling with stubs in the libc / libcxx might work, then I'd expect issues at runtime.

I assume similar issues, weak UTF / unicode and exception support, for other homebrew SDKs for smaller platforms (PS2? PSP? Event badges?). At least original Xbox has a modern compiler (LLVM clang) - I heard that some PS2 SDKs use an ancient gcc (not sure if Sony SDK or homebrew SDK).

---
</details>

I don't really understand why VVVVVV has to process strings itself anyway - most of it are probably constant texts or very simple formatting operations. So I'm not sure if the code would really need to understand UTF / Unicode - one could just map symbols to glyphs from data files.

Localization is obviously a great feature that I'd also want to use in the future. A lightweight or optional implementation of Unicode is totally fine with me.

Alternatively, you could also make the decision to drop support for niche platforms to focus on code quality / simplicity - which is also fine if it is communicated correctly, although it would be a bit disappointing.

flibitijibibo:
If we can remove the exceptions that’d be great - we don’t even handle them anyway so it’d be nice to make those optional (and submit the changes upstream of course).

I think the problem with data processing is the editor, we may have to worry about custom scripts and levels. Maybe not though, if it’s easy to read/write the correct format (presumably we’d write this format for the static data anyhow).

leo60228:
The main purpose of this was so custom levels could be written in foreign languages.

leo60228:
@JayFoxRox Does your toolchain just not have throwing std functions, or do they abort instead of throwing? Should I avoid using them?

leo60228:
I'm not catching the exceptions, for what it's worth.

leo60228:
Also, for what it's worth, no UTF-8 strings ever get passed to libc or std functions that handle them specially. std::string is encoding-agnostic to any 8-bit encoding (which is why utfcpp is necessary in the first place).

desktop_version/utfcpp/source/utf8/core.h
```diff
@@ -0,0 +1,292 @@
...
+
+#ifndef UTF8_FOR_CPP_CORE_H_2675DCD0_9480_4c0c_B92A_CC14C027B731
+#define UTF8_FOR_CPP_CORE_H_2675DCD0_9480_4c0c_B92A_CC14C027B731
+
+#include <iterator>
+#include <cstdio>
+#include <cstdlib>
+
+//#define UTF_CPP_THROW(X) throw (X);
+#define UTF_CPP_THROW(X) (printf("UTF-8 error: %s", (X).what()), exit(1))
```
flibitijibibo: In theory I don’t mind this, but we should ask the utfcpp maintainer about this first. We may also meed to define UTF_CPP_CPLUSPLUS to fix some of the exception keywords and includes in both files. @JayFoxRox, would you be able to help with this?

JayFoxRox: Unfortunately I can't help with this; I have many projects to maintain myself.

JayFoxRox:
Thanks for taking the time to respect demands of limited platforms :+1: 

> @JayFoxRox Does your toolchain just not have throwing std functions, or do they abort instead of throwing? Should I avoid using them?

It wouldn't even build, because the compiler will depend on features in our libc/libcxx. Small teams (like us hobbyists) don't have the capacity to support these features easily (it's easy to have a "functional" port of a libc/libcxx, but harder to have complete port which depends on proper stack unwinding and other platform / ABI dependent features).

Basically there are some non-essential C++ features that are "harder" to port; so for custom platforms these typically aren't implemented or very low priority. We also still lack RTTI (https://github.com/XboxDev/nxdk/issues/237).
**I should note, that I assume that this makes our toolchain non-standard compliant.** So wether VVVVVV respects such "incomplete" platforms has to be decided.

This usually isn't an issue as software from scratch can respect those limitations. Games typically don't use these features either - especially those which are suitable for even running on 20 year old hardware (low memory, slow CPU, limited resolution, no full blown OS, ..).

If I were to remove `-fno-exceptions` from our Makefile, then our toolchain causes this during linking:

```
lld: error: undefined symbol: ___std_terminate
lld: error: undefined symbol: ___CxxFrameHandler3
...
```

As our toolchain improves we'll probably add support for these features.

> I'm not catching the exceptions, for what it's worth.

Exactly. Ideally software that doesn't use these features shouldn't require them in the first place for portability sake (not just for our toolchain, but also other potential stakeholders which depend on non-FOSS toolchains which are impossible to update).

> Also, for what it's worth, no UTF-8 strings ever get passed to libc or std functions that handle them specially.

Sounds good, that shouldn't be an issue then :+1:

desktop_version/src/KeyPoll.cpp
```diff
@@ -93,7 +93,7 @@ void KeyPoll::Poll()
 
 			if (textentrymode)
 			{
-				if (evt.key.keysym.sym == SDLK_BACKSPACE)
+				if (evt.key.keysym.sym == SDLK_BACKSPACE && keybuffer.size() > 0)
```
Vest: Please pardon me, I am just curious. Would it be better to use here **keybuffer.empty** instead of **keybuffer.size**?

leo60228:
@Voct Probably, yes.

leo60228:
@Vest

Stelpjo:
What's the situation on this right now? I'm really looking forward to UTF-8 support in VVVVVV.

flibitijibibo:
We need to write (and ideally upstream) some patches to make the C++ exceptions optional. That's the only serious blocker I remember, tbh.

leo60228:
Those are already in the PR. I haven't tried to upstream them, though.

leo60228:
Oh, turns out that `#include <utf8/unchecked.h>` is specifically intended to compile without exceptions. I didn't realize that from the docs.

leo60228:
I'll migrate to that.

Stelpjo:
By the way, I did update the font file to include some more characters, including a few dingbats; it's available on the Community Edition repo.

leo60228:
Just rebased.

flibitijibibo:
As a reminder, we're still not hosting the custom fonts - that will only be in the CE repo. Will let the buildbots do the rest of the talking...

leo60228:
@flibitijibibo In VVVVVV-CE, I use FriBidi to handle bidirectional text. FriBidi is a very heavy library, but I wasn't able to find something more lightweight. Do you have any suggestions on this?

desktop_version/src/KeyPoll.cpp
```diff
@@ -94,11 +95,12 @@ void KeyPoll::Poll()
...
 			if (textentrymode)
 			{
-				if (evt.key.keysym.sym == SDLK_BACKSPACE)
+				if (evt.key.keysym.sym == SDLK_BACKSPACE && !keybuffer.empty())
 				{
-					bool kbemptybefore = keybuffer.empty();
-					keybuffer = keybuffer.substr(0, keybuffer.length() - 1);
-					if (!kbemptybefore && keybuffer.empty())
+					std::string::iterator iter = keybuffer.end();
+                                        utf8::unchecked::prior(iter);
```
flibitijibibo: Tab...

desktop_version/src/Graphics.h
```diff
@@ -280,6 +283,8 @@ class Graphics
 	bool translucentroomname;
 
 	bool showmousecursor;
+
+        std::map<int, int> font_positions;
```
flibitijibibo: Tabber...

desktop_version/src/Graphics.h
```diff
@@ -28,6 +29,9 @@ class Graphics
 
 	GraphicsResources grphx;
 
+        int bfontlen(char32_t ch);
```
flibitijibibo: ... TABBEST!

desktop_version/src/Graphics.cpp
```diff
@@ -3100,28 +3110,24 @@ void Graphics::bigrprint(int x, int y, std::string& t, int r, int g, int b, bool
...
 	int bfontpos = 0;
 	int cur;
-	for (size_t i = 0; i < t.length(); i++)
-	{
-		cur = (t.c_str())[i];
-		if (cur > 255 || cur < 0)
-		{
-			cur = '?';
-		}
+        std::string::iterator iter = t.begin();
```
flibitijibibo: Tabination!

leo60228:
I wasn't aware of #118 when I removed stelpjo, but it turns out they were in CONTRIBUTORS.txt twice after rebasing anyway.

flibitijibibo:
Looks like Graphics.cpp got pretty thrashed in the last update - the only way I can really see all the mismatches now is to Ctrl+F "    " and Highlight All. Still a lot of mixing and matching going on for some reason. All the other files seem okay.

leo60228:
@flibitijibibo I reindented every function I could recall changing. Should I just reindent the whole file with tabs?

leo60228:
I'm not sure what was already indented with spaces, so I've only reindented what was mentioned in the review.

flibitijibibo:
Squashed into upstream, thanks for all the work on this!

PR #48, opened 2020-01-12
Support symbolic links

leo60228:
## Legal Stuff:

## Changes:

Adds support for symbolic links.

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -37,6 +37,7 @@ int FILESYSTEM_init(char *argvZero)
 	char output[MAX_PATH];
 	int mkdirResult;
 
+        PHYSFS_permitSymbolicLinks(1);
 	PHYSFS_init(argvZero);
 
 	/* Determine the OS user directory */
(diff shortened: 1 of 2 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -37,6 +37,7 @@ int FILESYSTEM_init(char *argvZero)
 	char output[MAX_PATH];
 	int mkdirResult;
 
+        PHYSFS_permitSymbolicLinks(1);
```
JayFoxRox: nit: Wrong indentation?

flibitijibibo: Oh weird, it looked right in the editor- I’ll fix this one myself, good catch.

ePirat: Is it fine to call that before calling `PHYSFS_init`?

JayFoxRox: https://github.com/TerryCavanagh/VVVVVV/blob/f85622b7e95e8b63859e1f03eb900dec10f6ae5a/desktop_version/physfs/physfs.h#L701-L702

So @ePirat just found another bug with this.

However, in practice this doesn't cause any issues *yet*, because it only sets a variable and it's only reset by a `PHYSFS_deinit` or *failed* `PHYSFS_init` (possibly a bug in PhysFS). It should still be fixed.

flibitijibibo:
Will fix this in just a moment - if anyone wants to file a bug with Ryan, go for it.

PR #49, opened 2020-01-12
Fix mouse coords not being scaled properly in stretch mode

### LINKED ISSUE #27: [Editor] Mouse coordinates not scaled to viewport
It seems like if you're in stretch mode, the game will still keep tracking your mouse position as if the window was still 320x240. So if your window is (say) 640x480, then the mouse position is more accurate near 0,0 but gets slowly more out-of-sync with the in-game cursor the more to the right and bottom you move it.

I recommend trying this on a build without 146cdde217175b4ac0a37d7058ebe8b70bb6dc48 so you can clearly see the mouse cursor.

InfoTeddy:
## Legal Stuff:

## Changes:

* **Fix a bug where your mouse coords were not scaled to the viewport in stretch mode**

  Closes #27.

### DIFF
--- a/desktop_version/README.md
+++ b/desktop_version/README.md
@@ -9,7 +9,8 @@ VVVVVV's official desktop versions are built with the following environments:
 The engine depends solely on [SDL2](https://libsdl.org/) and
 [SDL2_mixer](https://www.libsdl.org/projects/SDL_mixer/). All other dependencies
 are statically linked into the engine. The development libraries for Windows can
-be downloaded from their respective websites, Linux and macOS developers should
+be downloaded from their respective websites, Linux developers can find the dev
+libraries from their respective repositories, and macOS developers should
 compile and install from source.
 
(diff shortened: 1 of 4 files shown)


InfoTeddy:
Ignore the extra 4 commits, they're just old history. Git does that sometimes.

flibitijibibo:
That... was way easier than I expected! Will `git am` the commit later today.

InfoTeddy:
Yes, but 90% of the time was me trying to track down where mouse coordinates are handled and where the variable for stretch mode actually is. Then I had to discard a previous attempt because it wasn't working and I misunderstood something. Then I had to figure out that you can't do this for any of the other graphics modes or they'll have the inverse problem. :stuck_out_tongue:

flibitijibibo:
Applied manually: https://github.com/TerryCavanagh/VVVVVV/commit/3e9bd3aaac77aec75e787b67e39dc31d42dcc664

PR #52, opened 2020-01-12
Handle filepaths with non-ASCII characters properly on Windows

Dav999-v:
## Legal Stuff:

## Changes:

On Windows, players with an account username that contains non-ASCII characters (any foreign characters, for example) will probably find that the VVVVVV directory is not created, or if it is, that no save/level files are created inside of it and any downloaded custom level cannot be loaded. This pull request fixes these issues.

The commit messages document things in more detail, but in summary: the two Windows API functions that are used to get the path to the Documents folder and to create new folders now use their "Wide" (UTF-16) variants converted to and from UTF-8, and loading and saving XML files now uses PHYSFS to load and save the files instead of using TinyXML's functions, since TinyXML doesn't handle UTF-8 paths properly, while PHYSFS does.

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -6,6 +6,7 @@ Contributors
 * Christoph Böhmwalder (@chrboe)
 * Charlie Bruce (@charliebruce)
 * Brian Callahan (@ibara)
+* Dav999 (Dav999-v)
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Oof, this one’s definitely on me - not even just the unicode part, but also the poor filesystem management in the TiXml code.

Thanks for the fix!

PR #58, opened 2020-01-12
Conditionally build PhysFS Windows and Unix sources too

ePirat:
## Legal Stuff:

## Changes:

Without doing this, due to preprocessor macros those files would
evaluate to an empty translation unit, and an empty object, which
produces warnings like:

```text
warning: ISO C requires a translation unit to contain at least one declaration
```

Note: WIN32 is true not only for 32bit Windows but 64bit too, so
should cover all cases.

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -107,13 +107,15 @@ SET(PFS_SRC
 	../third_party/physfs/physfs_byteorder.c
 	../third_party/physfs/physfs_unicode.c
 	../third_party/physfs/physfs_platform_posix.c
-	../third_party/physfs/physfs_platform_unix.c
-	../third_party/physfs/physfs_platform_windows.c
 )
-IF(APPLE)
+if(APPLE)
 	# Are you noticing a pattern with this Apple crap yet?
-	SET(PFS_SRC ${PFS_SRC} ../third_party/physfs/physfs_platform_apple.m)
-ENDIF()
+	list(APPEND PFS_SRC ../third_party/physfs/physfs_platform_apple.m)
+elseif(WIN32)
+	list(APPEND PFS_SRC ../third_party/physfs/physfs_platform_windows.c)
+else()
+	list(APPEND PFS_SRC ../third_party/physfs/physfs_platform_unix.c)
+endif()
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Could this possibly count as a PhysicsFS bug? If it’s conditionally built in Ryan’s CMake file we should do it here too, otherwise we may want to patch it there first.

ePirat:
Possibly, I'll have a look in a bit as I anyway wanted to check if we can get rid of Objective-C usage.

ePirat:
Indeed it is the same upstream: https://github.com/criptych/physfs/blob/master/CMakeLists.txt#L76 (I did not find a source viewer for the official repo so this is some git mirror…)

flibitijibibo:
Just checked and it is indeed monolithic - we can probably rip off FAudio here:

https://github.com/FNA-XNA/FAudio/blob/master/src/FAudio_ffmpeg.c#L451

We can file both the Obj-C and TU issues separately, my guess is he’ll be open to both.

ePirat:
Hi, sorry for lack of update on this.

Do you know how/where to file the issues upstream or contribute patches to PhysFS?
Should those just be send to their mailing list? I've never used mercurial before so I'll have to check how to produce a patch with it.

flibitijibibo:
The mailing list is probably fine, there's a Bugzilla but I don't know how closely it's monitored:

https://bugzilla.icculus.org/buglist.cgi?component=everything&product=PhysicsFS&resolution=---

InfoTeddy:
PhysFS has a [GitHub repository](https://github.com/icculus/physfs/) now.

Unless we really want to patch it here (@flibitijibibo?), I'm thinking we should just close this as "not our bug".

ePirat:
Feel free to close this as it's not directly a VVVVVV bug.

flibitijibibo:
At some point when I do the FAudio work I'll move everything in third_party to a submodule (to anyone reading this, don't do this yet!), and we can file a PhysFS bug at some point referencing this issue.

ePirat:
Ok, although be aware that Git submodules can be a bit of a pain to work with, especially if you had a folder before it was a submodule with the same name. (As it will then complain whenever you check out older revisions/branches that did not have the submodule yet and the other way around)

PR #59, opened 2020-01-12
fix actionscript misclassified as angelscript

notviri:
## Legal Stuff:

## Changes:

Fix all the ActionScript in the repository being misclassified as AngelScript. I know you can't actually see it on the branch because linguist doesn't work on branches [but you can see it here](https://github.com/notviri/test1234) unless the language bar dissapears (it keeps flickering because I just committed the contents of the entire repo there), but:
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.as linguist-language=ActionScript


notviri:
I'll delete the repo now that's just a clone of this (but I'll leave the image).

PR #61, opened 2020-01-13
Stop the game from freezing if we play a song during a fadeout

InfoTeddy:
## Legal Stuff:

## Changes:

* **Make the game no longer freeze when `music.play()` is called during a fadeout**

  If a `play()` is called during a fadeout, we'll write down the requested song, then wait until the fadeout ends to play it. Except if the fadeout comes from the internal command `musicfadeout`.


InfoTeddy:
I don't know what the .gch file is, I tried to un-add it from the commit but apparently it's still there.

Previously I tried to find such a function, but couldn't find it. Turns out it's actually `Mix_FadingMusic()`, so I'll use that one, thanks.

flibitijibibo:
This seems harmless enough, but since we're talking about a freeze fix I just want to be sure: Do you have an example of a before/after that I can try out? Mostly just want to make sure we don't accidentally cause anything weird like music not showing up or possible other weirdness from, for example, going back and forth to constantly spam the Fade(500) call (this is one thing mixer's API is bad at, how does it even deal with that?).

InfoTeddy:
This is mostly when loading into a custom level. It takes longer to load a level because the game freezes when doing the `play()` for the custom level's song since it's during a fadeout that it happens. As well, previously we tried to use `musicfadeout` in custom levels to have our own niceplay but it ended up freezing the game every time, because of the fadeout.

I've attached a level that demonstrates this.

* [freeze_fix_demonstration.zip](https://github.com/TerryCavanagh/VVVVVV/files/4051842/freeze_fix_demonstration.zip)

desktop_version/src/Music.cpp
```diff
@@ -260,6 +262,15 @@ void musicclass::play(int t)
...
 				// musicchannel.addEventListener(Event.SOUND_COMPLETE, loopmusic);
+				if (Mix_FadingMusic() == MIX_FADING_OUT) {
+					// We're already fading out
+					fadeoutqueuesong = t;
+					currentsong = -1;
+					if (!dontquickfade)
+						Mix_FadeOutMusic(500); // fade out quicker
+					else
+						dontquickfade = false;
+				} else
```
flibitijibibo: else on separate line

InfoTeddy: Fixed in cd31f0dbab85da2d38050231866a1b62662a7108.

flibitijibibo:
Got a chance to test real quick and it seems good. Once the .gch file is gone I'll pull this in (I should probably add a .gitignore afterward too).

InfoTeddy:
Bruh, I swear I removed it before... whatever, it should be removed now.

InfoTeddy:
It's also worth noting that removing this freeze also fixes a bug where, if you had the game muted, the game audio would temporarily unmute itself if this exact same thing happened, too. Apparently if the game freezes like this the mute button stops working, too. I didn't know that, and I thought the unmute bug was a separate issue.

So... I get credit for fixing two bugs with one patch, right?

PR #63, opened 2020-01-13
Add an accessibility option to toggle the fake loading screen

InfoTeddy:
## Legal Stuff:

## Changes:

* **Add an option in the Accessibility menu to toggle the fake loading screen**

  This is the variable `game.skipfakeload`, which gets saved as `<skipfakeload>` in `unlock.vvv`.

### DIFF
--- a/desktop_version/src/titlerender.cpp
+++ b/desktop_version/src/titlerender.cpp
@@ -530,8 +530,10 @@ void titlerender(Graphics& dwgfx, mapclass& map, Game& game, entityclass& obj, U
                 {
                     dwgfx.Print( -1, 105, "Game speed is at 40%", tr, tg, tb, true);
                 }
-            } else if (game.currentmenuoption == 4) {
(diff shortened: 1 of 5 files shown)


desktop_version/src/titlerender.cpp
```diff
@@ -530,6 +530,12 @@ void titlerender(Graphics& dwgfx, mapclass& map, Game& game, entityclass& obj, U
                 {
                     dwgfx.Print( -1, 105, "Game speed is at 40%", tr, tg, tb, true);
                 }
+            } else if (game.currentmenuoption == 4) {
+                dwgfx.bigprint(-1, 30, "Fake load screen", tr, tg, tb, true);
```
flibitijibibo: Brace on a separate line from else if, also capitalization on Fake Load Screen. Otherwise looks good.

InfoTeddy: Fixed in dd2cdaddf7a0202c07017e92cbbd01c80029b638.

PR #70, opened 2020-01-13
Don't deactivate the "- Press ENTER to Teleport -" prompt when dying

InfoTeddy:
## Legal Stuff:

## Changes:

* **Don't deactivate the teleport prompt when dying**

  This fixes a bug where if you died after activating a teleporter prompt and then respawned in the same room as the teleporter, the teleporter prompt would go away. Which would be very annoying if you wanted to, you know, teleport.

  You don't even have to R-press to make this happen. "Level Complete!" and Energize are two teleporter rooms in the main game that have hazards in them.

  I see no reason to set game.activetele to false when dying. For one, it gets set to false during a gotorom anyway. For another, I couldn't get the game to activate the teleport prompt in a room without a teleporter in my testing, mostly because when you touch a teleporter, your checkpoint is set to that teleporter, so you can't R-press to go to another room and carry over the teleport prompt.

### DIFF
--- a/desktop_version/src/Logic.cpp
+++ b/desktop_version/src/Logic.cpp
@@ -671,7 +671,6 @@ void gamelogic(Graphics& dwgfx, Game& game, entityclass& obj,  musicclass& music
                 }
 
                 game.gethardestroom(map);
-                game.activetele = false;
                 game.hascontrol = true;
 
 


flibitijibibo:
A possible test for this is placing the teleporter exactly on a death zone, then dying within a valid teleportation spot and hitting enter. I would be interested to know what would happen in that scenario.

InfoTeddy:
Like this? (custom graphics mine)
teleporter_death_test
It seems like the prompt is still there.

flibitijibibo:
Cool, so it doesn’t open any dialogs or run a teleportation? If not I can merge. Sometimes this stuff gets pretty fragile, so I try to think of the obvious exploits at least.

InfoTeddy:
Nope. It looks like pressing Enter has no effect. Probably because you're not allowed to press Enter during the death sequence, but also because you have to be standing still in order to activate any prompt in the game, and moving into a hazard means you're no longer standing still.

flibitijibibo:
Works for me!

PR #71, opened 2020-01-14
Re-add the font outline from the Flash and mobile versions

InfoTeddy:
## Legal Stuff:

## Changes:

* **Re-add the font outline from the Flash and mobile versions of the game**

  This uncomments the code for the font outline. Now, you will see the font outline on text that had it in the Flash version, most notably "- Press ENTER to Teleport -" and the time trial overlay text.

  This also fixes a problem with the commented code, where un-centered text didn't outline properly.

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -380,19 +380,19 @@ void Graphics::bprint( int x, int y, std::string t, int r, int g, int b, bool ce
 {
 
     //printmask(x, y, t, cen);
-    //Print(x, y - 1, t, 0, 0, 0, cen);
-    //if (cen)
-    //{
-    //	//TODO find different
-    //	PrintOff(-1, y, t, 0, 0, 0, cen);
-    //	PrintOff(1, y, t, 0, 0, 0, cen);
-    //}
-    //else
-    //{
-    //	Print(x  -1, y, t, 0, 0, 0, cen);
-    //	Print(x , y, t, 0, 0, 0, cen);
-    //}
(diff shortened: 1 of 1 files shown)


flibitijibibo:
There might be a historical reason for why this was disabled... @terrycanavagh, any ideas?

InfoTeddy:
Does editing in a ping actually ping the person involved? I know Discord doesn't ping someone if you edit in a ping, and will not remove a ping if you edit out a ping...

Just in case, @TerryCavanagh

TerryCavanagh:
It’s been a very long time, and I no longer fully remember why I removed the font outlines in the c++ build - but I suspect I did it intentionally because I didn’t like the inconsistent look of the outlined text.

But I dunno, it’s been 9 years, I can take a fresh look. Do you have a before/after in context?

InfoTeddy:
Before:
font_outline_before

After:
font_outline_after

For what it's worth, the mobile version still has the font outline, and they have the same look as they do here.

FredrIQ:
It could always be made a graphics option, but that might be overkill.

TerryCavanagh:
Thanks for going to the trouble of getting both screenshots!

I'm a little conflicted about this, but the outline looks fine in context, and if it's in the mobile version anyway then I don't see the harm in making the changed. Merging!

flibitijibibo:
I like @FredrIQ’s idea of making it an option - if not accessibility I’d be up for having it in graphics too.

flibitijibibo:
Toggle has been pushed: https://github.com/TerryCavanagh/VVVVVV/commit/6e9712f113388af4e3711773555e55c8620102b1

PR #78, opened 2020-01-14
Fix the flipgravity() internal command

InfoTeddy:
## Legal Stuff:

## Changes:

* **Fix the `flipgravity()` internal command**

  There are two main problems with `flipgravity()`:
  ​ ​ 1. It doesn't work for an already-flipped crewmate.
  ​ ​ 2. It doesn't work on the player.

  This PR addresses both of those issues.

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -925,11 +925,7 @@ void scriptclass::run( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map,
 			else if (words[0] == "flipgravity")
 			{
 				//not something I'll use a lot, I think. Doesn't need to be very robust!
-				if (words[1] == "player")
-				{
-					i=obj.getplayer();
-				}
-				else if (words[1] == "cyan")
+				if (words[1] == "cyan")
 				{
 					i=obj.getcrewman(0);
 				}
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This seems like something that would break compatibility - are we aware of anything that might depend on this behavior?

desktop_version/src/Script.cpp
```diff
@@ -954,8 +950,23 @@ void scriptclass::run( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map,
...
 				}
 
-				obj.entities[i].rule =7;
-				obj.entities[i].tile = 6;
+				if (words[1] == "player")
+				{
+					game.gravitycontrol = !game.gravitycontrol;
+				}
+				else
+				{
```
JayFoxRox: For code readability the `getcrewman` if/else from above could be moved  into this scope.

InfoTeddy:
@flibitijibibo I don't think anyone's deliberately inserting invalid `flipgravity(player)` commands into their level. I ran `grep -io 'flipgravity[(,)]player' more_levels/*.vvvvvv | wc -l`, and it returned 0 results. My `more_levels/` folder contains a huge archive of levels, and according to `ls more_levels | wc -l` it contains 1286 files.

As for using `flipgravity()` on an already-flipped crewmate, there might be one or two levels that have done this accidentally and have forgotten to take it out even though it doesn't work, but I don't know of any. Usually a level creator would realize that it doesn't work and find a workaround through using `gotoroom()`. But unlike `flipgravity(player)`, if `flipgravity()` is accidentally called on an already-flipped crewmate, the effect is purely visual and there's nothing I can think of that depends on crewmates' positions in custom levels (whereas if `flipgravity(player)` was called when it shouldn't be, that would be breaking gameplay).

desktop_version/src/Script.cpp
InfoTeddy: Fixed in 7173fce493453e903fd01133e58c950ef5128d52.

flibitijibibo:
After the style fix it now makes a lot more sense... that was a really good call. And yeah, if existing data never used it, and the single-player data is okay, this should be safe enough.

PR #88, opened 2020-01-16
PR template: switch position of Legal Stuff & Changes

HorayNarea:
## Changes:

this makes the mouse-over tooltip of PRs actually useful instead of showing the same thing for every PR like this:

Compared with this PR:
mouseover2

## Legal Stuff:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,6 +19,12 @@ things for you to read through first:
 If you understand these notes, you can delete the text in this section. Pull
 requests that still have this text will be closed automatically.
 
+
+## Changes:
+
+Describe your patch here!
+
+
 ## Legal Stuff:
 
 By submitting this pull request, I confirm that...
@@ -27,7 +33,3 @@ By submitting this pull request, I confirm that...
   example, a 2.3 update on Steam for Windows/macOS/Linux)
 - [ ] I will be credited in a `CONTRIBUTORS` file for all of said releases, but
   will NOT be compensated for these changes
-
-## Changes:
-
-Describe your patch here!
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -13,6 +13,7 @@ Contributors
 * Marvin Scholz (@ePirat)
 * Keith Stellyes (@keithstellyes)
 * Elijah Stone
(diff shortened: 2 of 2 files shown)


HorayNarea:
I also have a bit of a meta-question:
`CONTRIBUTORS.txt` is `Ordered alphabetically by last name.`

…if this PR is accepted, where should I put my name?
At the end of `S`-last names since `ä` is a non-ascii or at the beginning¹ since the german umlauts can officially be transcribed as
`ä` -> `ae`
`ü` -> `ue`
`ö` -> `oe`
if they are not supported :D

¹ (before or after `Sa…`?) :grinning:

flibitijibibo:
What does `sort` have to say on the ordering? I’ve never tried it in this situation...

Whatever it decides, add yourself in the appropriate spot and I’ll merge this in. I never noticed this since I only ever look at e-mail notifications!

HorayNarea:
> What does `sort` have to say on the ordering? I’ve never tried it in this situation...

That probably depends on the `LC_*` environment variables you have set :grin:

If you are ok with it, I would add me at the end of S…

HorayNarea:
> That probably depends on the `LC_*` environment variables you have set grin

```
$ cat test.txt
Sanger
Saenger
Sänger
Sznger
```

```
$ LC_COLLATE=de_DE.UTF-8 sort test.txt
Saenger
Sanger
Sänger
Sznger
```
```
$ LC_COLLATE=C sort test.txt   
Saenger
Sanger
Sznger
Sänger
```

JayFoxRox:
As a German, I'd still expect you to be listed between `Sa*` and `Sb*` (so a, ä, b, ..). Or directly consider it like an `a`. I wouldn't even consider the `ae` or "add at end" option (it's absolutely not where I'd expect such names).
Note how German names in English also just usually become the same but umlauts become the base letter. That said, I can't wait to see the ordering for when umlauts and accents or other special letters mix.

HorayNarea:
I think using the `LC_COLLATE=C` ordering is the best option since then everyone gets annoyed/offended evenly… :D

Sorry for making this harder then it has to be /o\

flibitijibibo:
I’ve not come across this before so I’m actually really interested in it! But yeah, as someone who’s had to set LC_ALL=C more times than I want to admit, that’s probably the “best” one to use :P

Thanks for the suggestions!

PR #90, opened 2020-01-16
Add contents of CONTRIBUTORS.txt to in-game credits

AllisonFleischer:
## Changes:

Thought it'd be nice for source contributions to be properly acknowledged in the actual game, so I added everyone on CONTRIBUTORS.txt to the main menu credits and the end credits roll.

Updating these is a bit annoying because the positions are hardcoded, but since we already have everyone maintaining CONTRIBUTORS.txt itself it'd only really be necessary to update the in-game credits right before a release. So there's no need to expect new contributors to go through adding to this code.

Also, I shrunk down some code used to handle buttons in the main menu credits to make it easier to read and maintain.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/titlerender.cpp
+++ b/desktop_version/src/titlerender.cpp
@@ -263,6 +263,33 @@ void titlerender(Graphics& dwgfx, mapclass& map, Game& game, entityclass& obj, U
             dwgfx.Print( -1, 130, "Ending Picture by", tr, tg, tb, true);
             dwgfx.bigprint( 40, 145, "Pauli Kohberger", tr, tg, tb, true);
         }
+        else if (game.currentmenuname == "credits_git1")
+        {
+            dwgfx.Print(-1, 20, "GitHub source contributors:", tr, tg, tb, true);
+
+            dwgfx.Print(-1, 40, "Christoph B{hmwalder", tr, tg, tb, true);
+            dwgfx.Print(-1, 50, "Charlie Bruce", tr, tg, tb, true);
+            dwgfx.Print(-1, 60, "Brian Callahan", tr, tg, tb, true);
+            dwgfx.Print(-1, 70, "Dav999", tr, tg, tb, true);
+            dwgfx.Print(-1, 80, "Allison Fleischer", tr, tg, tb, true);
+            dwgfx.Print(-1, 90, "Daniel Lee", tr, tg, tb, true);
+            dwgfx.Print(-1, 100, "Elliott Saltar", tr, tg, tb, true);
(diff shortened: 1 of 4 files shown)


flibitijibibo:
This is a pretty significant change and seems to compound on the hardcodedness of everything - can we limit it to just the ending of the game, while also fixing the hardcoded Y values to draw them out? The menu credits will probably have to stay as they were before since the pages could possibly go on forever, and it's harder for contributors to poke their name in when they get their first patch merged.

flibitijibibo:
Looking at the credits section again, I think I know how to make it more maintainable - I’ll do a different version of this within the next day!

desktop_version/src/titlerender.cpp
```diff
@@ -1428,7 +1456,30 @@ void gamecompleterender(Graphics& dwgfx, Game& game, entityclass& obj, UtilityCl
...
+    if (dwgfx.onscreen(1710 + game.creditposition)) dwgfx.Print(-1, 1710 + game.creditposition, "Brian Callahan", tr, tg, tb, true);
+    if (dwgfx.onscreen(1720 + game.creditposition)) dwgfx.Print(-1, 1720 + game.creditposition, "Dav999", tr, tg, tb, true);
+    if (dwgfx.onscreen(1730 + game.creditposition)) dwgfx.Print(-1, 1730 + game.creditposition, "Allison Fleischer", tr, tg, tb, true);
+    if (dwgfx.onscreen(1740 + game.creditposition)) dwgfx.Print(-1, 1740 + game.creditposition, "Daniel Lee", tr, tg, tb, true);
+    if (dwgfx.onscreen(1750 + game.creditposition)) dwgfx.Print(-1, 1750 + game.creditposition, "Elliott Saltar", tr, tg, tb, true);
+    if (dwgfx.onscreen(1760 + game.creditposition)) dwgfx.Print(-1, 1760 + game.creditposition, "Marvin Scholz", tr, tg, tb, true);
+    if (dwgfx.onscreen(1770 + game.creditposition)) dwgfx.Print(-1, 1770 + game.creditposition, "Keith Stellyes", tr, tg, tb, true);
+    if (dwgfx.onscreen(1780 + game.creditposition)) dwgfx.Print(-1, 1780 + game.creditposition, "Elijah Stone", tr, tg, tb, true);
+    if (dwgfx.onscreen(1790 + game.creditposition)) dwgfx.Print(-1, 1790 + game.creditposition, "Thomas S|nger", tr, tg, tb, true);
```
JayFoxRox: Will this work correctly?

PR #92, opened 2020-01-17
Don't render "[Press ENTER to return to editor]" if "- Press ACTION to advance text -" is also being rendered

InfoTeddy:
## Changes:

* **Don't render "[Press ENTER to return to editor]" if "- Press ACTION to advance text -" is also being rendered**

  This prevents a visual clash if both are being rendered on top of each other and there are no cutscene bars.

  Before:

  press_enter_advancetext_clash_before

  After:

  press_enter_advancetext_clash_after

## Legal Stuff:


TerryCavanagh:
Makes sense! Thanks!

PR #101, opened 2020-01-20
Reset the Direct Mode property of each room in editorclass::reset()

InfoTeddy:
## Changes:

* **Reset the Direct Mode property of each room when `editorclass::reset()` is called**

  There's a long-standing issue where the Direct Mode status of the loaded custom map in memory never resets properly. So if you load a level with a certain layout of Direct Mode rooms, that same layout will be preserved if you start making a new level in the editor. This PR fixes that issue.

## Legal Stuff:


TerryCavanagh:
Thanks!

PR #103, opened 2020-01-22
Fix bug where translucent pixels were drawn badly

InfoTeddy:
## Changes:

* **Fix the bug with translucent pixels being drawn incorrectly**

  For a long time, VVVVVV has had an issue where if you used custom graphics with translucent pixels (i.e. pixels whose opacity was neither 0% nor 100%, but somewhere between 0% and 100%), those pixels would end up looking very ugly. They seemed to be overlaid on top of some weird blue color, instead of actually being translucent.

  This bug only occurred when in-game and not in towermode. So, most of the time, basically. The translucent pixels only displayed properly inside the Tower, both minitowers, and the editor (when not playtesting).

  I traced this down to the use of the magic number 0xDEADBEEF in `Graphics::drawmap()` and `Graphics::drawfinalmap()`. Looks like someone had fun finding a color. Was it you, simoroth? Anyway, I've changed it so it simply uses 0 instead. Note that this magic number needs to be changed for BOTH `FillRect()` and `OverlaySurfaceKeyed()`, or else the game's background will be some random solid color instead of actually being drawn properly.

  Here are pictures for both before and after this change so you can see the problem that this PR fixes:

  **Before:**
  How translucent pixels were drawn before the bug was fixed.

  **After:**
  How translucent pixels were drawn when the bug got fixed.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -2342,7 +2342,7 @@ void Graphics::drawmap( mapclass& map )
     ///TODO forground once;
     if (!foregrounddrawn)
     {
-        FillRect(foregroundBuffer, 0xDEADBEEF);
+        FillRect(foregroundBuffer, 0x00000000);
         if(map.tileset==0)
         {
             for (j = 0; j < 29+map.extrarow; j++)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I guess this was meant to be like XNA where if you saw the ugly clear color you didn’t properly draw to the frame, but I think this game is done with that feature now... it might be useful to clear to a terrible debug color for testing new graphics? That can come later if needed.

PR #104, opened 2020-01-22
Fix an out-of-bounds array access in findstartpoint()

InfoTeddy:
## Changes:

* **Fix out-of-bounds array indexing in `editorclass::findstartpoint()`**

  Out-of-bounds array access is Undefined Behavior, which means Bad Things.

  In this particular case, it was indexing an array by using the `testeditor` variable. Which is fine, except it was indexing that array *in a conditional that only happens if `testeditor` is -1*. So it was indexing an array at position -1, which is Out of Bounds and is Not Good.

## Legal Stuff:


flibitijibibo:
I still want to know why that access was there at all, but oh well...

PR #108, opened 2020-01-22
Remove compilation warnings where possible.

kxtcd950:
## Changes:

Removes compilation warnings, except for a couple of which are genuine (I suspect) bugs, or which will impact cross-platform functionality.
Now if warnings are introduced, the code can be seen to be incorrect a lot easier.

Tested compilers: MSVC2019, MSVC2010.

## Legal Stuff:

- [Y] My changes may be used in a future commercial release of VVVVVV (for
- [Y] I will be credited in a `CONTRIBUTORS` file for all of said releases, but


flibitijibibo:
Static Cast: The Pull Request!

This is a lot of work for what's supposed to be nothing different, functionally... is it possible to just suppress the warning instead? The templates in particular make me nervous, but any template tends to do that. The unused parameter fixes and brace additions are definitely good though, I'd be up for taking those separately while the rest of the C++ "stuff" finishes arguing with itself.

The only thing I'd definitely cut is the i/j/k changeset - the original variables are awful, but that's going to need a whole separate audit for sure.

kxtcd950:
The templates are a naughty way of fixing a larger problem in the codebase; notably that most of the colour parameters are specified as integers, but are passed as flaoting point values because of fades.
The other ways of doing this would be to either put static_cast<>() all over the place in the code, making the code readability (in my opinion) generally worse; believe it or not, I've tried to tread as light as I could in the codebase to keep disruption down to a minimum.  
Otherwise, the alternative is to instantiate all the various call signatures for those functions which take r,g,b values with int and float where appropriate, and quite honestly, I thought getting the compiler to do that grunt work was a better plan; it's more likely to get it correct far quicker than I could...

I realise this is a *huge* change for no net functionality; but when I came to look at the codebase and ran through a compiler, I wondered what the morass of warnings was hiding.  There are a couple of instances in the code where the warnings are quite justified (from memory, Script.cpp uses atoi a lot to turn strings into integers) where the warning is actually warning about something that's not just noise - string variables are being assigned from integer values.

I did perform the various classes of warning removal (largely) all separately and committed each as I went along; I did forsee that some of the changes wouldn't be necessarily seen to be desirable.  This approach should make cherry picking specific types of fix an option, rather than an all-or-nothing.  However, I can't guarantee that the changesets absolutely contain that type of fix and only that tye of fix; the odd missed one from a past set (or future set for that matter) may cross-contaminate.  These will be in the singles though; manually removing the singletons shouldn't be an issue.

i/j/k is the changeset I'm least happy with; with the values of i/j/k being saved in the class in the original code, and the values of i/j/k being used - in places - without being initialised, there's always the danger of state being lost and bugs introduced.

I'm still slightly miffed that the mac compiler doesn't like lambdas.  Need another solution to the ::toupper problem (which is only a warning in the 2019 compiler anyway)...

InfoTeddy:
There's a glitch some custom levels have used that directly depends on the i/j/k stuff. They use it to spawn their own custom activity zone with `createactivityzone()` by exploiting `everybodysad()`, because both commands use the same variable `i`.

InfoTeddy:
Also, looks like the CI build on macOS fails due to your use of magical syntax:
```
../src/editor.cpp:2954:93: error: expected expression
                            std::transform(tstring.begin(), tstring.end(), tstring.begin(), [](char x)->char { return static_cast<char>(::toupper(x)); });
                                                                                            ^
1 error generated.
```

InfoTeddy:
For what it's worth, whenever I compile on Linux, I don't get any warnings. Seems like the approach to just suppress the warnings was taken there.

TerryCavanagh:
> 
> 
> There's a glitch some custom levels have used that directly depends on the i/j/k stuff. They use it to spawn their own custom activity zone with `createactivityzone()` by exploiting `everybodysad()`, because both commands use the same variable `i`.

This right here is the final word on accepting this pull request, sorry. Quite a lot of attempts to clean up VVVVVV's source code are going to run into this issue - I don't want to break any work from any of the people who've made player levels through the years, a lot of which now depends on exploiting bugs. Maybe consider contributing to the community edition instead? https://github.com/v6cord/VVVVVV-CE

leo60228:
VVVVVV-CE has the same stance on this. We go out of our way to keep this bug intact, despite refactored code.

PR #117, opened 2020-01-24
Add a null terminator to loaded TinyXML files

FredrIQ:
## Changes:

The TinyXML parse() function expect a C-like string, including terminator.
When xml loading was changed, it loaded the file, but included no such thing.
Thus, we load the file, then reallocate the memory so that we can insert a
null terminator to it, before passing it to parse().

## Legal Stuff:

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -165,11 +165,17 @@ bool FILESYSTEM_loadTiXmlDocument(const char *name, TiXmlDocument *doc)
 {
 	/* TiXmlDocument.SaveFile doesn't account for Unicode paths, PHYSFS does */
 	unsigned char *mem = NULL;
-	FILESYSTEM_loadFileToMemory(name, &mem, NULL);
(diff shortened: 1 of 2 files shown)


Dav999-v:
Wouldn't it be better to just let `FILESYSTEM_loadFileToMemory` `malloc` the extra byte, while keeping `len` in that function the same? If you load binary data, it doesn't matter if there's a null at the end since it's beyond the buffer. I'm not sure how efficient `realloc` is, especially when level files can easily be 5 megabytes and hundreds of them can need to be loaded for the levels list.

flibitijibibo:
Was just about to say, this might be a good enough reason to duplicate the loadToMemory function to include the extra byte (documenting the reason of course).

FredrIQ:
Is this better? Instead of code duplication or adding a null pointer unconditionally, I just added an argument to the loadFileToMemory function for whether or not to append a null terminator.

flibitijibibo:
Good enough for me - it's not a hot path so the bool isn't a big deal. Thanks for the fix!

PR #120, opened 2020-01-24
Fix level editor not using LoadTiXmlDocument

FredrIQ:
## Changes:

This fixes the level editor not using LoadTiXmlDocument,
and thus retaining the null terminator issue as addressed previously.

## Legal Stuff:


flibitijibibo:
Oof, good catch on the pointer deref, totally missed that.

PR #121, opened 2020-01-24
Fix tower quicksave bug (only appeared in some builds)

FredrIQ:
## Changes:

I ran the game through Valgrind to catch various issues.
One thing caught my attention -- map.resumedelay. This is
used by The Tower/etc to add a delay before the game is
resumed (probably for camera controls delaying it) for
spawning the player. This variable was uninitialized. Notably,
if I explicitly set it to 592851 or similar, it reproduces the
bug, even if I cannot reproduce it with my typical compilation
flags. So fixing this by initializing it to 0, alongside some
other Valgrind warnings, should fix the problem entirely.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -51,6 +51,7 @@ void entityclass::init()
     nentity = 0;
     nblocks = 0;
 
+    skipblocks = false;
     skipdirblocks = false;
     platformtile = 0;
     customplatformtile=0;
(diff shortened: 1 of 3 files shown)


FredrIQ:
Build failure appears to be on GitHub side

flibitijibibo:
Yeah, the CI has been having a fit all evening so ignore that... The only thing it might complain about once it’s back is the Entity initialization, can we put that in the constructor instead?

FredrIQ:
Ah, I probably overlooked that function. Changed.

PR #122, opened 2020-01-25
Add an accessibility option to be able to see through the room name background

InfoTeddy:
### Summary

This PR adds an option to be able to make the room name background translucent.

Such an option would be very useful when playing through custom levels, since certain custom levels have problems with making their visuals clear with regards to the roomname. They might put collision behind it, making it look like there's an exit and you'd go through to the room below, even though they intend for you to magically know that there's collision there and then walk on it. Or they might just be mean and decide to hide spikes there.

**With an opaque room name background:**
You can't tell which one is real.

**With a translucent room name background:**
Now you can tell.

### Details

* **Render room name backgrounds using an `SDL_Surface` and `SDL_BlitSurface()` instead of using `SDL_FillRect`**

  Unfortunately, it would be easy if we could simply change the `FillRect()` to use a translucent color. However, it seems like you cannot draw translucent rectangles this way. Someone should complain to the SDL people...

  So what I have to do instead is to make a new `SDL_Surface` named `dwgfx.footerbuffer`, and then use `SDL_BlitSurface()` to draw that. Then I can simply set the opacity with `SDL_SetSurfaceAlphaMod()`!

* **Add the accessibility option to make room name backgrounds translucent**

  This is controlled by the variable `dwgfx.translucentroomname`, and is stored in `unlock.vvv` as the XML tag `<translucentroomname>`.

  If it is true, then room names are drawn with opacity 127, which is between 0 and 255, and is thus half of opaque opacity.

* **Draw room names with text outline**

  Having a text outline on the room names makes it easier to see them when you have a translucent room name background.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -3442,15 +3442,13 @@ void editorrender( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map, ent
                 if(ed.tiley<28)
                 {
                     if(ed.roomnamehide>0) ed.roomnamehide--;
-                    FillRect(dwgfx.backBuffer, 0,230+ed.roomnamehide,320,10, dwgfx.getRGB(0,0,0));
-                    dwgfx.Print(5,231+ed.roomnamehide,ed.level[ed.levx+(ed.maxwidth*ed.levy)].roomname, 196, 196, 255 - help.glow, true);
                 }
                 else
                 {
                     if(ed.roomnamehide<12) ed.roomnamehide++;
-                    FillRect(dwgfx.backBuffer, 0,230+ed.roomnamehide,320,10, dwgfx.getRGB(0,0,0));
-                    dwgfx.Print(5,231+ed.roomnamehide,ed.level[ed.levx+(ed.maxwidth*ed.levy)].roomname, 196, 196, 255 - help.glow, true);
                 }
+                FillRect(dwgfx.backBuffer, 0,230+ed.roomnamehide,320,10, dwgfx.getRGB(0,0,0));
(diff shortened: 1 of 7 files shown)


flibitijibibo:
FillRect is meant to be more like memset, so it should be a fast path when blending isn’t being used. Would it be possible to keep using FillRect for the non-alpha path, for performance?

InfoTeddy:
Fast path added in 10a44071dcaf0477b5aa01c528ef1e511d3240b3.

desktop_version/src/editor.cpp
```diff
@@ -3442,15 +3442,21 @@ void editorrender( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map, ent
...
                     if(ed.roomnamehide>0) ed.roomnamehide--;
-                    FillRect(dwgfx.backBuffer, 0,230+ed.roomnamehide,320,10, dwgfx.getRGB(0,0,0));
-                    dwgfx.Print(5,231+ed.roomnamehide,ed.level[ed.levx+(ed.maxwidth*ed.levy)].roomname, 196, 196, 255 - help.glow, true);
                 }
                 else
                 {
                     if(ed.roomnamehide<12) ed.roomnamehide++;
+                }
+                dwgfx.footerrect.y = 230+ed.roomnamehide;
+                if (dwgfx.translucentroomname)
```
flibitijibibo: Should this part be in the branch above where the FillRect+Print used to be?

InfoTeddy: Well, I'm setting the y-position of `dwgfx.footerrect`, because in the branch where `dwgfx.translucentroomname` is true, I use the y-position of that rect when I pass the rect in the fourth argument to `SDL_BlitSurface()` - `SDL_BlitSurface(dwgfx.footerbuffer, NULL, dwgfx.backBuffer, &dwgfx.footerrect);`.

Thinking about it now, though, I could at least move it into the if-branch, so it's clearer. Doing that now.

PR #124, opened 2020-01-26
Remove from M&P the room data of Final Level/Lab/Overworld/Tower

InfoTeddy:
## Changes:

This is to make sure that the room data of those areas are not distributed with the M&P binary, even if they are already inaccessible because of other M&P ifdefs (I tested).

## Legal Stuff:


flibitijibibo:
Always happy to have smaller binaries!

PR #125, opened 2020-01-27
Draw editor entities according to ingame order

FredrIQ:
Ingame entities are drawn backwards, probably to draw the player on top,
being entity 0 (usually, at least). Make the level editor draw entities
in the same order.

## Legal Stuff:


flibitijibibo:
Seems harmless enough... I’m not aware of any specific reason why it’s like this but I can see how it’d make editing annoying.

PR #126, opened 2020-01-27
Fix and improve handling of cancelling roomtext/scripttext input

FredrIQ:
First, two bug fixes. Room text input mode wasn't properly unset
upon pressing Esc, making the prompt get stuck, requiring you to
add roomtext again and finish it to make it go away. Secondly,
escaping script text input would remove the wrong entity.

I also tweaked the handling slightly so that instead of deleting
the entity if it already existed if escaping from text input,
it merely reverts the change in script name/roomtext to what it
was previously.

I considered refactoring the editor text input handler entirely,
but figured such a change would be a bit too extensive for the
purpose of this repository.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -260,7 +260,7 @@ void editorclass::reset()
     boundy2=0;
 
     scripttextmod=false;
-    scripttextent=0;
+    textent=0;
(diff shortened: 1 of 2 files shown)


TerryCavanagh:
Ah, this is great! Thanks @FredrIQ :)

PR #127, opened 2020-01-27
Don't re-draw horizontal and vertical warp backgrounds if gotoroom()ing to the same room in custom levels

InfoTeddy:
### Summary

**Don't redraw horizontal and vertical warp backgrounds if `gotoroom` is being called on the current room in custom levels**

This has two benefits:
 1. The game uses less resources when it is asked to `gotoroom` to the same room from a `gotoroom` script box, because it is no longer re-drawing the warp background every single frame, which is very wasteful.
 2. The warp background no longer freezes or flickers if the player is standing inside a `gotoroom` script box (which calls `gotoroom` every frame or every other frame, because every time the `gotoroom` happens the script box gets reloaded, thus doing another `gotoroom`, etc.)

To demonstrate the fix, I have attached a custom level that has `gotoroom` script boxes so you can compare before and after. Before, the game is re-drawing the warp background every other frame, which is very wasteful, and causes a visual flicker which is jarring and distracting. After, the game no longer re-draws the warp background every other frame, which takes up less resources and also fixes the visual bug.

* [gotoroom_warp_bg_fix_demo.zip](https://github.com/TerryCavanagh/VVVVVV/files/4119737/gotoroom_warp_bg_fix_demo.zip)

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Game.h
+++ b/desktop_version/src/Game.h
@@ -121,6 +121,7 @@ public:
     int door_up;
     int door_down;
     int roomx, roomy, roomchangedir;
+    int prevroomx, prevroomy;
     int temp, j, k;
 
     int savex, savey, saverx, savery;
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -1059,8 +1059,17 @@ void mapclass::gotoroom(int rx, int ry, Graphics& dwgfx, Game& game, entityclass
 	loadlevel(game.roomx, game.roomy, dwgfx, game, obj, music);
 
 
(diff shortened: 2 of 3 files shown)


TerryCavanagh:
This seems reasonable enough, though I don't really understand the purpose of calling gotoroom in this way! I assume internal scripting trickery? :D

Anyway, can't see what harm this does for regular content, so happy to merge it.

InfoTeddy:
> I don't really understand the purpose of calling gotoroom in this way! I assume internal scripting trickery? :D

Yeah, pretty much. Using `gotoroom` like this is a part of the way to making a `delay()` that doesn't lock up player controls!

PR #129, opened 2020-01-29
Allow using 'help', 'graphics', 'music', 'game', 'key', 'map', and 'obj' everywhere

InfoTeddy:
## Changes:

* **Allow using `help`, `graphics`, `music`, `game`, `key`, `map`, and `obj` everywhere**

  This PR makes `help`, `graphics`, `music`, `game`, `key`, `map`, and `obj` essentially static global objects that can be used everywhere. This is useful in case we ever need to add a new function in the future, so we don't have to bother with passing a new argument in which means we have to pass a new argument in to the function that calls that function which means having to pass a new argument into the function that calls *that* function, etc. which is a real headache when working on fan mods of the source code.

  Note that this changes *none* of the existing function signatures, it merely just makes those variables accessible everywhere in the same way `script` and `ed` are.

  Also note that some classes had to be initialized after the filesystem was initialized, or else the game would segfault. But C++ would keep initializing them before the filesystem got initialized, because I *had* to put them at the top of `main.cpp`, or else they wouldn't be global variables.

  The only way to work around this was to use entityclass's initialization style (which I'm pretty sure entityclass of all things doesn't need to be initialized this way), where you actually initialize the class in an `init()` function, and so then you do `graphics.init()` after the filesystem initialization, *after* doing `Graphics graphics` up at the top.

  I've had to do this for `graphics` (but only because its child GraphicsResources `grphx` needs to be initialized this way), `music`, and `game`. I don't think this will affect anything. Other than that, `help`, `key`, and `map` are still using the C++-intended method of having ClassName::ClassName() functions.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -113,9 +113,9 @@ bool GetButtonFromString(const char *pText, SDL_GameControllerButton *button)
 }
 
 
-Game::Game(void):
-    mutebutton(0)
+void Game::init(void)
 {
+    mutebutton = 0;
     infocus = true;
     paused = false;
     muted = false;
--- a/desktop_version/src/Game.h
+++ b/desktop_version/src/Game.h
@@ -17,7 +17,7 @@ class musicclass;
 class Game
 {
 public:
-    Game(void);
+    void init(void);
     ~Game(void);
 
 
@@ -361,5 +361,7 @@ public:
 
     bool skipfakeload;
 };
+
+extern Game game;
 
 #endif /* GAME_H */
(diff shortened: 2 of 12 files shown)


TerryCavanagh:
Huh! Yeah, this seems like a useful change, and a good first step towards eventually stabilising this jenga tower of a game (i.e. removing the reference passing stuff from virtually every function in the game).

I can't think of any reason this might cause a problem - any thoughts, @flibitijibibo?

flibitijibibo:
None I can think of - may be worth checking the Win64 issue after this, it might fix that as a side effect.

PR #134, opened 2020-01-31
Fix text() centering with x=-1

InfoTeddy:
## Changes:

* **Fix `text()` centering with x=-1**

  The game uses magic values x=-500 and y=-500 to indicate when a text box should be centered horizontally or vertically. It does this for x=-1 too, but it's buggy because it only looks at the first line of the text box to center it. In this commit I fix it so that it will look at all of the lines of the text box to center it instead.

  Here's what x=-1 centering looked like before:
  x=-1 centering: before

  Here's what x=-1 centering looks like after:
  x=-1 centering: after

  The script is:

  ```
  text(gray,-1,-500,2)
  this is
  a text box with two lines
  speak
  endtext
  ```

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -597,7 +597,7 @@ void scriptclass::run( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map,
 					dwgfx.textboxmoveto(textx);
 				}
 
-				if (textx == -500)
+				if (textx == -500 || textx == -1)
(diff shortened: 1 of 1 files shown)


InfoTeddy:
The CI build failed on Linux because of downloading something from Microsoft's servers and having it be an unexpected file size. A bit annoying.

TerryCavanagh:
Yep, good fix! Thank you 👍

PR #138, opened 2020-02-02
Remove compiler warnings

kxtcd950:
## Notes

This is go number two - return of static_cast<> - guided by feedback from the previous attempt;  hopefully this change set will be less contentious than the previous one.  
This change set doesn't have the i/j/k/temp triggered warnings removed, just suppressed by the compiler (MSVC only) I'm assuming MAC/Linux compiler suppression is handled differently but don't have a compiler to test with for those platforms.  Windows warning suppression tested with VS2010 and VS2019 compilers.  
This change set also handles the ::toupper|::tolower warning removal differently to - hopefully - avoid the Mac compiler failure of the previous change set.  I'm hoping for a successful compile across the board this time.

## Changes:

Removed compiler warnings by static_cast (and judicious use of templates where appropriate; but all template functions go though static_cast anyway, so compiler will error if a cast to the underlying function type isn't available).

I've not touched the i/j/k/temp issue because it removes bugs which are historic and required for backward compatibility with old levels and/or game-play which exploit these bugs. As a result, some warnings have been suppressed on the compiler's command line. These are suppressed as they are all to do with the i/j/k/temp issue; new code shouldn't be triggering these warnings - i/j/k/temp re-use is likewise verboten as it will also interfere with backwards compatibility, and new code shouldn't use this code pattern anyway...

The only two compiler warnings which remain are believed to be genuine bugs; probably easy to fix, but given the size of the change set, they've been left for more qualified people to inspect and fix.

## Legal Stuff:

- [ ✓ ] My changes may be used in a future commercial release of VVVVVV (for
- [ ✓ ] I will be credited in a `CONTRIBUTORS` file for all of said releases, but

### DIFF
--- a/desktop_version/src/Entity.h
+++ b/desktop_version/src/Entity.h
@@ -242,4 +242,6 @@ public:
     int customcrewmoods[6];
 };
 
+extern entityclass obj;
+
 #endif /* ENTITY_H */


flibitijibibo:
Is it possible to separate these out into individual commits like the original changeset? I still like the unused param fix, for example, but I’m still not into the templates (seems like it adds just as much code as actually fixing the casts, but with a massive output binary size). I’d like to be able to cherry pick the goof stuff and leave the scary stuff for later.

kxtcd950:
I'll separate them out; I understand where you're coming from, and I wish I'd done it like that originally now...
The templating should only add functions which need to be instanced, and given that the instance is always only one call to the original function with casts to sort out parameters, it should only add 8-10 machine-size words per instance (the call plus stack pre- and post- amble).  Which for "createinstance" is 8 instances, so 80 machine words inflation of binary size.  Templates get a bad rap...
I'm not sure how to go about gitting them apart; I may be gone some time.  If you want to close this, I'll add a new pull request when I've separated them out though.

flibitijibibo:
We can leave this open so people know it's being worked on - this isn't critical so take whatever time you need.

(Also I probably should have referred to the `-g` size, hopefully the optimized output isn't too bad... but I've definitely made good money on stepping through templates in GDB though!)

kxtcd950:
I don't remember closing this, sorry; the force push probably did it, but I had to do that to remove some existing commits so that I could unwind them individually (I'm not too familiar with github, we use TFS at work and it's all different).

Quick style guide clarification required:

I'm more than happy (having worked out some handy macros) to go through and fix-up parameters to functions which are of the wrong type by modifying the call.  However, this - to my mind - is deleterious to the readability of the code (although does indicate through the liberal sprinkling of static_casts that there may be something off).
Essentially I think I'm asking whether you think that the parameter ambiguity issue that is extant will - at some point - get sufficiently annoying to maintainers that they will get fixed, or whether any effort put into any further cleanup (be it template magic or static_cast sprinkling) along those lines will just be indefinitely parked?

InfoTeddy:
What are the two remaining warnings you talk about? I'm not getting them on Linux and I don't want to try compiling on Windows (Windows is not a good place for dev environments).

kxtcd950:
The two remaining warnings (which are template warnings) are only flagged in later - more standards conformant? - compilers. These two are where the result of an ::atoi() call has been assigned to a std::string typed variable. These two remaining warnings are the only two warnings from the entire codebase which I believe to be unintended bugs, and are both - from memory - to do with the “hardestroom” variable in Game.cpp.

InfoTeddy:
Looks like with the force-push, what you did was remove all commits from the branch. So GitHub closed the PR automatically.

PR #143, opened 2020-02-03
Add and draw one more row to all rooms with room names

InfoTeddy:
### Summary

Since translucent roomname backgrounds were introduced in TerryCavanagh/VVVVVV#122, it exposes one glaring flaw with the game that until now has been kept hidden: in rooms with room names, the game cheapens out with the tile data and doesn't have a 30th row, because the room name would hide the missing row. As a result, rooms with room names have 29 rows instead of 30 to fill up the entire screen. And it looks really weird when there's nothing but empty space behind the translucent room name background.

Translucent room name background, missing 30th row

To remedy this, I added one row to each room with a room name in the level. First, I had to filter out all the rooms with no room names. However, that's actually all contained in `Otherlevel.cpp`, the Overworld, which contains 221 rooms (8 of which are the Secret Lab, 6 more of which are the Ship, so 207 are the actual Overworld, right? Wrong, 2 of those Overworld no-roomname rooms are
in the Lab, so there are actually 205 Overworld rooms). The remaining level data files all contain rooms with room names.

But the process wasn't that easy. I noticed a while ago that each room contains 29 `tmap.push_back()`s, one for each row of the room, and each row is simply a string containing the 40 tiles for that row, concatenated with commas.

However, I decided to actually check my intuition by doing a grep on each level file and counting the number of results, for example `grep 'push_back' Labclass.cpp | wc -l`. Whatever number comes out should be divisible by 29. That particular grep on `Labclass.cpp` returns 1306, which divided by 29 is 45 with a remainder of 1.

So what does that mean? Does that mean there's 45 rooms each, and 1 leftover row? Well, not exactly. The extra row comes from the fact that Outer Space has 30 rows instead of 29. Outer Space is the room that comes up when the game finds a room is non-existent, which shouldn't happen with a properly-working game, except in Outside Dimension VVVVVV. In fact, each level file has their own Outer Space, and every single Outer Space also has 30 rows. So really, this means there are 44 rooms in the Lab and one Outer Space room. (Well, in reality, there are 46 rooms in the Lab, because 2 of them use the Outside tileset but have no room names, so they're stored in `Otherlevel.cpp` instead.)

We find the same result for the Warp Zone. `grep 'push_back' WarpClass.cpp | wc -l` returns 697, which is 24 remainder 1, meaning 23 rooms of 29 rows and 1 room of 30 rows, which corresponds with 23 rooms in the Warp Zone and one Outer Space room.

However, Outside Dimension VVVVVV + Tower Hallways and Space Station 1 and 2 are both odd curiosities. `Finalclass.cpp` contains Outside Dimension VVVVVV, (which is Intermission 1 and 2 and the Final Level), but also the Tower Hallway rooms, i.e. the auxiliary Tower rooms that are not a part of the main tower. `Spacestation2.cpp` contains both Space Station 1 and 2, so don't be deceived by the name.

`grep 'push_back' Finalclass.cpp | wc -l` returns 1597, which is actually 55 remainder 2. So... are there two rooms with 30 rows? Yes, in fact, The Gravitron and Outer Space both contain 30 rows. So there are actually 55 rooms stored in `Finalclass.cpp` (not including the minitowers Panic Room and The
Final Challenge), 54 rooms of actual level data and one Outer Space room, and breaking down the 54 rooms even further, 51 of them are actually in Outside Dimension VVVVVV and 3 of them are Tower Hallways. Of the 51 Outside Dimension VVVVVV rooms, 14 of those are Intermission 1, 4 of them are Intermission 2, and the rest of the 33 rooms are the Final Level (again, not including the minitowers).

`grep 'push_back' Spacestation2.cpp | wc -l` returns 2148, which is 74 remainder 2. Are there two rooms with 30 rows again? No; one of those counted 2148 rows is a false-positive, because there's an if-else in Prize for the Reckless that replaces the row with spikes with a row without spikes if you are in a time trial or in No Death Mode. So there's 73 rooms in Space Station 1 and 2, and one Outer Space room.

With all this in mind, I decided to duplicate the current last row of each room, the 29th row, to add a 30th row. However, I wasn't going to do this automatically! But neither was I going to write some kludge-y code to parse each nightmare of a level file and duplicate the rows that way.

Enter: Vim macros! (Er, well, actually, I use Neovim.) I first did `/push_back`, so that pressing `n` would keep going to the next `push_back` in the file. Then I went to the 29th row of the first room in the file, did a `Yp`, and then started my macro with `qq`. The macro went like this: `30nYp`, which is simply going to the 29th row of the next room over and duplicating it. And that's all there was to it. However, I had to make sure that (1) my cursor was before the `push_back` on the line of the 29th row of the room, and (2) that I didn't skip rooms, both of which were problems I encountered when pressing Ctrl+Z on a given invocation of the macro (the Ctrl+Z is just a metaphor, you actually undo by typing `u` in Vim). And also I had to make sure to be careful around the extra lines of `push_back`s in Prize for the Reckless and The Gravitron, and make sure I didn't run past the end of the file and loop back around. Thankfully, all Outer Space rooms are at the end of each file.

But first, I had to increase the number of rows drawn in `Graphics.cpp` by 1 in order to compensate for this, and do the same when reading the tile data in `Map.cpp`. I had to change `fillcontent()`, `drawmap()`, `drawfinalmap()`, `drawtowermap()`, and `drawtowermap_nobackground()`. Funnily enough, the tower functions already used 30 rows, but I guess it's an off-by-one due to the camera scrolling, so they now draw 31 rows each.

Then, I went in-game to make sure that the row behind each room name looked fine. I checked *every* single room with a room name. I turned on invincibility mode and added a temporary line to `hardreset()` that always turned on `game.nocutscenes` for a smoother playtesting experience. And to make sure that rooms which have entirely empty bottom rows actually still have 30 rows, instead of having 29 and the game assuming that the 30th row was empty (because that sounds like it could lead to Undefined Behavior), I added this temporary debugging line to the start of `mapclass::fillcontent()`:

```
printf("(%i,%i) has %i rows\n", game.roomx, game.roomy, (int) tmap.size());
```

Everywhere I checked - and I made sure to check all rooms - every room had 30 rows and not 29 rows.

Unfortunately, some rooms simply couldn't be left alone with their 29th row duplicated and had to be manually edited. This was because the 29th row would contain some edge tiles because the player would be able to walk somewhere on the 28th, 27th, and 26th rows, and if you duplicated said edge tiles behind the room name, it would look bad.

Bad-looking duplicated row in Comms Relay

Here's a list of rooms whose 30th rows I had to manually edit:

 - Comms Relay
 - The Yes Men
 - Stop and Reflect
 - They Call Him Flipper
 - Double-slit Experiment
 - Square Root
 - Brought to you by the letter G
 - The Bernoulli Principle
 - Purest Unobtainium
 - I Smell Ozone
 - Conveying a New Idea
 - Upstream Downstream
 - Give Me A V
 - $eeing Dollar $ign$
 - Doing Things The Hard Way
 - Very Good
 - Must I Do Everything For You?
 - Now Stay Close To Me...
 - ...But Not Too Close
 - ...Not as I Do
 - Do Try To Keep Up
 - Whee Sports
 - As you like it

   This is actually a strange case where it looked bad because of the 29th row, instead of the 30th row, and I had to change the 29th row instead of the 30th row to fix it.
 - Maze With No Entrance
 - Ascending and Descending
 - Mind The Gap

   Same strange case as "As you like it" (it's the 29th row I had to change that was the problem, not the 30th).
 - 1950 Silverstone Grand V
 - The Villi People

I found that Panic Room and The Final Challenge also looked strange behind the roomname background, but I can't do much about either because towers' tile data wrap around at the top and bottom, and if I added another row to either it would be visible above the room name.

I've considered updating the development editors with these new level tiles, but I decided against it as the development editors are already pretty outdated anyway.

With all that said, the rooms will now look good if you have a translucent room name background, and will look like this now:

Fixed 30th row in Teleporter Divot

## Legal Stuff:


TerryCavanagh:
This one is a substantial change - the shorter room lengths was a deliberate thing, and I'm worried about how this will affect timing or the feel of certain parts of the game. I also don't really see what it solves, if I'm honest. (Regarding https://github.com/TerryCavanagh/VVVVVV/pull/122, it's not a big deal, but I'm not sure if I would have merged that one personally - custom levels *shouldn't* be using the bottom row to place content, and if they are it's kinda just bad design.)

I'm going to leave this open until I have time to rebuild the game and try it out with the change, which could be a while - have a fairly packed schedule right now.

(Also, this feels as good a place to say this as anywhere: I've noticed, and really appreciate your consistently good contributions. Thank you very much!)

InfoTeddy:
Ok, and I appreciate the thanks.

leo60228:
Vertex Vortex, an included player level, does this to hide a trinket. I'm not sure if that makes this a better or worse change, though.

InfoTeddy:
Rebased onto latest master and force-pushed. The description is a little outdated now since the game no longer uses `tmap.push_back()`s, but whatever.

I think this change is basically purely visual and shouldn't affect gameplay at all. I mean, you basically never touch the bottom row of the room, right? It won't affect collision either, because in 29-row rooms the game will check for collision on the 30th row by checking the 29th row (`if (y == 29+extrarow) return collide(x, y - 1);` in `mapclass::collide()`). Adding tiles there doesn't affect anything, even if it was all completely solid.

TerryCavanagh:
What I'm not sure about, and specifically wanted to check, is what the extra row would do to timing on rooms like e.g. `Please Forgive Me!`. Again, I don't *really* see what this pull request solves - it seems like the best case scenario here is that it maybe doesn't break anything!

InfoTeddy:
I'm pretty sure this doesn't break anything. The amount of rows in a room don't affect how much you're moved whenever you transition between rooms, and even if it did, I haven't changed `map.extrarow` at all - whatever it controls, it will still control. (E.g. the timer for bringing up/down the pause screen.) All I've done is make it no longer control how many rows the game renders, which is only visual.

TerryCavanagh:
`I haven't changed map.extrarow at all - whatever it controls, it will still control.`

Ah, ok, that satisfies my concerns here. If this is purely graphical, then it's fine to merge. Thanks!

PR #146, opened 2020-02-04
Render screen shaking in towers

InfoTeddy:
## Changes:

* **Render screen shaking in towers**

  It's really obvious that screen shaking is not processed in towers if you bring up the pause menu then quickly quicksave and bring it back down. The screen won't shake, but it will suddenly start shaking if you exit the tower, finishing off the stalled screenshake timer.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/titlerender.cpp
+++ b/desktop_version/src/titlerender.cpp
@@ -2933,7 +2933,15 @@ void towerrender(Graphics& dwgfx, Game& game, mapclass& map, entityclass& obj, U
         dwgfx.flashlight();
     }
 
-    dwgfx.render();
+    if (game.screenshake > 0 && !game.noflashingmode)
+    {
+        game.screenshake--;
+        dwgfx.screenshake();
+    }
+    else
+    {
+        dwgfx.render();
+    }
 }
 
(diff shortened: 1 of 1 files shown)


desktop_version/src/titlerender.cpp
```diff
@@ -2933,7 +2933,15 @@ void towerrender(Graphics& dwgfx, Game& game, mapclass& map, entityclass& obj, U
         dwgfx.flashlight();
     }
 
-    dwgfx.render();
+    if (game.screenshake > 0 && !game.noflashingmode)
+    {
+        game.screenshake--;
```
flibitijibibo: Does this need to be decremented even if noflashingmode is on? Wondering if this would cause a screenshake to build up until the mode was disabled, triggering a seemingly random screenshake.

InfoTeddy: This is the exact same code used everywhere else, and is also used for screen flashing (making the screen white).

It might be a good idea to decrement both the shake and flash timers even if screen effects are disabled, but that would be another PR. Same thing with resetting screen shaking and flashing when exiting to the menu.

flibitijibibo: Makes sense, as long as it’s consistent!

PR #147, opened 2020-02-04
Make hardest room "Dimension VVVVVV" if it has no room name

InfoTeddy:
## Changes:

* **Make the hardest room be named "Dimension VVVVVV" if it has no room name**

  The only places where you can die in a room without a room name is the Overworld, and I feel that calling it Dimension VVVVVV is appropriate.

  You can't naturally die in The Ship nor the Secret Lab, and you can only do it by pressing R, so I didn't feel it appropriate to add checks to make the hardest room be "The Ship" or "The Secret Lab" if you managed to get your hardest room to be a room in either of those areas.

## Legal Stuff:


TerryCavanagh:
Makes sense to me! Thanks 👍

PR #149, opened 2020-02-05
Make tiles in tower mode consistent with tower tileset elsewhere

FredrIQ:
Previously, in tower mode, being inside walls would just kill you, unlike
being inside walls outside tower mode, which was somewhat confusing.

Also, spikes behaved differently with regards to invincibility, being
unsolid in towers but solid outside them.

This does not change the behaviour of the "edge" spikes in towers.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -37,12 +37,6 @@ bool entityclass::checktowerspikes(int t, mapclass& map)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


TerryCavanagh:
Oh, hmm - I guess this change is related to push to support Tower maps in custom levels in CE? It looks like it shouldn't have any consequences for the original game, so happy to merge it. Thank you!

PR #153, opened 2020-02-08
Revert "Remove from M&P the room data of Final Level/Lab/Overworld/Tower"

leo60228:
Reverts TerryCavanagh/VVVVVV#124

This segfaults on many platforms, due to `fillcontents` dereferencing an uninitialized pointer. See https://github.com/NixOS/nixpkgs/pull/79068#issuecomment-583738049.

```
Starting program: /nix/store/24g187p6zjgawsgcss05rdq7nbirj8z6-vvvvvv-bin-unstable-2020-02-02/bin/VVVVVV -assets /nix/store/30jgxphl7vl4iwvsxcs5yv28z6868rwm-mapdata.zip
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/nix/store/6yaj6n8l925xxfbcd65gzqx3dz7idrnn-glibc-2.27/lib/libthread_db.so.1".
[New Thread 0x7ffff68fb700 (LWP 29967)]
[New Thread 0x7fffe6516700 (LWP 29968)]

Thread 1 "VVVVVV" received signal SIGSEGV, Segmentation fault.
std::__cxx11::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> >::basic_stringbuf (__mode=24, __str=<error reading variable: Cannot access memory at address 0x8>, this=0x7fffffff7098)
    at /nix/store/qksib0d7ygx1mmqk7172yjz02ya13lsh-gcc-9.2.0/include/c++/9.2.0/sstream:127
127             _M_string(__str.data(), __str.size(), __str.get_allocator())
(gdb) bt
#0  std::__cxx11::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> >::basic_stringbuf (__mode=24, __str=<error reading variable: Cannot access memory at address 0x8>, this=0x7fffffff7098)
    at /nix/store/qksib0d7ygx1mmqk7172yjz02ya13lsh-gcc-9.2.0/include/c++/9.2.0/sstream:127
#1  std::__cxx11::basic_stringstream<char, std::char_traits<char>, std::allocator<char> >::basic_stringstream (__m=24, __str=<error reading variable: Cannot access memory at address 0x8>, this=0x7fffffff7080, __in_chrg=<optimized out>,
    __vtt_parm=<optimized out>) at /nix/store/qksib0d7ygx1mmqk7172yjz02ya13lsh-gcc-9.2.0/include/c++/9.2.0/sstream:774
[...]

PR #155, opened 2020-02-08
Initialize level data with 0s in M&P

InfoTeddy:
## Changes:

* **Initialize level data with 0s in Make & Play edition**

  Following discussion on #153, I suggested that instead of reverting my M&P guards from #124 (which would only revert it for The Final Level, The Lab, Overworld, and The Tower, leaving Space Station 1 & 2 and The Warp Zone alone which could potentially cause the same problem that motivated #153), we should initialize the map data with 0s instead.

## Legal Stuff:


flibitijibibo:
@leo60228, can you verify this patch? If it works for you we’ll go with this as the fix.

leo60228:
i will never get to sleep tonight

leo60228:
It does fix the segfault, yes.

flibitijibibo:
Cool, thanks to both of you for checking this out.

PR #157, opened 2020-02-10
Add option to compile without the level editor

mwpenny:
## Changes:

These changes allow the game to be compiled without the level editor code if the `NO_EDITOR` CMake flag is specified. This is desirable for embedded platforms with no mouse, etc. If the flag is set, game menus are updated accordingly:

image

## Legal Stuff:


InfoTeddy:
Going off of the screenshot, this also removes the ability to even play custom levels in the first place, which I feel is removing way too much. Please make it so that the menu for custom levels isn't removed.

InfoTeddy:
I have re-read the diffs more closely, and you remove the entirety of `editor.cpp`. However, you shouldn't trust the filename - `editor.cpp` doesn't contain *just* the editor, it also contains the code to play custom levels, too!

(VVVVVV has a problem with misleading filenames, or at least filenames that omit information. So a file not only contains the thing its filename says, but *more* things which may or may not be related. For example, `Spacestation2.cpp` doesn't contain only Space Station 2, it also contains Space Station 1. `titlerender.cpp` doesn't contain the render functions for only the title screen, it contains every single render function. The hallway rooms in The Tower that are not a part of the main tower are stored in `Finalclass.cpp`, which is the file containing the level data for The Final Level.)

mwpenny:
I understand the convoluted nature of this codebase (but I'm glad it's out there nonetheless, thanks Terry!) and that this flag also removes the ability to play custom levels.

I am trying to remove the entire editor class in order to ease ports to limited systems. This is just a compile-time flag, so those who wish to play custom levels can leave the option disabled (which it is by default), no?

InfoTeddy:
I think it'd be useful to be able to play custom levels but not have a level editor. Having more fine-grained control like that would ease ports, too.

mwpenny:
If this is removing too much regardless, perhaps some additional granularity: I could add a second flag which disables the editor but not playing custom levels.

mwpenny:
Ha, beat me to it!

mwpenny:
The CMake flag is now a string, `CUSTOM_LEVEL_SUPPORT`, that can take one of three options:
1) `ENABLED` - Default behavior. No changes
2) `NO_EDITOR` - Ability to play custom levels, but not create them
3) `DISABLED` - Removes all custom level capability (along with the entire editor class)

I agree with you that this level of control is a lot better. I can imagine situations where you'd want user levels but an editor is infeasible.
image

InfoTeddy:
Thanks, that's a lot better!

flibitijibibo:
Second go is much better, thanks!

PR #158, opened 2020-02-10
Make "[Press ENTER to return to editor]" fade out after a bit

AllyTally:
## Changes:

This makes the `[Press ENTER to return to editor]` text in the editor fade out after a few frames, allowing screenshots of custom levels to be cleaner and to make sure nothing is obscured while the user is editing their level.
This commit also adds alpha support in BlitSurfaceColoured, where it takes into account the alpha of the pixel *and* the alpha of the color.
`graphics::getRGBA(r,g,b,a)` was added to help with this.

This is useful for future changes if anyone needs to use BlitSurfaceColoured with an alpha color, which could be something like making the translucent roomname bar use BlitSurfaceColoured instead of an entirely new surface.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -264,13 +264,18 @@ void Graphics::MakeSpriteArray()
 }
 
 
-void Graphics::Print( int _x, int _y, std::string _s, int r, int g, int b, bool cen /*= false*/ )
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Functionally looks good to me, designwise I’ll leave it up to @TerryCavanagh...

TerryCavanagh:
Looks good. Thanks @AllyTally!

PR #160, opened 2020-02-11
Use dynamic credits lists for main menu credits screens (and add GitHub contributors)

mwpenny:
## Changes:

In commit 2aa4aa5, @flibitijibibo added the names of GitHub contributors to the end-of-game credits and refactored the code to use 3 lists (`superpatrons`, `patrons`, and `githubfriends`) instead of hard-coding everything. However, the names in the credits accessible from the main menu were still hard-coded as well as lacking the names of GitHub contributors.

This change refactors the main menu accessible credits to use the 3 lists and adds the names of GitHub contributors.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -235,6 +235,7 @@ void Game::init(void)
     currentmenuoption = 0;
     menuselection = "null";
     currentmenuname = "null";
+    current_credits_list_index = 0;
     menuxoff = 0;
     menuyoff = 0;
     menucountdown = 0;
@@ -406,6 +407,7 @@ void Game::init(void)
     superpatrons.push_back("Daniel Benmergui");
     superpatrons.push_back("David Pittman");
     superpatrons.push_back("Ian Bogost");
+    superpatrons.push_back("Ian Poma");
     superpatrons.push_back("Jaz McDougall");
     superpatrons.push_back("John Faulkenbury");
     superpatrons.push_back("Jonathan Whiting");
@@ -7259,36 +7261,6 @@ void Game::createmenu( std::string t )
     else if (t == "credits6")
     {
         menuoptions[0] = "next page";
-        menuoptionsactive[0] = true;
-        menuoptions[1] = "return";
-        menuoptionsactive[1] = true;
-        nummenuoptions = 2;
-        menuxoff = 20;
-        menuyoff = 64;
-    }
-    else if (t == "credits7")
-    {
-        menuoptions[0] = "next page";
-        menuoptionsactive[0] = true;
-        menuoptions[1] = "return";
-        menuoptionsactive[1] = true;
-        nummenuoptions = 2;
-        menuxoff = 20;
-        menuyoff = 64;
-    }
-    else if (t == "credits8")
-    {
-        menuoptions[0] = "next page";
-        menuoptionsactive[0] = true;
-        menuoptions[1] = "return";
-        menuoptionsactive[1] = true;
-        nummenuoptions = 2;
-        menuxoff = 20;
-        menuyoff = 64;
-    }
-    else if (t == "credits9")
-    {
-        menuoptions[0] = "first page";
         menuoptionsactive[0] = true;
         menuoptions[1] = "return";
         menuoptionsactive[1] = true;
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -7260,7 +7260,7 @@ void Game::createmenu( std::string t )
     }
     else if (t == "credits6")
     {
-        menuoptions[0] = "next page";
+        menuoptions[0] = "first page";
         menuoptionsactive[0] = true;
         menuoptions[1] = "return";
         menuoptionsactive[1] = true;
--- a/desktop_version/src/titlerender.cpp
+++ b/desktop_version/src/titlerender.cpp
@@ -279,106 +279,64 @@ void titlerender(Graphics& dwgfx, mapclass& map, Game& game, entityclass& obj, U
         }
         else if (game.currentmenuname == "credits3")
         {
-            dwgfx.Print( -1, 20, "VVVVVV is supported by", tr, tg, tb, true);
-            dwgfx.Print( 40, 30, "the following patrons", tr, tg, tb, true);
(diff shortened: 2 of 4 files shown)


desktop_version/src/Maths.h
```diff
@@ -6,6 +6,7 @@
 
 //// This header holds Maths functions that emulate the functionality of flash's
 
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
```
InfoTeddy: The game already uses `std::min()` in various places.

InfoTeddy:
You should cast all the `.size()`s to an `int` whenever you use lesser- and greater- than signs, otherwise the compiler emits warnings about comparing ints with different signedness, which is very noisy.

desktop_version/src/Maths.h
mwpenny: Hm my build failed when I tried

InfoTeddy: It shouldn't fail, because otherwise the build would have already failed on these usages of `std::min()` and `std::max()`.
std::min() and std::max() in VVVVVV
I think your build failed because you failed to cast the `.size()`s to `int`s (see comment below) when plugging them in to `std::min()`, so the compiler complained.

mwpenny: Ah it's because `.size()`. Returns a `size_t` and the first argument is an `int`. It's midnight here :p

mwpenny:
Good point about the signed/unsigned comparison warnings

desktop_version/src/Maths.h
mwpenny: Ha that's day 2 of you typing something at the same time as me

mwpenny:
I removed my `MIN` macro and used `std::min` instead, and casted the `size()` values to `int`.

desktop_version/src/Game.cpp
```diff
@@ -406,6 +407,7 @@ void Game::init(void)
     superpatrons.push_back("Daniel Benmergui");
     superpatrons.push_back("David Pittman");
     superpatrons.push_back("Ian Bogost");
+    superpatrons.push_back("Ian Poma");
```
mwpenny: This name was present in the old main menu credits but missing here.

desktop_version/src/titlerender.cpp
```diff
@@ -279,106 +279,64 @@ void titlerender(Graphics& dwgfx, mapclass& map, Game& game, entityclass& obj, U
...
-            dwgfx.Print( 80, 100,"Simon Schmid", tr, tg, tb);
-            dwgfx.Print( 80, 110,"Stephen Maxwell", tr, tg, tb);
-            dwgfx.Print( 80, 120,"Swing Swing Submarine", tr, tg, tb);
-            dwgfx.Print( 80, 130,"Tam Toucan", tr, tg, tb);
-            dwgfx.Print( 80, 140,"Terry Dooher", tr, tg, tb);
-            dwgfx.Print( 80, 150,"Tim W.", tr, tg, tb);
-            dwgfx.Print( 80, 160,"Timothy Bragan", tr, tg, tb);
-        }
-        else if (game.currentmenuname == "credits9")
+            for (int i = startidx; i < startidx + 14 && i < game.patrons.size(); ++i)
```
InfoTeddy: Forgot an `(int)` here, compiler spits out a warning on this line.

mwpenny: The whole loop condition can actually be changed to `i < endidx`

mwpenny: Done

desktop_version/src/editor.cpp
```diff
@@ -355,6 +355,8 @@ void editorclass::reset()
     hookmenu=0;
     numhooks=0;
     script.customscript.clear();
+
+    returneditoralpha = 0;
```
InfoTeddy: I get that you're trying to make this initialization consistent with the rest of the initializations, and want to reset it properly, too, but why touch this in this PR? (Also touched in `editor.h`.)

mwpenny: I am using an old compiler and could not build with it initialized in the header.

desktop_version/src/titlerender.cpp
InfoTeddy: Looks good, thanks!

desktop_version/src/editor.cpp
mwpenny: As @flibitijibibo said on #47:

> Try to aim for the absolute minimum C++ standard possible (if you can get away with it, write it as basically C). We have to support some very old compilers, so there’s no room to be overly fancy. [link](https://github.com/TerryCavanagh/VVVVVV/pull/47#issuecomment-573376232)

InfoTeddy: Alright, well, put it in another PR, then.

mwpenny: I can do it as a separate PR if you want. Another problem: #47 introduced the use of `char32_t` which only works on compilers that support C++11. I could roll fixes for both into one "compatibility" PR.

mwpenny: Done. Undid that change

InfoTeddy: Yeah, fix the `char32_t` stuff and make it a general "fix compatibility" PR.

mwpenny: Sounds good

flibitijibibo:
I wasn’t sure about this menu since the list could potentially get infinitely long... but I suppose we could get creative with this one dynamic menu if we needed to.

Only thing I would change is the order of the menus, we should keep them in the same order as the final credit scroll.

mwpenny:
True. For now it should just show as many pages as needed, with a handful of names per page. 

As for the ordering, no problem. I wasn't sure which way to go because of the wording. The list of people who "support VVVVVV" ends with "and you!", with a thank you note to the player. It made it seem like that should be the last page. If that doesn't matter, I can put GitHub contributors after patrons.

flibitijibibo:
For now let’s keep the order the same across both - if we end up changing the order we can do both at the same time.

mwpenny:
Sounds good, will do

mwpenny:
Alright I moved GitHub contributors to appear after patrons

PR #163, opened 2020-02-12
Implement command-line playtesting

leo60228:
## Changes:

A long-requested feature of third-party VVVVVV level editors like Ved
has been to support playtesting. However, that hasn't been possible. For
April Fools Ved included Vis, a clone of VVVVVV with jumping advertising
itself as just having physics glitches, but until recently there hasn't
been any way to *actually* playtest a level. In November, I created
libvloader, which only on Linuxenabled this functionality. It was
quickly integrated into Ved in a PR, but the difficulties of using
LD_PRELOAD from Lua, as well as only supporting Linux, this stalled.
VVVVVV-CE has replicated this functionality, but due to the lack of a
stable release as well as minor incompatibilities, Ved has been
reluctant to add playtesting support that only supports VVVVVV-CE, but
this PR ports the playtesting support back to upstream. If there's
anything VVVVVV-CE specific I accidentally included in this PR, please
tell me!

## Legal Stuff:

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -148,27 +148,27 @@ char *FILESYSTEM_getUserLevelDirectory()
 void FILESYSTEM_loadFileToMemory(const char *name, unsigned char **mem,
                                  size_t *len, bool addnull)
 {
-        if (strcmp(name, "levels/special/stdin.vvvvvv") == 0) {
-            // this isn't *technically* necessary when piping directly from a file, but checking for that is annoying
-            static std::vector<char> STDIN_BUFFER;
-            static bool STDIN_LOADED = false;
-            if (!STDIN_LOADED) {
-                std::istreambuf_iterator<char> begin(std::cin), end;
-                STDIN_BUFFER.assign(begin, end);
-                STDIN_BUFFER.push_back(0); // there's no observable change in behavior if addnull is always true, but not vice versa
-                STDIN_LOADED = true;
-            }
+	if (strcmp(name, "levels/special/stdin.vvvvvv") == 0) {
+		// this isn't *technically* necessary when piping directly from a file, but checking for that is annoying
+		static std::vector<char> STDIN_BUFFER;
+		static bool STDIN_LOADED = false;
+		if (!STDIN_LOADED) {
+			std::istreambuf_iterator<char> begin(std::cin), end;
+			STDIN_BUFFER.assign(begin, end);
+			STDIN_BUFFER.push_back(0); // there's no observable change in behavior if addnull is always true, but not vice versa
+			STDIN_LOADED = true;
+		}
 
-            auto length = STDIN_BUFFER.size() - 1;
-            if (len != NULL) {
-                *len = length;
-            }
+		size_t length = STDIN_BUFFER.size() - 1;
+		if (len != NULL) {
+			*len = length;
+		}
 
-            ++length;
-            *mem = static_cast<unsigned char*>(malloc(length)); // STDIN_BUFFER.data() causes double-free
-            std::copy(STDIN_BUFFER.begin(), STDIN_BUFFER.end(), reinterpret_cast<char*>(*mem));
-            return;
-        }
+		++length;
+		*mem = static_cast<unsigned char*>(malloc(length)); // STDIN_BUFFER.data() causes double-free
+		std::copy(STDIN_BUFFER.begin(), STDIN_BUFFER.end(), reinterpret_cast<char*>(*mem));
+		return;
+	}
 
 	PHYSFS_File *handle = PHYSFS_openRead(name);
 	if (handle == NULL)
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -164,7 +164,7 @@ void FILESYSTEM_loadFileToMemory(const char *name, unsigned char **mem,
                 *len = length;
             }
 
-            if (addnull) ++length;
+            ++length;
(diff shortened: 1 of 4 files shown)


flibitijibibo:
The command line portion seems okay, it's actually the filesystem stuff that really throws me off - is this not compatible with PhysicsFS at all? The only reason we have platform paths in there at all is because of the migration logic, which is probably going away sooner than later... I would like to not have any platform paths whatsoever after that (minus the Win32 save path, barf), so if we're able to use physfs to load levels from the command line I would definitely prefer that.

InfoTeddy:
Is PhysFS able to load levels outside the game's given directory? If so, maybe we can use PhysFS instead of having platform paths.

If not, and we *can't* not use PhysFS, then it makes using these playtesting args very inconvenient on Ved's end. The previous solution with libvloader required making our own VVVVVV user directory, setting `XDG_DATA_HOME` (which doesn't have a nice equivalent on Windows) to that, and then having to symlink the user's custom assets so we didn't have to copy them. The problem with the symlinks is that a broken symlink to `vvvvvvmusic.vvv` or `mmmmmm.vvv` would be recognized by the game, and the game would not default to the default music files and would simply not play any music. Also, Ved's filesystem functions don't have support symlinks in the first place, making this much more of a headache when I tried to add playtesting with libvloader.

leo60228:
@flibitijibibo I'm somewhat confused about the issue. The new FileSystemUtils functions are just cross-platform versions of POSIX `realpath`, `dirname`, and `basename`.

Dav999-v:
If you use Windows API functions, make sure not to break compatibility with unicode paths (like international usernames); functions like `GetCurrentDirectory`, `PathRemoveFileSpecA` and `PathFindFileNameA` unfortunately work with single-byte charsets and will use question marks for unsupported characters.

leo60228:
@Dav999-v Windows 10 1903 adds a manifest property to have -A functions work on UTF-8. If that's not sufficient, I can try to use the PhysFS conversion functions.

Dav999-v:
@leo60228 I'd think it's better to not rely on the manifest property, people are definitely still going to be using Windows 7 (or 8.1) or something in a few years and wonder why this isn't working for them.

leo60228:
To be fair, Windows 7 is already EOL, and Windows 8.1 will be in less than 3 years.k

InfoTeddy:
Any update on this PR? It's been a while.

flibitijibibo:
The filesystem code is enough for me to reject this - if it can’t be fixed to be written portably it’ll have to be a CE feature only.

leo60228:
The diff was completely ruined by my fake "merge," let me fix the history...

leo60228:
This is used from a shell like:
```
ls ~/.local/share/VVVVVV/levels/level.vvvvvv # level.vvvvvv
ls external_level.vvvvvv # external_level.vvvvvv
VVVVVV -p level # starts level.vvvvvv
VVVVVV -p special/stdin < external_level.vvvvvv # starts external-level.vvvvvv
```
This completely sidesteps the Unicode and PhysFS problems by bypassing PhysFS and not requiring Unicode to be handled in VVVVVV itself.

leo60228:
`special/stdin` was chosen to prevent conflicts with "real" levels, as `/` is the only character allowed in arguments but not in paths on either Linux or Windows.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -141,9 +145,37 @@ char *FILESYSTEM_getUserLevelDirectory()
...
+    return static_cast<unsigned char>(orig);
+}
+
 void FILESYSTEM_loadFileToMemory(const char *name, unsigned char **mem,
                                  size_t *len, bool addnull)
 {
+        if (strcmp(name, "levels/special/stdin.vvvvvv") == 0) {
+            // this isn't *technically* necessary when piping directly from a file, but checking for that is annoying
+            static bool STDIN_LOADED = false;
+            static std::vector<char> STDIN_BUFFER;
```
flibitijibibo: My C++ is a little rough, but would using `<unsigned char>` here make it so we don't need the `std::transform` at the bottom? This seems like a really unnecessary memory duplication because of C++-isms; frankly I'd prefer the longer C version if the tradeoff is that the actual code in execution is simpler. (I suppose you could also use a local `char* dst` and then do `*mem = (unsigned char*) dst` at the end, the compiler shouldn't complain there...)

Either way, this is _way_ better, it's a really neat idea!

leo60228: That second idea is smart, I'll try that.

InfoTeddy:
Just rebased onto master, is there anything else we need to do before you'll merge this in?

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -144,6 +148,28 @@ char *FILESYSTEM_getUserLevelDirectory()
...
+            static bool STDIN_LOADED = false;
+            if (!STDIN_LOADED) {
+                std::istreambuf_iterator<char> begin(std::cin), end;
+                STDIN_BUFFER.assign(begin, end);
+                STDIN_BUFFER.push_back(0); // there's no observable change in behavior if addnull is always true, but not vice versa
+                STDIN_LOADED = true;
+            }
+
+            auto length = STDIN_BUFFER.size() - 1;
```
flibitijibibo, line 162: This can be `size_t`, other than that everything looks good now! I'd still be interested in finding a way to write directly to *mem without the stdin buffer, but since it's static memory I'm not too heartbroken about it.

InfoTeddy: Alright, fixed. I also fixed the indentation here as well.

PR #164, opened 2020-02-12
Fixed some music bugs

Stelpjo:
## Legal Stuff:

## Changes:

Previously, when MMMMMM is installed but the user is using PPPPPP, niceplay would still restart the song even if it's the same. That has been fixed. In addition, Plenary and Path Complete no longer loop when MMMMMM is installed but PPPPPP is in use.

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -246,7 +246,7 @@ void musicclass::play(int t)
 		{
 			// musicfade = 0;
 			currentsong = t;
-			if (currentsong == 0 || currentsong == 7)
+			if (currentsong == 0 || currentsong == 7 || currentsong == 16 || currentsong == 23)
 			{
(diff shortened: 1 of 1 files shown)


parkerlreed:
Just wanted to say thank you. I didn't even know the alternate soundtrack existed.

AllyTally:
There's an issue with this PR--custom levels used the Plenary/Path Complete looping bug. The first one that comes to mind is [Unshackled](https://distractionware.com/forum/index.php?topic=3388), using the bug to have two extra looping music slots that would otherwise be impossible to have. A solution for this would be fixing the bug only for the main game and not for custom levels.

flibitijibibo:
@AllyTally Makes sense to me, write up the patch and I'll merge it in right away.

InfoTeddy:
#190 implements what Ally suggested.

PR #169, opened 2020-02-20
Make 'script.commands', 'ed.sb', and 'ed.hooklist' no longer use separate length-trackers

InfoTeddy:
## Changes:

* **Make `script.commands`, `ed.sb`, and `ed.hooklist` not use separate length-trackers**

  This is a refactor that turns the script-related arrays `ed.sb`, and `ed.hooklist` into C++ vectors (`script.commands` was already a vector, it was just misused). The code handling these vectors now looks more like idiomatic C++ than sloppily-pasted pseudo-ActionScript. This removes the variables `script.scriptlength`, `ed.sblength`, and `ed.numhooks`, too.

  This reduces the amount of code needed to e.g. simply remove something from any of these vectors. Previously the code had to manually shift the rest of the elements down one-by-one, and doing it manually is definitely error-prone and tedious.

  But now we can just use fancy functions like `std::vector::erase()` and `std::remove()` to do it all in one line!

  Don't worry, I checked and `std::remove()` is in the C++ standard since at least 1998.

  This patch makes it so the `commands` vector gets cleared when `scriptclass::load()` is ran. Previously, the `commands` vector never actually properly got cleared, so there could potentially be glitches that rely on the game indexing past the bounds set by `scriptlength` but still in-bounds in the eyes of C++, and people could potentially rely on such an exploit...

  However, I checked, and I'm pretty sure that no such glitch previously existed at all, because the only times the vector gets indexed are when `scriptlength` is either being incremented after starting from 0 (`add()`) or when it's underneath a `position < scriptlength` conditional.

  Furthermore, I'm unaware of anyone who has actually found or used such an exploit, and I've been in the custom level community for 6 years.

  So I think it's fine.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -11,12 +11,10 @@ scriptclass::scriptclass()
     	//Start SDL
 
 	//Init
-	commands.resize(500);
 	words.resize(40);
 	txt.resize(40);
 
 	position = 0;
-	scriptlength = 0;
 	scriptdelay = 0;
 	running = false;
 
@@ -72,7 +70,7 @@ void scriptclass::run( KeyPoll& key, Graphics& dwgfx, Game& game, mapclass& map,
 {
 	while(running && scriptdelay<=0 && !game.pausescript)
 	{
-		if (position < scriptlength)
+		if (position < (int) commands.size())
 		{
 			//Let's split or command in an array of words
 			tokenize(commands[position]);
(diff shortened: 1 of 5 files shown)


flibitijibibo:
lgtm, I don't know of any compatibility issues but I don't know much about what's out there.

InfoTeddy:
I accidentally deleted the branch then restored it, whoops

InfoTeddy:
Is there any update on this PR? It looks like Terry hasn't responded to it yet.

flibitijibibo:
Might be a while since Dicey Dungeons Switch is going on... this looks fine to me though, if it breaks compat we’ll revert it later.

PR #172, opened 2020-03-01
Refactor edentities to not use separate length-trackers

InfoTeddy:
## Changes:

* **Refactor edentities to not use separate length-trackers**

  This turns the array `edentity` into a proper vector, and removes the need to use a separate length-tracking variable and manually keep track of the actual amount of edentities in the level by using the long-winded `EditorData::GetInstance().numedentities`. This manual tracking was more error-prone and much less maintainable.

  `editorclass::naddedentity()` has been removed due to now functionally being the same as `editorclass::addedentity()` (there's no more `EditorData::GetInstance().numedentities` to not increment) and for also being unused in the first place.

  `editorclass::copyedentity()` has been removed because it was only used to shift the rest of the edentities up manually, but now that we let C++ do all the hard work it's no longer necessary.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -1646,7 +1646,7 @@ void mapclass::loadlevel(int rx, int ry, Graphics& dwgfx, Game& game, entityclas
 		//Entities have to be created HERE, akwardly
 		int tempcheckpoints=0;
 		int tempscriptbox=0;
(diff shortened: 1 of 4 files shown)


flibitijibibo:
I bet there are a lot of these... probably a consequence of the Flash roots? I liked fixed-size arrays but yeah, this was really tough to track.

PR #174, opened 2020-03-03
Refactor entities, blocks, linecross, and textboxes to not use the 'active' attribute system and to not have separate length-trackers

InfoTeddy:
### Summary

This patch removes the variables `nentity`, `nblocks`, `nlinecrosskludge`, and `ntextbox` from `entityclass` (and `Graphics`) in favor of using `entities.size()`, `blocks.size()`, `linecrosskludge.size()`, and `textbox.size()`, which are amount-tracking methods handled automatically by C++. Furthermore, entities, blocks, and textboxes no longer have an `active` attribute, and if an entity, block, or textbox exists, it is always guaranteed to be real and not fake, unlike the previous system using the `active` attribute.

As a bonus, `numlines` is removed from `textboxclass` and `txtnumlines` is removed from `scriptclass` in favor of using `line.size()` and `txt.size()` respectively.

Doing these things not only makes code maintenance easier, but it also makes debugging easier, even when it's not explicitly related to working on the codebase. I'm a fan of attaching GDB to VVVVVV and printing out various things to debug issues in my custom levels, ever since I discovered that the Linux and macOS versions have debug symbols (thanks @flibitijibibo!), but unfortunately whenever I do something like `print obj.blocks`, that means I get 500 slots, most of which are not even real, because from the point of view of C++ there are 500 slots in obj.blocks. Then whenever I'm using my brain to analyze a given slot, I always have to make sure that its `active` is set to true before I do any thinking, else it's not real and I'm just wasting time, which is always an annoyance.

Furthermore, this patch also fixes segfaults resulting from Undefined Behavior in a few custom levels, most notably Unshackled v1.2.

&#x200b;

The function `entityclass::setblockcolour()` has been moved to now be on the `blockclass` itself, so it is now `blockclass::setblockcolour()`. So instead of `entityclass::setblockcolour()` taking in the index of the block whose color you want to set, you simply take the block and call its `setblockcolour()` method. Same thing for `entityclass::setenemyroom()`, `entityclass::setenemy()`, and `entityclass::settreadmillcolour()` - they are now on the entclass as `entclass::setenemyroom()`, `entclass::setenemy()`, and `entclass::settreadmillcolour()`, respectively.

For the `linecrosskludge` handling, instead of copying only specific attributes, I've decided to just copy the full attributes of each object instead. I'm pretty sure it doesn't affect anything, and I've checked.

I also took the opportunity to clean up some formatting and code style, such as using a case-switch for checking `para` in `entclass::setenemy()` instead of an `if`-`else if` chain, cleaning up some mixed indentation, removing commented-out code from earlier development of this game, removing unnecessary int casts in assignments (they definitely should be kept in comparisons, though, and any assignment expression more complicated than `a = b`), collapsing double-nested if-statements by putting them together into one with an "and" (the biggest offenders of this were `entityclass::hplatformat()` and `entityclass::customwarplinecheck()`), and fixing spacing.

The biggest code cleanup (besides the removal of the `active` system) is that I've added a formal function to remove an entity, `entityclass::removeentity()`. I found it weird that there's an `entityclass::removeblock()` but no `entityclass::removeentity()`. And in this patch I use `entityclass::removeentity()` in place of setting the `active` attribute of an entity to false.

Furthermore, this makes functions like `entityclass::cleanup()` unnecessary, nor the equivalent for blocks that never actually got formalized into a function and just got copy-pasted everywhere, which is typical of this game. Thus `entityclass::cleanup()` has been removed due to it no longer being necessary, since we no longer manually keep track of the actual amount of active entities separately, so there's no way that the amount of entities will desync with the actual amount of entities. The same thing has been done for blocks, removing that copy-pasted equivalent for blocks from everywhere it was copy-pasted.

&#x200b;

There are now macros `removeentity_iter()` and `removeblock_iter()` to remove an entity or a block while iterating through `obj.entities` or `obj.blocks`. It assumes that `obj.entities.size()` or `obj.blocks.size()` will always decrease by one, and that you're not iterating through these vectors backwards. I think most of the time those assumptions are sound.

Although, iterating backwards avoids the above problem entirely. Luckily, the game already does iterate backwards in the cases where I can't simply decrement the index when removing an entity, which is in `entityclass::updateentities()`. I was a bit spooked, however, when `entityclass::updateentities()` called itself. Apparently it's just a way to re-update an entity, because each entity doesn't actually have its own update method, and instead it's just all conflated into one giant function (not unlike the rest of the game's code). I had to make sure that each entity type's update code that called `entityclass::updateentities()` didn't also call `removeentity()`, or else it could result in removing the wrong entity. Fortunately, this was the case, and I could rest easy once again.

&#x200b;

One of the concerns I had while writing this patch was that there would be some behavior dependent on the game forgetting to check the `active` attribute of an object before doing something. As far as I can tell, there are no such exploits except in very contrived almost-fictional scenarios. For example, `getcrewman()` didn't check the `active` attribute, which could theoretically lead to the game targeting a deactivated crewmate instead of a real one, but there's no way to deactivate crewmates without `gotoroom`, which deactivates all entities (and runs `entityclass::cleanup()`, fixing said theoretical issue) anyway.

However, see a few paragraphs below for a real (but trivial) case where the game forgot to check the `active` attribute of the current active activity zone, that I fixed.

In VVVVVV: Community Edition, there's a `destroy(crewmates)` command, which can be used to deactivate a crewmate and thus *could* create the above issue, but I want such issues to be fixed anyway. Hence, this patch removing the need to check the `active` attribute, because now if you have an entity object, it's guaranteed to be actually real and not fake.

&#x200b;

There's a curious oddity in `WarpClass.cpp` for the room Murdering Twinmaker. Apparently, the game adds one deactivated entity before adding the teleporter, so the teleporter is separated from the rest of the entities by one entity. I'm not sure for what purpose this was done - if the teleporter here was being referenced by a hardcoded index somewhere else in the code, it would break if you came back here after rescuing Verdigris.

Given that my new system means you simply cannot have deactivated entities in between other active entities, and that I have no idea what this oddity means, I've decided to not keep this oddity.

&#x200b;

While working on this patch, I found that the game was not setting `game.activeactivity` to -1 after pressing Enter on a given activity zone and removing it. This meant that the game was still referencing the prompt and color of a deactivated activity zone, which isn't *technically* undefined behavior but is pretty much in the same vein as it, and got turned into *actual* undefined behavior when I was working on this patch, causing segfaults.

The previous behavior was that when you pressed Enter on an activity zone, the rendering of the prompt afterwards was simply solid black, but this behavior was actually a bug. It was referencing a deactivated block that had its `clear()` method called on it, which is why it was solid black! So now when you press Enter on an activity zone, the prompt will nicely fade out and won't suddenly turn to black and then after a few frames disappear just as suddenly.

As you can see, removing this whole `active` system not only means better-working code in general, but also exposing (and fixing) certain lurking bugs in the game that wouldn't have otherwise been brought to light!

## Legal Stuff:

### DIFF
--- a/desktop_version/src/BlockV.cpp
+++ b/desktop_version/src/BlockV.cpp
@@ -7,7 +7,6 @@ blockclass::blockclass()
 
 void blockclass::clear()
 {
-	active = false;
 	type = 0;
 	trigger = 0;
 
@@ -33,4 +32,69 @@ void blockclass::rectset(const int xi, const int yi, const int wi, const int hi)
 	rect.y = yi;
 	rect.w = wi;
 	rect.h = hi;
+}
+
+void blockclass::setblockcolour(std::string col)
+{
+	if (col == "cyan")
+	{
+		r = 164;
+		g = 164;
+		b = 255;
+	}
+	else if (col == "red")
+	{
+		r = 255;
+		g = 60;
+		b = 60;
+	}
+	else if (col == "green")
+	{
+		r = 144;
+		g = 255;
+		b = 144;
+	}
+	else if (col == "yellow")
+	{
+		r = 255;
+		g = 255;
+		b = 134;
+	}
+	else if (col == "blue")
+	{
+		r = 95;
+		g = 95;
+		b = 255;
+	}
+	else if (col == "purple")
+	{
+		r = 255;
+		g = 134;
+		b = 255;
+	}
+	else if (col == "white")
+	{
+		r = 244;
+		g = 244;
+		b = 244;
+	}
+	else if (col == "gray")
+	{
+		r = 174;
+		g = 174;
+		b = 174;
+	}
+	else if (col == "orange")
+	{
+		r = 255;
+		g = 130;
+		b = 20;
+	}
+	else
+	{
+		//use a gray
+		r = 174;
+		g = 174;
+		b = 174;
+	}
 }
--- a/desktop_version/src/BlockV.h
+++ b/desktop_version/src/BlockV.h
@@ -12,9 +12,10 @@ public:
     void clear();
 
     void rectset(const int xi, const int yi, const int wi, const int hi);
+
+    void setblockcolour(std::string col);
 public:
     //Fundamentals
-    bool active;
     SDL_Rect rect;
     int type;
     int trigger;
--- a/desktop_version/src/Ent.cpp
+++ b/desktop_version/src/Ent.cpp
@@ -8,7 +8,6 @@ entclass::entclass()
 void entclass::clear()
 {
 	// Set all values to a default, required for creating a new entity
-	active = false;
 	invis = false;
 	type = 0;
 	size = 0;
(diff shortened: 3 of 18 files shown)


InfoTeddy:
Any update on this PR? It's been a few weeks.

flibitijibibo:
6,000 lines in a single commit is a lot - is it possible to split this up?

InfoTeddy:
Alright. Yeah, I probably shouldn't have thrown in all the other vectors in to the patch as well, in hindsight.

Should I just break it up by each individual vector? If so, I think the biggest patch will probably be `entities`, followed by `blocks`, which I think is fine also.

flibitijibibo:
That sounds good - if they still break a few thousand lines you may have to trim it down further with non-essential changes (like the other patch).

InfoTeddy:
Ok, makes sense to me as well.

PR #177, opened 2020-03-04
Fix unfocusing the game while in fullscreen mode

InfoTeddy:
## Changes:

* **Fix unfocusing the game while in fullscreen mode**

  If you Alt+Tabbed while in fullscreen mode, the game would stay in fullscreen instead of switching to windowed, but there was a chance it would EITHER use the same internal resolution which would mismatch the window resolution (don't know when exactly this happens, but still) and stay being in an actual windowed mode, OR switch between fullscreen/windowed every other time you re-focused the window, which is annoying.

  Now, whenever you Alt+Tab in fullscreen, the game will be in windowed mode, and then when you re-focus it will go back to fullscreen. Consistently.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/KeyPoll.cpp
+++ b/desktop_version/src/KeyPoll.cpp
@@ -248,6 +248,7 @@ void KeyPoll::Poll()
 					SDL_SetWindowFullscreen(window, 0);
 				}
 				SDL_DisableScreenSaver();
+				resetWindow = true;
 			}
 			else if (evt.window.event == SDL_WINDOWEVENT_FOCUS_LOST)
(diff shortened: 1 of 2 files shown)


flibitijibibo:
This ends up calling SDL_SetWindowFullscreen every frame, which definitely seems wrong - we should probably fix the reset bool instead (maybe trigger of focus change?).

InfoTeddy:
Alright, I've done that instead. Is it better now?

flibitijibibo:
Yes, definitely! Thanks for the fix.

PR #178, opened 2020-03-05
Stop passing 'graphics', 'map', 'obj', 'music', 'key', and 'help' everywhere

InfoTeddy:
### Summary

This patch removes `graphics`, `map`, `obj`, `music`, `key`, and `help` from every single function signature. In other words, it strips out the argument passing stuff that is prevalent in this game.

As a bonus, I took the opportunity to do some cleanup of function arguments in general. I removed unused parameters from functions, and I removed some unused functions or variables, as well as ones that were essentially useless and did nothing (see below).

Furthermore, all expressions of the form `UtilityClass::`*`something`* have been
changed to `help.`*`something`*. It's shorter this way.

The following functions had unused parameters:
 - `entityclass::checkdirectional()`
 - `entityclass::getcompanion()`
 - `Graphics::drawtile()`
 - `Graphics::drawtile2()`
 - `Graphics::Hitest()`

The following unused or useless functions and variables have been removed:
 - `Game::telegotoship()`
 - `Game::sfpsmode`
 - `Game::telecookieexists`
 - `Game::quickcookieexists`
 - `updategraphicsmode()`
 - `editorclass::weirdloadthing()` (instead, make a copy of the string and use
   `editorclass::load()`)
 - `Screen::ClearScreen()`

I also cleaned up some formatting, such as removing mixed indentation, fixing spacing, and removing some commented-out code from earlier versions of the game.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -3,7 +3,7 @@
 #include "Map.h"
 #include "UtilityClass.h"
 
-bool entityclass::checktowerspikes(int t, mapclass& map)
+bool entityclass::checktowerspikes(int t)
 {
     tempx = entities[t].xp + entities[t].cx;
     tempy = entities[t].yp + entities[t].cy;
@@ -210,19 +210,19 @@ void entityclass::swnenemiescol( int t )
     }
 }
 
-void entityclass::gravcreate( Game& game, int ypos, int dir, int xoff /*= 0*/, int yoff /*= 0*/ )
+void entityclass::gravcreate(int ypos, int dir, int xoff /*= 0*/, int yoff /*= 0*/)
 {
     if (dir == 0)
     {
-        createentity(game, -150 - xoff, 58 + (ypos * 20)+yoff, 23, 0, 0);
+        createentity(-150 - xoff, 58 + (ypos * 20)+yoff, 23, 0, 0);
     }
     else
     {
-        createentity(game, 320+150 + xoff, 58 + (ypos * 20)+yoff, 23, 1, 0);
+        createentity(320+150 + xoff, 58 + (ypos * 20)+yoff, 23, 1, 0);
     }
 }
 
-void entityclass::generateswnwave( Game& game, UtilityClass& help, int t )
+void entityclass::generateswnwave(int t)
 {
     //generate a wave for the SWN game
     if(game.swndelay<=0)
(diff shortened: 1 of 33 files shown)


desktop_version/src/Entity.cpp
```diff
@@ -210,19 +210,19 @@ void entityclass::swnenemiescol( int t )
     }
 }
 
-void entityclass::gravcreate( Game& game, int ypos, int dir, int xoff /*= 0*/, int yoff /*= 0*/ )
+void entityclass::gravcreate(int ypos, int dir, int xoff /*= 0*/, int yoff /*= 0*/)
 {
     if (dir == 0)
     {
-        createentity(game, -150 - xoff, 58 + (ypos * 20)+yoff, 23, 0, 0);
+        createentity(-150 - xoff, 58 + ypos*20 + yoff, 23, 0, 0);
```
flibitijibibo: Keep parens here to reduce size of the diff

InfoTeddy:
Just rebased onto master, along with removing the useless function `Screen::ClearScreen()`.

InfoTeddy:
Just rebased onto master again. CI on macOS seems to still be failing.

Is there any update on this PR? It's been a few weeks now.

flibitijibibo:
I’m guessing the repo that the Mac build uses is busted - if it doesn’t fix itself at some point I’ll just remove it.

As for the patch, 10,000 lines of changes in a single commit is just too much for me to review properly - this needs to be broken up into incremental changes, and the style changes may have to come separately (including the comment deletions).

InfoTeddy:
> 10,000 lines of changes in a single commit is just too much for me to review properly

Alright, fair enough.

How should I break it up? Should I just do one file at a time?

Also I should probably ask, do you care if I change all the `dwgfx`es into `graphics` like I did here? I was under the impression that `dwgfx` was just a remnant from the Flash version, and `graphics` was the new name, but actually `graphics` is only used in `main.cpp`, so I was thinking that if anything, the `graphics`es in `main.cpp` should be renamed to `dwgfx` instead of all `dwgfx`es being renamed to `graphics`.

flibitijibibo:
The design changes are okay as far as I can tell - just make each API change one at a time, and make each change one commit (for example, maybe do style fixes for each file as individual commits first, then delete the comments in each file next, then each API change can come at the end in separate commits). There are probably lots of ways to approach splitting these up, as long as they’re not thousands of lines each.

InfoTeddy:
Alright, makes sense to me.

PR #180, opened 2020-03-12
Allow using F11 and (on macOS) RCommand+Enter/F to toggle fullscreen

InfoTeddy:
## Changes:

* **Allow using F11 and, on macOS, RCommand+Enter/F to toggle fullscreen**

  This patch allows the use of pressing F11 to toggle fullscreen, as well as allowing the use of Right Command when using the Command+Enter/F shortcut on macOS. Apparently Alt+Enter isn't the only shortcut to toggle fullscreen, Alt+F also works, which I didn't know before.

  I'm adding F11 as a shortcut because it's a far more natural shortcut to toggle fullscreen than this Alt+Enter or Alt+F business, which seems to be a relic mimicking some other games and some Microsoft stuff?

  I'm also adding RCommand+Enter/F because I see no reason not to. If you can use RAlt on non-macOS, why can't you use RCommand on macOS, too?

  I also cleaned up the formatting relating to the shortcut code, and took the ifdef for macOS outside of the if-statement to make it more readable.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/KeyPoll.cpp
+++ b/desktop_version/src/KeyPoll.cpp
@@ -81,14 +81,13 @@ void KeyPoll::Poll()
 			{
 				pressedbackspace = true;
 			}
-			else if (	(	evt.key.keysym.sym == SDLK_RETURN ||
-						evt.key.keysym.sym == SDLK_f	) &&
-#ifdef __APPLE__ /* OSX prefers the command key over the alt keys. -flibit */
-					keymap[SDLK_LGUI]	)
+			else if (((evt.key.keysym.sym == SDLK_RETURN || evt.key.keysym.sym == SDLK_f) &&
+#ifdef __APPLE__ /* OSX prefers the command keys over the alt keys. -flibit */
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Functionally it’s fine, but I wonder if we should move all these checks into individually named variables, then make the if statement readable with those names. As of now... I don’t even know if my own style would be tolerable to read :(

InfoTeddy:
Yeah, the ifdef being in the middle of the if-statement is probably not a good thing.

Also, I just noticed another bug - pressing Alt+F in the script editor will still type F. We should probably attach the `if (textentrymode)` conditional to the if-tree with "if Alt+Enter/F or F11".

InfoTeddy:
Done in 0ac67199064e0413ad6be9af5933e9ba503c3bbf. Looks like that "typing F" bug can't be fixed by simply changing the control flow like I suggested, and I have no idea why. I'll leave it until I figure out how to properly fix it later.

flibitijibibo:
Perfect, thanks!

PR #181, opened 2020-03-12
Update lodepng and utfcpp

AlexApps99:
## Changes:

Updated `lodepng` and `utfcpp`

## Legal Stuff:

### DIFF
--- a/third_party/lodepng/LICENSE.txt
+++ b/third_party/lodepng/LICENSE.txt
@@ -1,0 +1,21 @@
+Copyright (c) 2005-2018 Lode Vandevenne
+
+This software is provided 'as-is', without any express or implied
+warranty. In no event will the authors be held liable for any damages
+arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose,
+including commercial applications, and to alter it and redistribute it
+freely, subject to the following restrictions:
+
+    1. The origin of this software must not be misrepresented; you must not
+    claim that you wrote the original software. If you use this software
+    in a product, an acknowledgment in the product documentation would be
+    appreciated but is not required.
+
+    2. Altered source versions must be plainly marked as such, and must not be
+    misrepresented as being the original software.
+
+    3. This notice may not be removed or altered from any source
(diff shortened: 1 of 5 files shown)


InfoTeddy:
Hm, why doesn't the game use libpng, or better yet SDL_image, instead of lodepng?

InfoTeddy:
Also, no TinyXML update? Or is there none?

AlexApps99:
That is a good point you raise, should I create a PR to use `SDL_image` instead?

AlexApps99:
The other dependencies were up-to-date

InfoTeddy:
> That is a good point you raise, should I create a PR to use `SDL_image` instead?

If you're up for it, yes.

InfoTeddy:
> The other dependencies were up-to-date

That's what I figured.

flibitijibibo:
I specifically removed SDL_image and replaced it with lodepng, so don’t worry about that part. Can also confirm tinyxml can’t really be touched for reasons I never figured out.

This is fine for now, but for future reference we probably won’t update dependencies until an official VVV update is coming up or a security vulnerability is exposed.

InfoTeddy:
> I specifically removed SDL_image and replaced it with lodepng

Why did you do that, then? It's a big headscratcher for anyone not familiar with the decision.

flibitijibibo:
One less DLL.

AlexApps99:
Can't it be statically linked?

AlexApps99:
Besides, AFAIK Steam provides DLLs for SDL2 so in most cases they wouldn't be needed

InfoTeddy:
Well, yes, SDL2 DLLs are in the Steam runtime, however there are plenty of non-Steam versions of the game such as Make & Play, where we would have to distribute DLLs anyway. I still agree though, can't it be statically linked?

flibitijibibo:
The base library could be, but then you’re also pulling in libpng and zlib as well. Point being, one is a single file with no configuration setup, the other is 3+ DLLs with autoconf and separate VS projects.

PR #183, opened 2020-03-13
Allow returning results from the earlier two SDL functions in Screen::ResizeScreen()

InfoTeddy:
## Changes:

* **Allow returning results from the earlier two SDL functions in `Screen::ResizeScreen()`**

  Any particular invocation of `Screen::ResizeScreen()` will end up calling three SDL functions, but there can only be one error code returned. My previous patch (#182) prioritized the last SDL function at all times, even if the earlier functions failed.

  So if the third fails, return the result of the third. Otherwise, if the second fails, return the result of the second. Else just return the result of the first.

  This is in sync with the fact that `SDL_GetError()` will prioritize the last failed function called, so the error code given is the one corresponding with `SDL_GetError()` in the end.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -92,6 +92,8 @@ Screen::Screen()
 int Screen::ResizeScreen(int x, int y)
 {
 	int result = 0; // 0 is success, nonzero is failure
+	int result2 = 0;
+	int result3 = 0;
 
 	static int resX = 320;
 	static int resY = 240;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Rather than accumulate the error codes we should probably do the full error checking (including SDL_GetError logs) inside that function, bailing early when something fails. That way it’s failing a little more gracefully and avoid the multiple variables.

InfoTeddy:
Alright, I'll try again with another PR, then.

PR #184, opened 2020-03-13
Do proper error handling inside Screen::ResizeScreen()

InfoTeddy:
## Changes:

* **Properly handle errors inside `Screen::ResizeScreen()`**

  Instead of passing the error codes out of the function, just handle the errors directly as they happen, and fail gracefully if something goes wrong instead of continuing.

  This should be done for every instance an SDL function is called, but this is a good start.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -89,10 +89,8 @@ Screen::Screen()
     glScreen = true;
 }
 
-int Screen::ResizeScreen(int x, int y)
+void Screen::ResizeScreen(int x, int y)
 {
-	int result = 0; // 0 is success, nonzero is failure
-
 	static int resX = 320;
 	static int resY = 240;
 	if (x != -1 && y != -1)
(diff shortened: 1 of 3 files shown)


desktop_version/src/Screen.cpp
```diff
@@ -184,11 +205,9 @@ void Screen::FlipScreen()
 	SDL_FillRect(m_screen, NULL, 0x00000000);
 }
 
-int Screen::toggleFullScreen()
+void Screen::toggleFullScreen()
 {
 	isWindowed = !isWindowed;
-	int result = ResizeScreen(-1, -1);
```
flibitijibibo: We still need the resize, right? Other than that lgtm

InfoTeddy: Haha, whoops, looks like I got a little too delete-happy.

InfoTeddy: Ok, fixed now.

PR #191, opened 2020-04-04
Refactor entities, blocks, and textboxes to not use the 'active' system and to not use separate length-trackers

InfoTeddy:
### Summary

This is a re-do of #174 but instead of everything being in one commit, it has been split into 28.

This PR removes the variables `obj.nentity`, `obj.nblocks`, `obj.nlinecrosskludge`, `graphics.ntextbox`, `script.txtnumlines`, and `textboxclass::numlines`, in favor of using `obj.entities.size()`, `obj.blocks.size()`, `obj.linecrosskludge.size()`, `graphics.textbox.size()`, `script.txt.size()`, and `textboxclass::line::size()`, respectively. Furthermore, all entclass, blockclass, and textboxclass objects no longer have an `active` attribute, meaning that they are guaranteed to be real and you don't have to check their `active` attribute before doing any operations with them.

See #174 for more information, I don't feel like re-typing or re-copy-pasting everything here.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/BlockV.cpp
+++ b/desktop_version/src/BlockV.cpp
@@ -34,3 +34,68 @@ void blockclass::rectset(const int xi, const int yi, const int wi, const int hi)
 	rect.w = wi;
 	rect.h = hi;
 }
+
+void blockclass::setblockcolour( std::string col )
+{
+	if (col == "cyan")
+	{
+		r = 164;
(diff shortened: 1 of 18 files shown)


flibitijibibo:
Once again, really good work and much clearer with the separate commits!

PR #201, opened 2020-04-17
Refactor the menu system and make it a little bit better

InfoTeddy:
### Summary

This PR refactors the menu system. The most important thing is that instead of having to fiddle with `menuoptions[0] = ...; menuoptionsactive[0] = true; menuoptions[1] = ...; menuoptionsactive[1] = true;` and `nummenuoptions = ...;`, you just do `option(...); option(...);`, and menu options are automatically defaulted to being true, and you don't have to fiddle with `nummenuoptions`. Oh, and menu options are no longer ad-hoc objects (I've removed the `menuoptionsactive` vector and converted `menuoptions` to a vector of a new struct `MenuOption`). Also, I've de-duplicated code relating to having possible permutations of menu options, converted menu names to enums, and converted as much "else if"-chains to case-switches as possible.

For reference, here is the list of possible menu permutations:

- Main menu
  - "start game" is gone with the `MAKEANDPLAY` define
  - "player levels" is gone with the `NO_CUSTOM_LEVELS` define
  - "view credits" is gone with the `MAKEANDPLAY` define

- "start game"
  - "secret lab" depends on `unlock[8]`

- "game options"
  - "unlock play data" is gone with the `MAKEANDPLAY` define
  - "soundtrack" depends on the existence of an `mmmmmm.vvv` file

- "player levels"
  - "level editor" is gone with the `NO_EDITOR` define

With that said, I've also made some changes not-so-behind-the-scenes that users can notice.

* **Remember selected menu option and previous menu name when hitting any "return" menu option**

  I've added a fancy system where a stack frame containing the selected menu option and name gets pushed onto a stack when you go into a menu, and then that stack frame is popped and restored when you return to the previous menu.

  So now for example, selecting "back to menu" in the "player levels" menu will put your cursor back on the "player levels" menu option in the main menu.

* **When pressing Esc in game.gamestate GAMEMODE and quitting, don't go back to main menu**

  So now if you were playing a custom level and quit to the title, you'd still be selecting the given custom level. If you were playing the main game, you'd still be in the play menu, etc.

* **Properly gray out the Secret Lab, Time Trial, and Invincibility options if they are disabled**

  If you have an accessibility option on like invincibility or slowdown, those options will be disabled, but they wouldn't be grayed out to properly indicate that they are disabled. So now they are.

* **Fix being able to mismatch the color of the quicksave/telesave summary and the menu text/background**

  Normally the text/background color is forced to be cyan when loading the quicksave/telesave summary, but a small oversight meant that if you pressed Esc to go to the `youwannaquit` menu, and then hit "no, return", it would've cycled the text/background color and you could've gotten a cyan summary box on top of a non-cyan text/background, which is a color mismatch and looks funny.

  But I fixed this.

* **Add a "previous page" option to the levels list**

  This makes the levels list much better to deal with if you happen to be a person who just keeps collecting levels and never clears out your level folder, or put levels in subfolders, or anything. (How do I ping everyone on the VVVVVV Discord server, again...?)

  This menu option turns into "last page" if it's the first page of the levels list.

  Also this feature was added to VCE but hasn't been upstreamed until now.

* **Add a "previous page" option to the credits menu**

  The credits menu isn't as bad as having a really huge levels list, but it's still nice to have, for consistency.

  Again, this menu option turns into "last page" if it's the first page of the credits.

## Legal Stuff:


flibitijibibo:
Everything looks good, but I’m just naturally wary of switch statements... missing break statements haunt my nightmares. But, I usually ban switch statements only in situations where I can’t force the compiler to error on missing breaks (C#, blarf). If we can add something to the CMake file that makes `-Wimplicit-fallthrough` a full error, that should prevent any nastiness from leaking in.

InfoTeddy:
Good thing you didn't immediately merge this, cuz I spotted some problems at the last second.

Anyway, my C++ linter already warns me about any implicit fallthroughs or missing `break` statements, so I'm pretty sure there are none in this PR.

However, I still added `-Werror=implicit-fallthrough` to the CMake file. People will have to re-`cmake` for the change to apply, but it's better than nothing.

flibitijibibo:
Looks like MSVC is fussy about the warning - we can just `if(NOT MSVC)` that block. I'm not aware of any MSVC equivalent of this exact warning so for now we'll skip it there, Clang/GCC should be good enough for testing.

flibitijibibo:
Merged, thanks! The original menu system was definitely... uh, "fragile".

PR #202, opened 2020-04-17
Optimize loading the levels list

InfoTeddy:
## Changes:

This PR optimizes loading the levels list when selecting "play a level" in the playerworlds menu.

Previously, it was *extremely* slow, because we kept parsing the *entire* XML document for every single level file, which gets pretty slow very quickly. In extreme cases, someone could have a cursed level file, like I did once, with 2<sup>18</sup> edentities inside it (don't ask<sup>note 1</sup>), making loading the levels list completely *choke* on it and temporarily take up a gigabyte of RAM. But if we no longer parse the entire XML document, bombs like these are avoided.

Instead, all we have to do is look for the metadata tags themselves, and what's contained between them. This is the same thing that Ved (an external VVVVVV level editor) does.

This optimization was added to VCE but hasn't been upstreamed until now, probably because Leo has a strange habit of using weird magic like exceptions and C++17 stuff for no good reason.

<sup>note 1</sup>: Ok, it was the result of a particularly nasty VCE bug that kept exponentially increasing the amount of edentities in a level, due to me screwing up when merging changes from upstream. The level file is 75 megabytes in size.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -189,7 +189,11 @@ void FILESYSTEM_loadFileToMemory(const char *name, unsigned char **mem,
 	{
 		*mem = (unsigned char*) malloc(length);
 	}
-	PHYSFS_readBytes(handle, *mem, length);
+	int success = PHYSFS_readBytes(handle, *mem, length);
+	if (success == -1)
+	{
+		FILESYSTEM_freeMemory(mem);
+	}
 	PHYSFS_close(handle);
 }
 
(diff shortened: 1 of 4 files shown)


TerryCavanagh:
This is a really great pull request, and is something I've wanted to add to VVVVVV for literally years. Thanks for putting it together!

leo60228:
> no good reason

`std::string::substr` is much less efficient than `std::string_view::substr`.

InfoTeddy:
> `std::string::substr` is much less efficient than `std::string_view::substr`.

Got any benchmarks for that? Especially the "much less efficient" part.

leo60228:
`std::string::substr` does a heap allocation and a memcpy. `std::string_view::substr` just copies a pointer and an integer to the stack.

PR #203, opened 2020-04-18
Add an "open level folder" menu option to the "play levels" menu

InfoTeddy:
## Changes:

This simply adds an "open level folder" option to the "play levels" menu, which was commented out and left unimplemented. If you press ACTION on it, it will open the level folder for you and minimize the game window. If it can't open the level folder (or the option is grayed out), it will play the sad sound effect.

Most of this code I stole from leo60228, and I kept that `#include <sys/types.h>` even though it seems to compile just fine without it because it sounds important or something.

This is yet another feature added to VCE that hasn't been upstreamed until now.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -19,6 +19,7 @@
 #if defined(_WIN32)
 #include <windows.h>
 #include <shlobj.h>
+#include <shellapi.h>
 int mkdir(char* path, int mode)
 {
 	WCHAR utf16_path[MAX_PATH];
@@ -33,6 +34,9 @@ int mkdir(char* path, int mode)
 /* These are needed for PLATFORM_* crap */
 #include <unistd.h>
 #include <dirent.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <spawn.h>
 #define MAX_PATH PATH_MAX
 #endif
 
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Functionally it seems okay, but I'm a little worried about how this will behave in steamcompmgr/gamescope - I'll try to test this sooner than later, but in case anyone is willing to make a Steambox to check this out, Ctrl+F "Bill's Hat" http://icculus.org/finger/flibitijibibo

InfoTeddy:
Hm, are you unable to minimize the window or open a folder on a Steambox?

Also not sure why Windows CI failed here, all I see is that it suddenly exited with code 1.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -465,3 +469,49 @@ void PLATFORM_copyFile(const char *oldLocation, const char *newLocation)
...
+
+#ifdef _WIN32
+bool FILESYSTEM_openDirectoryEnabled()
+{
+	return true;
+}
+
+bool FILESYSTEM_openDirectory(const char *dname)
+{
+	ShellExecute(Null, "open", dname, NULL, NULL, SW_SHOWMINIMIZED);
```
flibitijibibo, line 481: `NULL`, this is what tripped the test box!

flibitijibibo:
steamcompmgr is really odd, basically it takes the focused window and renders only that. It also has some peculiar rules in terms prioritizing Steam itself and an app that's run from Steam (custom atoms, blech), so the two things I'm not sure of:

- I think it just ignores Minimize, which is good only if...
- xdg-open will probably fail for this operation, unless it can somehow find a file picker for the `steam` user.

If it just happens to be a no-op there we're in good shape; if minimize is actually respected or xdg-open actually works we're in big trouble.

desktop_version/src/FileSystemUtils.cpp
InfoTeddy: Whoops, that's slightly embarrassing. Fixed and force-pushed.

flibitijibibo: If I had a dollar every time I swapped `null` and `NULL` when I was swapping between C and C#... >_>

InfoTeddy:
Hmm, and I wouldn't know how to detect that it's SteamOS and disable the option if so...

flibitijibibo:
A good multiplat fix for this would be to check for Steam Big Picture in general:

https://partner.steamgames.com/doc/features/steam_controller/getting_started_for_devs

> Start your game in fullscreen by default when the user is running Steam Big Picture (the "SteamTenfoot" environment variable will be set)

InfoTeddy:
Hm, what if the user is running it under Steam Big Picture on a normal desktop OS? Disabling this option would just confuse them for no reason.

flibitijibibo:
I don't recall if this is clearly documented in the Steamworks documentation, but there should still be a cert process for both Full Controller Support as well as Big Picture, and both launcher windows and file dialogs are both instant failure cases (I sorta got around the latter in TowerFall with [XNAFileDialog](https://github.com/flibitijibibo/XNAFileDialog)). VVVVVV technically still fails the controller cert because of the editor (yes, the editor still counts for cert) but I'd like to start going in the opposite direction for both of these if possible. Plus, if someone tries building for a non-desktop OS, they'll need a fallback path to work with.

InfoTeddy:
Alright, I'm fully disabling it for Big Picture, then.

leo60228:
> I'm pretty sure FreeBSD and OpenBSD and Haiku are POSIX enough that the "open" command will work on them, too.
Uh, the `open` command is non-standard, as far as I know. I just checked for Linux because the only Unixes VCE supports are Linux and macOS. Looks like Haiku and macOS use `open` and *BSD use `xdg-open`.

PR #204, opened 2020-04-18
Add analogue movement support using controller joystick

InfoTeddy:
## Changes:

This PR lets you use your gamepad joystick to move Viridian as much as you want them to move, instead of either doing "full speed ahead" or "no movement at all". This makes it easier and more fluid to use the joystick if you have a controller.

I verified with libTAS that this does not change any keyboard-only input, meaning it should be the same as before unless you use a controller.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Ent.cpp
+++ b/desktop_version/src/Ent.cpp
@@ -21,6 +21,8 @@ entclass::entclass()
 	ay = 0;
 	vx = 0;
 	vy = 0;
+	mx = 1;
+	my = 1;
 	w = 16;
 	h = 16;
 	cx = 0;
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Seems like a good idea to me, but I'll defer to @TerryCavanagh on this one... not sure if it's critical how quickly Viridian moves.

TerryCavanagh:
Hey, sorry, not going to merge this one. VVVVVV is a 10 year old game, I think it's important that changes at this point don't change anything fundamental about the core game - which I think includes improvements to how it feels to play. I think this option should be a CE only thing.

PR #215, opened 2020-04-29
Fix editor menu being drawn on top of editor with BGs disabled

InfoTeddy:
## Changes:

Previously, if you had backgrounds disabled in accessibility options, and went to the editor and opened up the editor menu, it would be drawn straight on top of what was already there in the editor instead of being drawn on top of black. So now it's drawn on top of black.

## Legal Stuff:

  for these changes


flibitijibibo:
Nothing quite like an overdraw bug!

PR #216, opened 2020-04-29
Optimize descending tower background drawing

InfoTeddy:
## Changes:

When working on my over-30-FPS patch, I noticed that the descending tower background kept getting re-drawn every frame, which is kind of inefficient. Also, it'd be better for my over-30-FPS patch if the background scrolled whatever it had (at least until the color change), because that way would be much easier to interpolate than having to redraw the background every frame.

And since the background is now scrolling instead of being redrawn, when the background *does* need to be redrawn anyway, it needs to account for `map.bscroll` to be aligned with the incoming new textures. But only for the descending background.

Also, I merged `Graphics::drawtowerbackgroundsolo()` into `Graphics::drawtowerbackground()` because it'll mean I have to copy-paste less for my over-30-FPS patch.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -2201,43 +2201,6 @@ void Graphics::drawtowerspikes()
     }
 }
 
-void Graphics::drawtowerbackgroundsolo()
-{
-    if (map.bypos < 0)
-    {
-        map.bypos += 120 * 8;
-    }
-
-    int temp = 0;
-
-    if (map.tdrawback)
-    {
-        //Draw the whole thing; needed for every colour cycle!
-        for (int j = 0; j < 31; j++)
-        {
-            for (int  i = 0; i < 40; i++)
-            {
-                temp = map.tower.backat(i, j, map.bypos);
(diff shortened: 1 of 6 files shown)


desktop_version/src/Graphics.cpp
```diff
@@ -2268,8 +2266,18 @@ void Graphics::drawtowerbackground()
         ScrollSurface(towerbuffer, 0, -map.bscroll);
         for (int i = 0; i < 40; i++)
         {
-            temp = map.tower.backat(i, 0, map.bypos);
-            drawtowertile3(i * 8, -(map.bypos % 8), temp, map.colstate);
+            if (map.scrolldir == 0)
```
flibitijibibo, line 2233: Should we put this branch outside of the loop?

InfoTeddy: If it's faster, then sure.

InfoTeddy: Done, rebased and force-pushed.

flibitijibibo:
Most compilers would probably have gotten in, it's the _other_ most that I worry about a lot... everything else looks good, thanks!

Issue #99, opened 2020-01-19
60 FPS?


ClementineAccount:
Would 60 fps be possible now?

flibitijibibo:
No, someone still has to make the game run in a more accurate fixed timestep loop and then fix all the 33ms dependencies. So far we haven’t had any volunteers.

TerryCavanagh:
This is possible, but it’s a lot more work than it appears at first. Reposting something I said about this on discord:

![97647D00-19F3-46FC-BD3F-1A8D55804B7C](https://user-images.githubusercontent.com/2874150/72682201-dcc62780-3aca-11ea-962d-844ee9c76b65.jpeg)

InfoTeddy:
Someone actually [did make a 60 FPS patch](/Colon-D/VVVVVV-over-30-FPS).

That said, it takes the approach of keeping timestep the same, but then interpolating the animations. So basically, the animations are at 60 FPS but the game is still actually at 30 FPS. That said...

1. **The cutscene bar animations go by way too quickly.** My guess is that they're somehow based on deltatime now instead of amount of frames? Because if I run the mod under the TAS tool libTAS, it doesn't seem to happen... I'm really not sure what this guy did with them.

   This is an issue because the `untilbars()` command _directly depends_ on the animation of the cutscene bars. This means the timings of any script that uses cutscene bars are changed dramatically.

   Not only do the cutscene bars go by too quickly, the flash effect that makes the screen white _also_ goes by too quickly. What did this guy do with time management in this game?

   Speaking of running the mod under libTAS...

2. **If I run the mod under libTAS at 60 FPS fully (1000/17 FPS), libTAS can't advance time and the game is stuck.** But it works fine if I run it under 30 FPS (1000/34 FPS).

   I'm not sure what else to say. I have no idea what this guy did to the time management in VVVVVV, but I can't run the mod at 60 FPS in libTAS or else I'll never get past the "Press ACTION to start" screen.

3. **Somehow, you can't edge-flip off of a moving platform.** I didn't think this would be an issue, because edge-flipping off of regular tiles seems to be fine and untouched.

   Usually, you can flip off of a moving platform on the frame you leave it, but for some reason in this mod, you can't. This means that my Dimension Open Any% TAS desyncs with it, because at one point in the TAS Viridian edge-flips off of a moving platform.

There are more issues, but I think you get the gist of it.

InfoTeddy:
[...]

PR #220, opened 2020-05-03
Make the game (visually) run at framerates above ~30 FPS

InfoTeddy:
## Changes:

VVVVVV is a game that runs at 1000/34 frames per second, which is approximately 29.4 frames a second, which is close to, but not quite, 30 frames a second. This odd number comes from the fact that the amount of times it takes for each frame is coded to be 34 milliseconds, and not 33.333333333... milliseconds.

This pull request unlocks VVVVVV's framecap of 1000/34 FPS. It does this in such a way that the underlying engine still runs at an actual framerate of 1000/34 FPS (which we should be calling tickrate now), and in between each actual frame we just draw what the frames in between each frame *should* be, using something called "linear interpolation". So now the game will visually run at the same rate as your monitor (my monitor is 60hz, I'm not a true gamer :( &#x200b; &#x200b; &#x200b; &#x200b; ).

What's linear interpolation, you might ask? Well it's basically like drawing a straight line between two points on a graph. But it's more akin to averaging something. All the rendering does is go like, "Ok so on this frame, we have 2, and on the next frame, we have 4. What's the average of 2 and 4? Why, it's 3! So we should draw a frame using 3."

This approach means that the underlying physics isn't changed, which is a very good thing because if I changed the physics around to update at, say, twice the speed but with half the values, that would run the risk of breaking custom levels. Also, it's easier to do it this way because it requires changing less values, although it does require disentangling all pieces of update code from render code so as to make sure things don't update as quickly as possible. And regardless of those concerns, this approach is the cleanest imo, since the physics don't subtly change depending on your refresh rate (unlike games like Super Meat Boy, which apparently [desyncs TASes depending on if your refresh rate is 59hz or 60hz](/clementgallet/libTAS/issues/220), oh and also different resolutions have different refresh rates, good luck).

Oh, and by the way, the Switch version (made by some weird company named Nicalis) actually does the screw-up-the-physics approach instead of what I'm doing. They've apparently had to change some custom levels because the way they did it broke things like edge-flipping. Maddening.

Most of the work I took was from [Colon-D's fork of VVVVVV attempting to run it at over 30 FPS](/Colon-D/VVVVVV-over-30-FPS). However, his mod is incomplete (for example, the title screen still runs at 1000/34 FPS) and also does some things inefficiently (like re-drawing warp backgrounds every single frame, instead of scrolling a surface + incoming textures after drawing an initial background). Also it hasn't been updated in months.

This patch also seems to improve the 100 milliseconds of `SDL_Delay` the unfocus-pause gives, in that it actually seems to have an effect now and be noticeable when you re-focus (possibly because before we were using `SDL_Delay` as the frame limiter, too, and I guess one of them kind of canceled the other out?).

To enable this, you need to go into "graphic options" and turn on the "toggle fps" option. You can turn on "toggle vsync" as well if you feel like it, too.

Honestly, after playing the game at 60 FPS, it feels really strange to play the game at normal 30 FPS.

#99 is already closed, but seems relevant to mention in this PR.

(Current base commit ID for this PR: 62441edbc9b279d4e37ea6a640f376288ffc8417)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Ent.cpp
+++ b/desktop_version/src/Ent.cpp
@@ -53,6 +53,8 @@ entclass::entclass()
 	walkingframe = 0;
 	dir = 0;
 	actionframe = 0;
+
+	realcol = 0;
 }
 
 bool entclass::outside()
@@ -583,6 +585,53 @@ void entclass::settreadmillcolour( int rx, int ry )
 		break; //Red
 	default:
 		return;
+		break;
+	}
+}
+
+void entclass::updatecolour()
+{
+	switch (size)
+	{
+	case 0: // Sprites
+	case 7: // Teleporter
+	case 9: // Really Big Sprite! (2x2)
+	case 10: // 2x1 Sprite
+	case 13: // Special for epilogue: huge hero!
+		graphics.setcol(colour);
+		realcol = graphics.ct.colour;
+		break;
+	case 3: // Big chunky pixels!
+		realcol = graphics.bigchunkygetcol(colour);
+		break;
+	case 4: // Small pickups
+		graphics.huetilesetcol(colour);
+		realcol = graphics.ct.colour;
+		break;
+	case 11: // The fucking elephant
+		if (game.noflashingmode)
+		{
+			graphics.setcol(22);
+		}
+		else
+		{
+			graphics.setcol(colour);
+		}
+		realcol = graphics.ct.colour;
+		break;
+	case 12: // Regular sprites that don't wrap
+		// if we're outside the screen, we need to draw indicators
+		if ((xp < -20 && vx > 0) || (xp > 340 && vx < 0))
+		{
+			graphics.setcol(23);
+		}
+		else
+		{
+			graphics.setcol(colour);
+		}
+		realcol = graphics.ct.colour;
+		break;
+	default:
 		break;
 	}
 }
--- a/desktop_version/src/Ent.cpp
+++ b/desktop_version/src/Ent.cpp
@@ -223,7 +223,9 @@ void entclass::setenemyroom( int rx, int ry )
 			w = 16;
 			h = 16;
 			xp -= 24;
+			oldxp -= 24;
 			yp -= 16;
+			oldyp -= 16;
 		}
 		else
 		{
@@ -236,7 +238,9 @@ void entclass::setenemyroom( int rx, int ry )
 			cx = 4;
 			size = 9;
 			xp -= 4;
+			oldxp -= 4;
 			yp -= 32;
+			oldyp -= 32;
 		}
 
 		break;
@@ -337,6 +341,7 @@ void entclass::setenemyroom( int rx, int ry )
 		w = 32;
 		h = 14;
 		yp += 1;
+		oldyp += 1;
 		break;
 	case rn(16, 2): // (Manequins)
 		tile = 52;
@@ -345,6 +350,7 @@ void entclass::setenemyroom( int rx, int ry )
 		w = 16;
 		h = 25;
 		yp -= 4;
+		oldyp -= 4;
 		break;
 	case rn(18, 0): // (Obey)
 		tile = 51;
--- a/desktop_version/src/Ent.h
+++ b/desktop_version/src/Ent.h
@@ -1,6 +1,8 @@
 #ifndef ENT_H
 #define ENT_H
 
+#include "Graphics.h"
+
 #define		rn( rx,  ry) ((rx) + ((ry) * 100))
 
 class entclass
@@ -15,6 +17,8 @@ public:
     void setenemyroom(int rx, int ry);
 
     void settreadmillcolour(int rx, int ry);
+
+    void updatecolour();
 
 public:
     //Fundamentals
@@ -45,6 +49,8 @@ public:
     //Animation
     int framedelay, drawframe, walkingframe, dir, actionframe;
     int yp;int xp;
+
+    Uint32 realcol;
 };
 
 #endif /* ENT_H */
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2008,6 +2008,10 @@ void entityclass::createentity( float xp, float yp, int t, float vx /*= 0*/, flo
     }
 
     entity.drawframe = entity.tile;
+    if (!entity.invis)
+    {
+        entity.updatecolour();
+    }
 
     entities.push_back(entity);
 }
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -62,6 +62,7 @@ void entityclass::init()
     customenemy=0;
     customwarpmode=false; customwarpmodevon=false; customwarpmodehon=false;
     trophytext = 0 ;
+    oldtrophytext = 0;
     trophytype = 0;
     altstates = 0;
 
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2376,6 +2376,7 @@ bool entityclass::updateentities( int i )
                 {
                     entities[i].statedelay = 6;
                     entities[i].xp -=  int(entities[i].para);
+                    entities[i].oldxp -=  int(entities[i].para);
                 }
                 break;
             case 18: //Special for ASCII Snake (right)
@@ -2383,6 +2384,7 @@ bool entityclass::updateentities( int i )
                 {
                     entities[i].statedelay = 6;
                     entities[i].xp += int(entities[i].para);
+                    entities[i].oldxp += int(entities[i].para);
                 }
                 break;
             }
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2012,6 +2012,32 @@ void entityclass::createentity( float xp, float yp, int t, float vx /*= 0*/, flo
     {
         entity.updatecolour();
     }
+    if (entity.type == 1)
+    {
+        switch (entity.animate)
+        {
+        case 0: // Simple Loop
+        case 1: // Simple Loop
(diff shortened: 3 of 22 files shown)


flibitijibibo:
Wow, that was fast!

It'll take me a while to get through the technical stuff, but @TerryCavanagh will be the expert on accuracy here, so I'll defer to him on how it plays and what doesn't feel right. I'm going to guess and say we'll want this off by default for preservation's sake, but it'll probably resolve the biggest complaint the game's ever had if we at least have the option...

InfoTeddy:
> I'm going to guess and say we'll want this off by default for preservation's sake

Yeah, I figured we should have a framerate-limiting option. I tried re-adding the old frame limiting system in, but it felt a bit dirty to be doing that, what with the `SDL_Delay` and all. I'm thinking the option should just be a toggle between "30" FPS and v-sync, no?

flibitijibibo:
Makes sense to me! I’ll start going through the commits on Monday.

flibitijibibo:
Ran through all the commits and they seem pretty solid - some of them could probably be split off into their own pull request but it's not really important.

At this point, we need people to test this _really_ thoroughly - if you're reading this and know how to compile the game, that means you. As long as it looks and plays sanely, both in 60 and 30 mode, this should be good to go.

InfoTeddy:
I fixed the preloader. I wanted to interpolate it until I found out that there's no movement and it's actually somewhat random each time, so. Also I rebased onto the latest upstream and force-pushed.

InfoTeddy:
@flibitijibibo Would applying #22 (after fixing conflicts) be a good method of v-sync? Also I'm thinking now that v-sync should be its own separate option, and 30/above-30 another one. If so then I can't be using sleep() -> update() -> sleep() to limit the framerate to 30.

flibitijibibo:
If it's possible to have both that'd work for me. The way I _like_ to have it in games (when possible) is to have both a fixed-step and vsync option, since fixed-step is good for determinism while vsync is good for visuals, and being able to toggle both cases is good for prioritizing accuracy vs. visuals without getting stuck with stuttering or weird lerpy glitches, for machines where their display isn't _exactly_ the same as the timestep. Dust: AET is usually my go-to example for this. For this game, having 30 or 30+ as well as sync/tear makes sense, since there are reasonable use cases for all 4 combinations:

30+Sync: Original visuals on a compatible monitor, no tearing
30+Tear: Original visuals and avoids stuttering
Unlocked+Sync: The ideal scenario
Unlocked+Tear: Better than 30+Tear visually and avoids stuttering, just maybe don't do this on a hot day...

ThatWeirdAndrew:
I've played both version for about 30 minutes now and have not noticed any issues so far.

AllyTally:
I have mentioned this on the VVVVVV Discord server, but it seems possible to get teleported to (20, 20) by going through a room that wraps. [Here's a video](https://o.lol-sa.me/2mMWFmd.mp4), and [this is possible with it](https://o.lol-sa.me/4onmV45.png). It's a pretty game breaking bug, as I have been teleported to (20, 20) multiple times while trying to play through the Warp Zone normally.

I'm making this a comment here because from what I can tell, this only happens for this pull request. I also compiled it on Windows 10 (which is how I found #224)

InfoTeddy:
I have also mentioned this on the VVVVVV Discord server, but this sounds like plain old Undefined Behavior that may just manifest from shifting memory around the right way.

parkerlreed:
Looking good! 500FPS VVVVVV is not something I ever expected to see 

image

parkerlreed:
Pointed out on Discord but should put it here for posterity. 

Intel GPU on Linux with Mesa doesn't Vsync to monitor 
image

flibitijibibo:
The vsync option can be labeled as an SDL issue, the problem is that SDL_Render does not listen to the hint as it should:

https://hg.libsdl.org/SDL/file/fdabcc76aaaf/src/render/SDL_render.c#l789
https://hg.libsdl.org/SDL/file/fdabcc76aaaf/src/render/opengl/SDL_render_gl.c#l1648

A workaround would be recreating the renderer, but that's a sloppy fix for something that SDL can easily listen for with an SDL_HintCallback and send to the backends, all of which trivially support dynamic swap intervals. I haven't checked if this is filed on the Bugzilla, but if not we should get that done on SDL's end.

leo60228:
I'm not sure if "as it should" is the right thing to say here. SDL_HINT_RENDER_VSYNC is only used by SDL_CreateRenderer, and the documentation doesn't really suggest otherwise.

InfoTeddy:
(Just rebased onto latest master and force-pushed, so #224 now applies to this branch.)

leo60228:
I'm trying to fix this in SDL, but I haven't used Mercurial before, so that's making it somewhat annoying. For now, maybe just change `main` to:
```c++
        game.init();
        graphics.init();
        game.loadstats();
        Screen gameScreen;
        graphics.screenbuffer = &gameScreen;
```
and add a note saying that a restart is required. Though, `game.loadstats()` mutates `gameScreen`... Maybe add null checks and call it twice?

InfoTeddy:
(Just rebased onto latest master and force-pushed, so #227 now applies to this branch.)

InfoTeddy:
(Just rebased and force-pushed, this is to fix the `<vsync>` option in `unlock.vvv` not being loaded properly.)

leo60228:
https://bugzilla.libsdl.org/show_bug.cgi?id=5123

InfoTeddy:
(Just rebased onto latest master, so #228 now applies to this PR.)

InfoTeddy:
(Rebased onto latest master, so #229 now applies to this patch.)

InfoTeddy:
I've rebased onto latest master, so #230 and #231 now apply to this PR. Additionally I've also made sure no interpolation occurs in 30-FPS-only mode, and fixed an off-by-one in the fade-to-menu/lab delays.

InfoTeddy:
Rebased and force-pushed, #232, #234, and #235 now apply to this branch. (Also a re-fix of #235 due to game loop changes.)

InfoTeddy:
(Rebased, #238 now applies.)

InfoTeddy:
(Rebased, #239, #240, and #241 now apply.)

InfoTeddy:
(Rebased, #242 applies.)

InfoTeddy:
(Rebased, #243, #244, and #245 now apply.)

InfoTeddy:
(Rebased, #246 and #247 now applies.)

InfoTeddy:
(Rebased, #248, #249, #250, #251, and #252 apply.)

InfoTeddy:
(Rebased, #143 now applies.)

InfoTeddy:
(Rebased, #253, #254, and #255 apply.)

InfoTeddy:
(Rebased, #243 applies.)

leo60228:
Rebasing every time master changes creates a bunch of pointless notifications. Maybe you could rebase locally and only push when something actually needs it? Or push the rebases to a different branch and `git reset --hard other-branch && git push --force`?

flibitijibibo:
I haven't been getting notifications other than posts saying that the branch has been rebased... is there a specific setting that notifies per-commit? I do get emails for unique commits, but force-pushes don't show up for me.

leo60228:
Those are the ones I'm getting. Now that you mention it, maybe they should just be edited into the PR itself?

InfoTeddy:
Alright, I'll get rid of all the rebase posts and put a base commit ID in the description. Every time I rebase, I'll edit the description to update that commit ID.

Fussmatte:
I've went through the first few levels of Dimension Open with this version and it works awesome! I don't notice any issue so far.

flibitijibibo:
Making a last call for testing this! I'll be running through the game this week, and if I find nothing wrong I'll merge this on the 19th.

flibitijibibo:
Did a 100% run on a single program instance and everything looked fine to me! I did notice that when running with Valgrind, performance was worse no matter which mode I was running with. On an i7 2600K I can maintain 30fps with Valgrind using the master branch, but couldn't get the same with this branch. We could probably get those frames back by optimizing BlitSurfaceColoured (it's a very particular type of blend operation though, blech), but it'd be worth checking this branch out on an obscenely bad PC first to make sure we're not accidentally bumping the system requirements.

InfoTeddy:
Mm, that could be down to no longer using `SDL_Delay()` to limit the framerate. I'm pretty sure that in 30-FPS mode, it does the rendering at 30 FPS too (it doesn't like, render unnecessary delta frames), so I'm not sure where the performance drop comes from.

I tested it out in Valgrind myself. On master, it's very stuttery 30 FPS, and on this branch it seems to be 15 FPS or so, in both modes. (I had to make `musicclass::play()` and `musicclass::playef()` do nothing though, otherwise Valgrind complains about an illegal opcode in Vorbis stuff in SDL2_mixer.) Although on both branches, there's a very large startup time (but there's also a very large startup time whenever I run the game under GDB, too, so).

flibitijibibo:
>Mm, that could be down to no longer using SDL_Delay() to limit the framerate.

Thinking about it, that's probably why the music thread is falling behind too; the sleep probably gives the VM the room it needs to mix everything in time. As long as a crappy PC doesn't die with the changes that's good enough for me!

InfoTeddy:
I have a terrible enough computer that zip decompression hurts to do...
Zip decompression hurts my computer
...but this patch runs fine on my system. So I think it's fine.

flibitijibibo:
Beautiful. So as long as nobody finds anything else this week we're good to go for Friday's merge.

parkerlreed:
What option enables this for master?

InfoTeddy:
Go to Graphic Options, there will be an option labeled Toggle FPS.

parkerlreed:
Oh I see over 30 fps in in the graphics options. Sweet

PR #224, opened 2020-05-04
Add bounds checks to functions taking in entity/blocks/linecrosskludge indices

InfoTeddy:
## Changes:

The main ones to beware of here are `entityclass::updateentities()`, `entityclass::updateentitylogic()`, and `entityclass::entitymapcollision()`. They would index out-of-bounds and thus commit Undefined Behavior if the entity was removed in `entityclass::updateentities()`.

And it would've been fine enough if I only added bounds checks to those functions. However, I decided to be a bit more defensive and play it safe, and added bounds checks to *all* functions taking in not only an entity indice, but also blocks and linecrosskludge indices. Just to ensure that nothing like this happens again.

Thanks to @AllyTally for helping debug this UB.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -7,6 +7,7 @@ bool entityclass::checktowerspikes(int t)
 {
     if (t < 0 || t >= (int) entities.size())
     {
+        puts("checktowerspikes() out-of-bounds!");
         return false;
     }
 
@@ -1040,6 +1041,7 @@ void entityclass::removeentity(int t)
 {
     if (t < 0 || t > (int) entities.size())
     {
+        puts("removeentity() out-of-bounds!");
         return;
     }
     entities.erase(entities.begin() + t);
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Do these happen regularly? If not, we can add SDL_asserts as well, so they pop up blatantly in debug builds.

InfoTeddy:
These happen when a trinket or custom crewmate gets collected by the player. Its state gets set to 1 upon collision, then it gets removed in `entityclass::updateentities()`, and then the update functions taking place after that end up referencing a non-existent entity.

flibitijibibo:
Gotcha - might be worth logging instead of asserting then, just so it's clear that it does happen in known scenarios but should be unexpected 99% of the time.

InfoTeddy:
53cf52f6d5901d9c4e4eff134a70473b7bf9e790 adds a `puts()` to every function except `updateentitylogic()` and `entitymapcollision()`.

flibitijibibo:
Perfect, thanks!

PR #230, opened 2020-05-07
Fix being able to interrupt the exiting-to-menu fadeout and remove pressing Esc when the screen is fully black taking you to the main menu no matter what

InfoTeddy:
Currently, quitting to the menu is kind of a kludge-filled mess. There are numerous issues with it:

1. You can sometimes interrupt the exit-to-menu fadeout. This seems to happen most often in towers, and it does so without you having to do any crazy hijinks or anything. It just always gets interrupted in towers for whatever reason.

2. As a hacky workaround, if the fadeout has been interrupted and you're on a black screen, you can press Esc to manually resume the fadeout and get back to the title screen. This "fix" seems to have been made somewhat intentionally.

3. ...But this hacky workaround is *always* active whenever you're on a black screen, and *not* just when you're fading out to the menu and getting interrupted. If you're fully blacked out during a cutscene, make sure not to hit Esc accidentally, or you'll be booted back to the title screen!

4. Oh, and it seems like this hack is *so* pervasive that pressing Esc while you're faded out *in playtesting mode* won't return you to the editor, it'll return you to the main menu! Hope you didn't need to save your hard work or anything...

5. Also, hm, why do activity zone prompts, "- Press ENTER to Teleport -", and trophy text show up during the fadeout, if you're standing in one? They don't show up on the pause screen, so why here?

What you're about to see is that exiting to the menu is a giant janky Rube Goldberg kludge, mostly because for some reason it just *insists* on using Terry's State Machine to do its work, and then the entire code path has to route around all the problems that causes. Oh and there's an aggressive `script.hardreset()` right as you press ACTION to quit, which doesn't help either.

Here's how the current system does it:

1. When you press ACTION on "yes, quit to menu", `script.hardreset()` gets called and the game starts fading out (fademode 2).

2. Then when the game is fully faded out (fademode 1), it sets the gamestate to 80. However, we have a problem. MAPMODE doesn't have a `game.updatestate()`, the only time it's used is in `gamelogic()`. (And we probably shouldn't put a `game.updatestate()` in there anyway, doing that would mean the gamestate wouldn't be paused on the quit/pause/teleporter screens.)

   So what does the game do instead? It sets `graphics.resumegamemode` to true. This is simply a variable that signals "okay, we should bring the pause screen down".

3. Then the code for bringing the pause screen down is ran. It has to wait through the *entire* bringing-menu-down animation before the game-gamestate (which is `game.gamestate`, and is *not* the gamestate for Terry's State Machine, which is `game.state`) is *finally* set to GAMEMODE. This results in 15 unnecessary frames of a black screen.

4. Alright, we're now in GAMEMODE, we can run `game.updatestate()` and then get booted back to the title screen.

So in summary: the game wants to use gamestate 80, however it can't run gamestates on the pause screen (oh and on a side note, it also aggressively runs `script.hardreset()` right when you press ACTION to quit to the menu, which we'll get to later). So what it does instead is it brings down the pause screen after the screen fades out, thus being able to run the gamestate. To summarize, the game *has* to bring down the pause screen in order to be able to go back to the title screen, which is a direct consequence of using the state machine.

And with that, let me explain all the numerous issues above:

 - You can interrupt the fadeout, because it takes you back to GAMEMODE for a short bit in order to run it (the return-to-menu gamestate) in `gamelogic()`. There are many, *many* other things that use the state machine, such as script boxes, teleporters, both gravitrons. By going back to GAMEMODE like this, any number of things could interrupt the fadeout.

   But the case is especially bad in towers, mostly due to the aggressive `script.hardreset()`. One of the things that `script.hardreset()` does is turn `map.towermode` off.

   This then means that when `gamelogic()` is run, the game thinks that you're no longer in a tower, and when it takes a look at your y-position, it thinks "Oh crap you're wayyyy off-screen! You should be going to the adjacent room over!"

   Then it calls `map.gotoroom()`. One of the things that `map.gotoroom()` does is set the gamestate to 0 (which is simply a state for doing nothing):

   ```cpp
    //a very special case: if entering the communication room, room 13,4 before tag 5 is set, set the game state to a background
    //textbox thingy. if tag five is not set when changing room, reset the game state. (tag 5 is set when you get back to the ship)
    if(!game.intimetrial && !custommode)
    {
        if (!obj.flags[5] && !finalmode)
        {
            game.state = 0;
            if (game.roomx == 113 && game.roomy == 104)
            {
                game.state = 50;
            }
        }
    }
   ```

   So wait, if you've already entered Comms Relay (flag 5 would be on), or you're in Outside Dimension VVVVVV (`finalmode` would be on), or you're in a time trial, or you're in a custom level, the gamestate wouldn't be set to 0, right?

   Wrong. Remember that `script.hardreset()` from earlier? Yeah, it means that (1) `game.intimetrial` is off, (2) `finalmode` is also off, (3) flag 5 is also off, and (4) `custommode` is also off at this point in time.

   *If a `gotoroom()` happens while you're fading out to the menu, your fadeout will be interrupted.*

   So all that means is that the only time your fadeout *won't* be interrupted when exiting to the menu from inside a tower is when you're at the top of it. This explains issue 1.

 - Step 2 (doing something when the game is fully blacked out) fully explains issues 2, 3, and 4. The thing is, MAPMODE checks if the game is fully faded out *at all times*, even when you're bringing up or down the quit/pause/teleporter screen.

   If your fadeout has been interrupted and you're fully blacked-out, when you press Esc, all you're doing is just moving the game back to the MAPMODE game-gamestate. Then you're letting MAPMODE run its course, and it notices that the screen is fully blacked out, and it thinks "Well this must mean we're going back to the main menu." This janky workaround wasn't intentionally added at all, it naturally arose from this entire Rube Goldberg kludge of code!

   Same thing when pressing Esc in playtesting mode (here you can press Enter too because Enter also returns to the editor). To return to the menu from the editor, EDITORMODE fades out, and then uses fademode 1 to go back to TITLEMODE. When you exit playtesting while in fademode 1, it isn't reset, and so EDITORMODE thinks that you're in the middle of fading out to the title screen, and thus goes back to the title screen.

 - And since fading to the menu requires going back to GAMEMODE, that's why "- Press ENTER to Teleport -", activity zone prompts, and trophy text show up when you're fading out. Issue 5 is explained.

And now here's what I do to fix all these issues:

 - Exiting to the menu no longer relies on `graphics.fademode` nor `game.state`. Instead it uses `game.fadetomenu` (bool) and `game.fadetomenudelay` (int). (Same with returning to the Secret Lab when in the Super Gravitron since there's an obvious visual glitch when you do so, which I also fixed. It uses variables `game.fadetolab` (bool) and `game.fadetolabdelay` (int).)

   - This means that `fademode` is no longer relied upon, so if you're blacked out when you press Esc to bring up the quit screen, you won't suddenly quit to the menu. And this means you can no longer accidentally press Esc *during a cutscene* to immediately return to the title screen without warning (well, you can still accidentally press Esc, but at least there's a prompt).

   - More importantly, it also means you can no longer interrupt the exit-to-menu fadeout, since the logic is all self-contained in MAPMODE and the game no longer has to go back to GAMEMODE at all!

 - Also, the aggressive `script.hardreset()` *right* as you press ACTION has got to go. I moved it to right when the game switches back to the title screen, after all the fading out is done. That way all the menu kludge variables are gone, too, which is cleaner.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -1336,44 +1336,7 @@ void Game::updatestate()
             if(graphics.fademode == 1)	state++;
             break;
         case 81:
-            gamestate = TITLEMODE;
-            graphics.fademode = 4;
-            music.play(6);
-            graphics.backgrounddrawn = false;
-            map.tdrawback = true;
-            graphics.flipmode = false;
-            //Don't be stuck on the summary screen,
-            //or "who do you want to play the level with?"
-            //or "do you want cutscenes?"
-            //or the confirm-load-quicksave menu
-            if (wasintimetrial)
-            {
-                returntomenu(Menu::timetrials);
-            }
-            else if (wasinintermission)
-            {
-                returntomenu(Menu::intermissionmenu);
-            }
-            else if (wasinnodeathmode)
-            {
-                returntomenu(Menu::playmodes);
-            }
-            else if (wasincustommode)
-            {
-                returntomenu(Menu::levellist);
-            }
-            else if (save_exists() || anything_unlocked())
-            {
-                returntomenu(Menu::play);
-            }
-            else
-            {
-                createmenu(Menu::mainmenu);
-            }
-            wasintimetrial = false;
-            wasinintermission = false;
-            wasinnodeathmode = false;
-            wasincustommode = false;
+            quittomenu();
             state = 0;
             break;
 
@@ -7357,4 +7320,46 @@ bool Game::anything_unlocked()
 bool Game::save_exists()
 {
     return telesummary != "" || quicksummary != "";
+}
+
+void Game::quittomenu()
+{
+    gamestate = TITLEMODE;
+    graphics.fademode = 4;
+    music.play(6);
+    graphics.backgrounddrawn = false;
+    map.tdrawback = true;
+    graphics.flipmode = false;
+    //Don't be stuck on the summary screen,
+    //or "who do you want to play the level with?"
+    //or "do you want cutscenes?"
+    //or the confirm-load-quicksave menu
+    if (wasintimetrial)
+    {
+        returntomenu(Menu::timetrials);
+    }
+    else if (wasinintermission)
+    {
+        returntomenu(Menu::intermissionmenu);
+    }
+    else if (wasinnodeathmode)
+    {
+        returntomenu(Menu::playmodes);
+    }
+    else if (wasincustommode)
+    {
+        returntomenu(Menu::levellist);
+    }
+    else if (save_exists() || anything_unlocked())
+    {
+        returntomenu(Menu::play);
+    }
+    else
+    {
+        createmenu(Menu::mainmenu);
+    }
+    wasintimetrial = false;
+    wasinintermission = false;
+    wasinnodeathmode = false;
+    wasincustommode = false;
 }
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -372,6 +372,8 @@ void Game::init(void)
 
     fadetomenu = false;
     fadetomenudelay = 0;
+    fadetolab = false;
+    fadetolabdelay = 0;
 
     /* Terry's Patrons... */
     superpatrons.push_back("Anders Ekermo");
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -135,7 +135,6 @@ void Game::init(void)
 
     teleportscript = "";
     savemystats = false;
-    menukludge = false;
     quickrestartkludge = false;
 
     tapleft = 0;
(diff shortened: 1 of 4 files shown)


InfoTeddy:
Heh, actually, it looks like `map.towermode` being set to false in `scriptclass::hardreset()` was actually not in 2.2! It was added in 9296547febab942a6ef6bd3a13207894e641b1ba by @FredrIQ. Looks like in an attempt to fix a bug, you end up inexplicably worsening another bug in the process with this game.

FredrIQ:
The commit was done for a reason -- to fix tower quicksaves. I was not aware that quitting the game would run the "initialize variables to prepare for starting the game", because doing so wouldn't make any sense. But in hindsight, this would explain why it's so easy to cause the problem in widely different ways -- standing between edges in a room that warps (which I think might be 3DS-only -- presumably it resets the warp background, causing you to gotoroom which causes problems as per above), hurting yourself (respawn would call gotoroom), exiting near edges of a room, etc.

InfoTeddy:
> standing between edges in a room that warps (which I think might be 3DS-only -- presumably it resets the warp background, causing you to gotoroom which causes problems as per above)

I was about to say, "Haha, Nicalis coding be like" but then I checked 2.2 code. `map.warpx` and `map.warpy` are both set to false in `scriptclass::hardreset()`. I tried it just now on 2.2, and indeed standing between warp edges also works to interrupt an exit-to-menu fadeout on PC.

PR #242, opened 2020-05-18
Refactor all tilemaps to no longer be stored in strings

InfoTeddy:
This refactors the main game tilemaps so that they are no longer stored in strings and loaded using a bunch of function calls, which means the game no longer has to *parse 29 or 30 strings* (!) whenever it loads a room. Instead they are stored in a constant int array, which gets loaded into the returned vector. This refactor includes the area lookup tilemap and menu background tilemaps as well.

Doing this fixes a problem where if you try to compile the game with `-O2`, the compiler could choke on the massive amount of strings in a level file, especially `Otherlevel.cpp`, and then die. Another solution would've been to move the level tilemaps out of the source files, but that raises the question of where they should be stored (maybe `data.zip`?), which entails having to read and parse them, and that's just too much unnecessary work imo.

I would've liked to not have to copy-paste the line `result.insert(result.end(), contents, contents+1160);` everywhere (or have braces with each case), and have just one of that line at the end, but that would've meant having to initialize an array with multiple different possible values, which is simply hard to do. Due to lexical scoping, we can't declare an array in the switch-case and then use it outside of the switch-case. However, if we declare an array before the switch-case, we can't re-initialize it afterwards. The language simply won't let us. If we used an `std::vector<int>` instead, we could do that. But that only works in C++11 and later.

But this copy-pasting doesn't cause any issues. If I forgot to paste one, `-Wunused-variable` would've caught it, but that warning doesn't pop up, so I'm confident I haven't forgotten any.

Anyway, after compiling this patch with `-O2`, GCC makes it through every level file much, much quicker, and especially makes it through `Otherlevel.cpp` at least ten or hundred times faster than it did before. However, it still chokes on `Scripts.cpp`, and at one point ended up using two gigs of my RAM to compile that file alone, and then also spat out this note:

Scripts.cpp is too large

But it still compiled fine. So it still looks like we have some work to do there.

## Legal Stuff:

  for these changes


TerryCavanagh:
Oh, whoa!

flibitijibibo:
This and Script.cpp are the two things that make this game the only thing I have that compiles slower with Clang than GCC... really glad this is fixed. Script will probably be a lot more work to fix, if I'm remembering that file correctly.

PR #243, opened 2020-05-18
Refactor main game scripts to not be stored in function parameters

InfoTeddy:
It's like #242 except for strings. Again, I would've not liked to copy-paste `filllines(lines);` everywhere, but due to lexical scoping and how array initialization works (and the fact that it's `const`, which wasn't essential in #242 but I think has much more of an impact in this patch because then the compiler doesn't have to account for being able to change the length of a string), I have to. I at least put it in a macro this time because the full expansion is much more of an eyesore with the `sizeof()` stuff, however it has to be a macro and not, say, an inline function, because `sizeof()` doesn't work with array parameters.

The basic point here is this: it seems like whenever the compiler encounters a constant string in a function argument, it has to go out of its way to find a dedicated location to store that string, which is inefficient. Every single time. And this apparently gets worse the higher amount of optimization you compile the game with.

What this patch does (along with some other stuff) is to replace all the `add(...);`s in `Scripts.cpp` and `TerminalScripts.cpp` with a `const char* lines[] = { ... };`. Putting all the strings in a nice and neat package like this is super helpful for the compiler, which now knows exactly how much space it will need, and so it doesn't have to keep being caught off-guard by every single function call that has a constant string.

After doing this, compiling `Scripts.cpp` (and to a lesser extent, `TerminalScripts.cpp`) is now ten or a hundred times faster than before, especially using `-O2`. And with this patch, compiling the game with `-O2` is no longer a huge PITA!

## Legal Stuff:

  for these changes


flibitijibibo:
Will try to check this out later in the week - FNA3D has most of my attention at the moment, so if someone else wants to check this out sooner this is open for review.

flibitijibibo:
For some reason I thought this patch was going to be more than what it really was... seems fine!

PR #245, opened 2020-05-18
Fix <hardestroom> having atoi() called on it when loading quicksave

InfoTeddy:
For some reason (probably a copy-paste error), this XML tag gets `atoi()` called on it before being assigned to `Game::hardestroom`. And only when loading a quicksave, at that.

This would result in `Game::hardestroom` being set to an empty string, which if you kept until Game Complete, would end up rendering as a single null byte (if you even have a font face for said null byte).

I'm not sure how this error compiles in the first place, but whatever.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -5118,7 +5118,7 @@ void Game::loadquick()
         }
         else if (pKey == "hardestroom")
         {
-            hardestroom = atoi(pText);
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Also on a side note, apparently passing invalid strings into `atoi()` can result in UB depending on which C library you're using. So maybe we should fix that, too...

PR #248, opened 2020-05-19
Remove 7 extra zeroes on the 21st row of Philadelphia Experiment

InfoTeddy:
This PR fixes... this issue.

Philadelphia Experiment misaligned
<sup>OH CHRIST OUR NUKE MISALIGNED THIS ROOM</sup>

It was caused by #242. I didn't think #242 would cause any problems, but apparently it does!

All I did in that PR was remove all `tmap.push_back("`s and replace all the `");`s with `,`s. I didn't add any new tiles or anything. So what's going on?

Well my first thought was to evenly space out each number in the source file, just to make the tilemap look prettier. Because right now it looks like this:

No spacing
<sup>Meh.</sup>

If I could clearly see each tile spaced out evenly, as it should be in the actual game, then maybe I can see what went wrong. And what do we get when we evenly space out the tiles?

With spacing, we see extra zeroes
<sup>Oh.</sup>

So as you can see, there are 7 extra zeroes on the 21st row of this room.

This didn't cause any issues on the old `push_back(string)` system, because when the game parsed the string it only did it 40 times and didn't ever go over the entire string, but it causes a huge alignment issue now. Don't know why these 7 zeroes are here, but I'm removing them.

## Legal Stuff:

  for these changes


flibitijibibo:
ASCII art is the best level format

PR #255, opened 2020-05-22
Remove unnecessary middleman ed.swapmap

InfoTeddy:
When the game loads a room in a custom level, previously it would load the tilemap of that room into `ed.swapmap`, and then `mapclass::loadlevel()` would manually go through each element in `ed.swapmap` to set each tile in `contents`. Why do that, when you can just return the vector from `editorclass::loadlevel()` and set it directly? `ed.swapmap` is really unnecessary.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -1568,17 +1568,11 @@ void mapclass::loadlevel(int rx, int ry)
 			roomname=ed.level[curlevel].roomname;
 		}
 		extrarow = 1;
-		ed.loadlevel(rx, ry);
+		contents = ed.loadlevel(rx, ry);
 
 
 		roomtexton = false;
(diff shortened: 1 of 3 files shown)


flibitijibibo:
It’d be useful to prevent allocations, but it doesn’t really do that... the only thing I see in the patch is a possible optimization to size the vector at 30*40 on startup so it doesn’t keep resizing as it loads.

InfoTeddy:
I'm just doing this to be a defensive coder. There's a bug that was inadvertently caused from a change I made, due to this design, and it's only luck that that bug also got inadvertently fixed:

A bug that was inadvertently caused and then inadvertently fixed

flibitijibibo:
Very well - I suppose we’ll get to this eventually as part of the Great Devectoring(TM).

PR #257, opened 2020-05-27
Fix vector indexing out-of-bounds if say()/reply()/text() asks for more lines than there are

InfoTeddy:
This PR adds bounds checks to those commands in case `say()`/`reply()` asks for more lines than there are left in the all-script-lines buffer (not just the current script, so in order for it to segfault your script has to be last in the all-script-lines vector), and in case `text()` asks for more lines than there are in the rest of the rest of the parsed internal script.

Also some small refactors.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -376,7 +376,10 @@ void scriptclass::run()
 				for (int i = 0; i < ss_toi(words[4]); i++)
 				{
 					position++;
-					txt.push_back(commands[position]);
+					if (position < (int) commands.size())
+					{
+						txt.push_back(commands[position]);
+					}
 				}
 			}
 			else if (words[0] == "position")
@@ -3794,7 +3797,10 @@ void scriptclass::loadcustom(std::string t)
 				int ti=atoi(words[1].c_str());
 				int nti = ti>=0 && ti<=50 ? ti : 1;
 				for(int ti2=0; ti2<nti; ti2++){
-					i++; add(customscript[i]);
+					i++;
+					if(i < (int) customscript.size()){
+						add(customscript[i]);
+					}
 				}
 
 				switch(speakermode){
@@ -3816,7 +3822,10 @@ void scriptclass::loadcustom(std::string t)
 				int ti=atoi(words[1].c_str());
 				int nti = ti>=0 && ti<=50 ? ti : 1;
 				for(int ti2=0; ti2<nti; ti2++){
-					i++; add(customscript[i]);
+					i++;
+					if(i < (int) customscript.size()){
+						add(customscript[i]);
+					}
 				}
 				add("position(player,above)");
 				add("speak_active");
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -3792,11 +3792,8 @@ void scriptclass::loadcustom(std::string t)
 					break;
 				}
 				int ti=atoi(words[1].c_str());
-				if(ti>=0 && ti<=50){
-					for(int ti2=0; ti2<ti; ti2++){
-						i++; add(customscript[i]);
-					}
-				}else{
+				int nti = ti>=0 && ti<=50 ? ti : 1;
+				for(int ti2=0; ti2<nti; ti2++){
 					i++; add(customscript[i]);
 				}
 
@@ -3817,11 +3814,8 @@ void scriptclass::loadcustom(std::string t)
 				add("text(cyan,0,0,"+words[1]+")");
 
 				int ti=atoi(words[1].c_str());
-				if(ti>=0 && ti<=50){
-					for(int ti2=0; ti2<ti; ti2++){
-						i++; add(customscript[i]);
-					}
-				}else{
+				int nti = ti>=0 && ti<=50 ? ti : 1;
+				for(int ti2=0; ti2<nti; ti2++){
 					i++; add(customscript[i]);
 				}
 				add("position(player,above)");
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -3557,12 +3557,12 @@ void scriptclass::loadcustom(std::string t)
 	for(size_t i=0; i<customscript.size(); i++){
 		if(scriptstart==-1){
 			//Find start of the script
-			if(script.customscript[i]==cscriptname+":"){
+			if(customscript[i]==cscriptname+":"){
 				scriptstart=i+1;
 			}
 		}else if(scriptend==-1){
 			//Find the end
-			tstring=script.customscript[i];
+			tstring=customscript[i];
 			if (tstring.size() > 0) {
 				tstring=tstring[tstring.size()-1];
 			} else {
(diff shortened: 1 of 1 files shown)


TerryCavanagh:
Does this impact the internal scripting exploit? 🤔

InfoTeddy:
No, it doesn't.

And because no one's bothered to ever properly document how and exactly *why* the internal scripting exploits work, I'll explain it now:

There are two internal scripting exploits, the first one giving you cutscene bars and the second one not giving you cutscene bars.

### Cutscene bars (`say(-1)`)

The one that gives you cutscene bars uses `say(-1)`, which is an interesting statement. Basically the custom script parser will only follow the number given to `say()` if it's between 0 and 50, inclusive. But if the number given fails that check, it will instead default to adding 1 line. Since we tell it -1 lines, it will add 1 line.

But when this gets converted to `text(gray,0,114,-1)` by the parser, the `text` command will actually follow our instruction to take in the next -1 lines. This essentially results in it taking in 0 lines, resulting in this parsed script line doing nothing.

Using that trick, we can smuggle in one internal command through `say(-1)`. Basically, that's all that the internal scripting exploits do - they smuggle in internal commands through `say()`, otherwise they'll be thrown away and ignored by the simplified-to-internal script parser. (And on a side note, instead of using `say(-1)` we could use `reply()` or any other negative number, but it has to be negative.)

The internal command to smuggle in here is `text()`, because since we're using `say()` (and we *have* to use `say()`, or `reply()`) we get a bunch of parser output lines that we really don't want, like `customposition(center)` and `speak_active`. But let's think ahead, because we'll need another `say(`*n*`)` block to smuggle in *n* lines of internal scripting after this one, and *that* will generate `squeak(terminal)` and `text(gray,0,114,`*n*`)`. We want exactly neither of these, especially not the `text()` since that will put all of our internal scripting into a text box instead of leaving them be.

Anyways, so the full line to smuggle in using `say(-1)` is `text(1,0,0,4)`, because if you look at the above, we've generated 4 lines we don't want (`customposition(center)`, `speak_active`, `squeak(terminal)`, `text(gray,0,114,`*n*`)`), so we tell the script executor that the next 4 lines are a text box, and then it won't execute them. Simple as that. (The "1" in `text(1,0,0,4)` is just a convention that @FredrIQ, who I think was the original discoverer of internal scripting, adopted. Personally I use `text,,,,4` before Ved's automatic Internal Scripting Mode was introduced.)

And on a side note, if you wanted to reduce script parser output lines, you can remove that `squeak(terminal)` line from being generated by doing `squeak(off)` at the top of the script. Of course, you'll then have to change `text(1,0,0,4)` to `text(1,0,0,3)` since only 3 lines are generated after that.

So we're limited to `say(50)`, because otherwise the simplified parser will only take in one line (and this was apparently limited to `say(5)` in 2.0), but the last line will *have* to be `text(1,0,0,4)` in order to make the script executor ignore any of the extra lines that our `say()` block generated. So in total, we get 49 lines of scripting each time.

However, this will have to be adjusted if we use `text()` in one of our internal scripting blocks, because since we use `text()` as well to overwrite lines we don't want, we end up destroying any previous text boxes that have not been presented to the screen with `speak` or `speak_active`. Thus, if you use `text()` in internal scripting, the `speak` or `speak_active` *has* to follow after within the same block, and *cannot* be in the next block.

We do need to stop the script sometime soon. Apparently the convention is to use `loadscript(stop)` (and this has to be an internal line, so we still get a maximum 49 lines of internal scripting for each block) because what this does is load an invalid script, which ends up having 0 lines, which means `scriptclass::run()` will set `running` to false.

This is ostensibly to avoid the last 2 generated lines of `customposition(center)`+`speak_active`, but I wonder if `text(1,0,0,2)` couldn't have been used instead? It's better that way, because you don't need to do an `endcutscene`+`untilbars` either.

Anyway, this is why the `say(-1)` method looks like this:

```
say(-1)
text(1,0,0,4)
say(5)
playef(11)
text(gray,-500,-500,1)
bruh
speak
text(1,0,0,4)
say(4)
endtext
endcutscene
untilbars
loadscript(stop)
```

### No cutscene bars (`iftrinkets(0,`*script*`)`)

This was apparently accidentally discovered by @FredrIQ by simple messing-around.

It looks like this: this is `script1`:

```
iftrinkets(0,script2)
```

And this is `script2`:

```
say(5)
playef(11)
text(gray,-500,-500,1)
bruh
speak
loadscript(stop)
text(1,0,0,4)
```

You run `script1`. Running `script2` won't work. This is weird, because we're doing the same `say()`-smuggling thing, yet apparently having that line at the end of `script2` and calling `script2` this way is necessary for it to work?

This exploit relies on a lot of coincidences:

- `scriptclass::loadcustom()` (in 2.2 and before, simply `scriptclass::load()`) aggressively calls `tokenize()` on every single line in the script. I mean, it has to do this in the first place to figure out if they're a valid simplified command, right?

  But actually, what this does is leave the last line of the simplified script in `words`, even if it's not a valid simplified command. When loading `script2` above, `words` will essentially be `text(1,0,0,4)`.

- At first glance, it looks like all internal commands in `scriptclass::run()` use a simple `else if`-chain, right? Wrong. What it actually does is *sometimes* use `if`, `if`, `if`, and other times it's an `if`, `else if`, `else if`.

  Basically, it's an `else if`-tree that's been broken in some parts. And this is where the exploit comes in:

  - `script1`'s `iftrinkets(0,script2)` gets parsed to `customiftrinkets(0,script2)`.
  - When the script executor (`scriptclass::run()`) runs the `customiftrinkets` internal command, it'll call `load()` (which in 2.3, calls `loadcustom()`, but either way it still does the custom script parser).
  - `load()` will have put `text(1,0,0,4)` into the `words` vector.
  - *And then*, due to the broken `if`-tree, the `text()` internal command is able to be executed *in the same frame* that the new script gets loaded, since `text()` is *not* in the same branch as `customiftrinkets()`!

Due to these coincidences, this lets us overwrite the first 4 lines of `cutscene()`+`untilbars()`+`squeak(terminal)`+`text(gray,0,114,`*n*`)` by using the `text(1,0,0,4)` at the end of the script.

----------

So, basically: the `text(1,0,0,4)` at the end of the script doesn't actually ask for any lines beyond the end of the script. That's not what's special about it, but it being on the last line of the script *is* special. The exploit is in the fact that it's the last line of the script, so it gets loaded into the `words` vector whenever `customiftrinkets()` in `script1` calls `load()`, and because `customiftrinkets()` is in its own branch separate from `text()` *and* `text()` comes later in the frame ordering than `customiftrinkets()` does, we get to start overwriting some lines, which enables internal scripting.

And that's why this PR doesn't impact internal scripting, it just fixes a segfault is all.

FredrIQ:
Internal scripting was generally known before I messed with VVVVVV scripting, but it was believed to be unreachable in 2.1. There's one level in 2.0 that uses a completely different method to get colored text boxes that no longer functioned starting from 2.1 by virtue of relying on say(6) not working as it should in 2.0, but he/she never took advantage of something like text(1,0,0,4) to sneak in additional internal scripting lines. say(-1) was discovered as part of me trying to find an entry point to get access to the internal scripting engine (similar concept as entry points in console hacking), the second method meanwhile was discovered by sheer luck.

PR #259, opened 2020-05-29
Statically allocate level arrays

mwpenny:
## Changes:

This prevents a potential stack overflow if the compiler tries to allocate memory for all of the arrays at once before initializing them (observed on MSVC). Performance should be better here too as the initialization won't be done each time the function is called. I cloned the repo, hit build, went to "start game" in the menu and it immediately crashed because of this. Everything works fine after the fix.

## Legal Stuff:

  for these changes


flibitijibibo:
Well, that was easy.

How much memory does this end up taking? Do we have a before/after comparison? I was thinking about making the arrays individual functions to keep it on the stack, if it took too much space.

mwpenny:
It uses ~104MB of RAM before and after this change running in debug on Win64. The EXE is about half as big (likely due to this change allowing the compiler to perform better optimizations)

flibitijibibo:
Whew, okay then - i’ll check this out on flibitBuild to see what these old compilers think and merge if nothing shows up!

InfoTeddy:
Seems to compile fine with `-std=c++03`.

Damn, I didn't think that prepending `static` everywhere would work!

PR #260, opened 2020-05-29
Statically allocate built-in scripts as well

InfoTeddy:
Although it's not an issue, this should minimize the stack footprint of calling `scriptclass::load()`, especially if it goes down to calling `scriptclass::loadcustom()` or `scriptclass::loadother()`.

## Legal Stuff:

  for these changes


flibitijibibo:
I feel like we're going to find a lot of these...

PR #262, opened 2020-05-29
Per-level custom asset loading functionality

Fussmatte:
## Changes:

With this, players can include a subfolder in their levels folder with the name of their level file (eg levelname.vvvvvv will have a folder named "levelname") and put assets into it, and it will automatically load when the level is loaded from either the editor or custom levels list. In addition, pressing F9 in the level editor will reload the level's resources, for easier level development.

## Legal Stuff:

  for these changes


Fussmatte:
Additional note I forgot: you can also put your assets directly next to your level file as a zip file, eg "levelname.data.zip".

Fussmatte:
Wait, wait, apparently there's some issues with the code, namely that it uses C++11 stuff I think, which I didn't realise. I'm going to have to fix this up a bit.

Fussmatte:
_Please_ do let me know if there's anything else in the code that needs to be fixed before this can be merged, if it will be merged. Thanks!

AllyTally:
I'd like to mention that this commit in no way breaks old levels, and won't prevent 2.3 levels from working in 2.2, if that was a concern. People that use 2.2 will just do what they have always done to load custom graphics, and people who use 2.3 can have life a little easier.

Also, some changes are being discussed about this PR on the VVVVVV Discord server, such as removing a lot of the code commented out by earlier commits.

InfoTeddy:
Btw changing that `1` to a `0` in that `PHYSFS_mount()` in `FILESYSTEM_init()` is apparently necessary, otherwise you can't load another custom level after having already loaded one. I'm not sure why, other than that @leo60228 did it in VCE in February.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -149,6 +151,40 @@ char *FILESYSTEM_getUserLevelDirectory()
 	return levelDir;
 }
 
+bool FILESYSTEM_directoryExists(const char *fname)
+{
+    return PHYSFS_exists(fname);
+}
+
+const char pathSeparator =
```
flibitijibibo, line 159: This can be replaced with PHYSFS_getDirSeparator().

desktop_version/src/editor.cpp
```diff
@@ -1640,6 +1645,38 @@ void editorclass::load(std::string& _path)
...
+    if (cstr && FILESYSTEM_directoryExists(zippath.c_str())) {
+        printf("Custom asset directory exists at %s\n",zippath.c_str());
+        FILESYSTEM_mount(zippath.c_str());
+        graphics.reloadresources();
+        music.init();
+    } else if (zip_path != "data.zip" && !endsWith(zip_path, "/data.zip") && endsWith(zip_path, ".zip")) {
+        printf("Custom asset directory is .zip at %s\n", zip_path.c_str());
+        PHYSFS_File* zip = PHYSFS_openRead(zip_path.c_str());
+        zip_path += ".data.zip";
+        if (zip == nullptr) {
```
flibitijibibo, line 1664: Should be NULL for old compilers

desktop_version/src/FileSystemUtils.cpp
Fussmatte: Do you mean the entire FILESYSTEM_directoryExists() function? I'm a little confused here.

desktop_version/src/editor.cpp
Fussmatte: Sounds good to me.

desktop_version/src/FileSystemUtils.cpp
InfoTeddy: No, you just use `PHYSFS_getDirSeparator()` instead of having to make your own `const char pathSeparator`.

PR #265, opened 2020-06-01
Allow for conditional building of Steam and GOG APIs

InfoTeddy:
I think it's a bit silly to always include the Steam and GOG APIs whenever we build VVVVVV, since the only time they'll ever be used is in a live build and not a dev build.

So now Steam and GOG are disabled by default. If you want them, you'll need to add `-DSTEAM=ON` or `-DGOG=ON` respectively at CMake time.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -10,6 +10,14 @@ OPTION(ENABLE_WERROR "Treat compilation warnings as errors" OFF)
 SET(CUSTOM_LEVEL_SUPPORT ENABLED CACHE STRING "Optionally disable playing and/or editing of custom levels")
 SET_PROPERTY(CACHE CUSTOM_LEVEL_SUPPORT PROPERTY STRINGS ENABLED NO_EDITOR DISABLED)
 
+SET(STEAM OFF CACHE BOOL "Use the Steam API")
+SET(GOG OFF CACHE BOOL "Use the GOG API")
+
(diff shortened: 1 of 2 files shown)


InfoTeddy:
```
D:\a\VVVVVV\VVVVVV\desktop_version\src\Network.c(41,44): error C2466: cannot allocate an array of constant size 0 [D:\a\VVVVVV\VVVVVV\desktop_version\VVVVVV.vcxproj]
D:\a\VVVVVV\VVVVVV\desktop_version\src\Network.c(41,45): error C2133: 'backends': unknown size [D:\a\VVVVVV\VVVVVV\desktop_version\VVVVVV.vcxproj]
```
MSVC, why must you be such a PITA...?

InfoTeddy:
Ok, so I fixed it. `backends` is no longer created if `NUM_BACKENDS` is 0, however this results in me having to macro-guard all references to `backends`, and then I have to make an `UNUSED` macro to suppress compiler warnings about unused variables if there are 0 backends. But hopefully this is fine.

flibitijibibo:
Would it make more sense for these to default to `ON` when the build type is set to Release or RelWithDebInfo? Disabling for debug makes sense to me but adding build steps (and 2^2 possible configs) for something that _technically_ doesn't pull in any added dependencies seems like overkill.

InfoTeddy:
Hm, okay, that makes sense. Is there an elegant way to check for either Release or RelWithDebInfo in CMake?

InfoTeddy:
And what about MinSizeRel?

InfoTeddy:
Ok, so these APIs are now both included in release builds.

flibitijibibo:
Looks good, thanks!

PR #188, opened 2020-04-03
Remove the passing around of 'dwgfx', 'map', 'obj', 'music', 'key', and 'help'

InfoTeddy:
### Summary

This PR is a re-do of #178, except this time everything is not in one commit. It has been split up into 62 commits. Also it contains more cleanups that I had missed before.

`dwgfx`, `map`, `obj`, `music`, `key`, and `help` are no longer passed around as parameters in any function that needs them. Additionally, all `dwgfx` has been renamed to `graphics` since it's a clearer name (I'm not sure what the "dw" in "dwgraphics" stands for, and it sounds like a remnant from the Flash days).

Additionally, since this change affects pretty much the entire codebase, I also took it upon myself to clean up mixed indentation, remove outdated comments, remove useless or unused functions, remove unused parameters, remove unused stuff like testing code and input recording code, and remove the `temp` global file-level variable and move `tempstring` off of class objects. Also, all expressions of the form `UtilityClass::`*`something`* have been changed to `help.`*`something`*.

The standard of indentation I want is to at least have files be consistent with their indentation. A file can indent with tabs or spaces, but it ought to be at least consistent. The annoying part is when you suddenly start switching indentation in the middle of a function. The exception here is `Graphics.cpp`, which switches to tabs starting from `Graphics::setcol()`, which I thought was fine because it wasn't in the middle of a function, and it consistently uses tabs below `Graphics::setcol()` and spaces above, so it's fine.

The following functions had unused parameters:
 - `entityclass::getcompanion()`
 - `Graphics::Hitest()`
 - `Graphics::drawtile()`
 - `Graphics::drawtile2()`
 - `musicclass::playef()`

The following functions or variables were unused or useless:
 - `entityclass::checkdirectional()`
 - `Game::telegotoship()`
 - `Game::sfpsmode`
 - `Game::telecookieexists`
 - `Game::quickcookieexists`
 - `stage`
 - `swfStage`
 - `class Stage`
 - `updategraphicsmode()`
 - `Screen::ClearScreen()`
 - `musicclass::initefchannels()`
 - `Screen::glScreen`
 - `editorclass::weirdloadthing()` (instead, make a copy of the string and use `editorclass::load()`)
 - `musicclass::processmusicfade()`

## Legal Stuff:

By submitting this pull request, I confirm that...

- [X] My changes may be used in a future commercial release of VVVVVV (for
  example, a 2.3 update on Steam for Windows/macOS/Linux)
[...]

PR #272, opened 2020-06-04
Add localization system

Dav999-v:
## Changes:

This PR, when finished, adds a localization system to VVVVVV.

The most important design goals I hope to achieve are the following:
- The system should be easy enough for everyone involved to understand: ideally it should not get in the way of code contributors knowing how to add features that have text to the game, code contributors and maintainers should preferably not have to read too many instructions or follow too complex/repetitive/error-prone procedures on how to maintain full translatability.
- It should not be hard for translators to avoid accidentally making text run offscreen, out of boxes, etc; there should be enough room for translations where possible; translators should not have to manually wordwrap their strings (except CJK because it's hard to automate there). Strings should come with defined limits where possible.
- Context and meaning of text should be clear as much as possible; in particular, roomnames should be translated while seeing the room in action.
- If all language resource files (including any 'English' files) are missing, the game should still be fully functional in English.
- People should be able to make fully-functional translations by just making resource files for it, without needing to recompile the game (if that language doesn't require technical changes, such as RTL text, or otherwise special unsupported font handling)

The idea is more or less inspired by the gettext approach, by surrounding hardcoded English text by a function `loc::gettext()`. If the language is not set to English, this function would look up the string in the list of translations and return the matching translation if found. The language files are XML files that look somewhat like this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<strings>
	<string english="Game paused" translation="Spel gepauzeerd"/>
	<string english="Aa" translation="Bb" explanation="appears somewhere" max="30"/>
</strings>
```
Cutscenes, roomnames, etc have their own more specialized XML files and functions. There is also an Excel spreadsheet with macros that can be used as an "editor" for the XML files, as an alternative for translators who prefer working in a spreadsheet instead of with XML.

One of the problems that also needed to be solved in VVVVVV is text wrapping, both in menus and in cutscenes, because manually inserting newlines would require a lot of effort to get right and would be extremely error-prone. So, an automatic text wrapping function has been added for the interface and cutscenes, and while strings were extracted, manually wrapped strings have been changed to use the automatic wrapping function instead. For cutscenes, I have even been able to leave the cutscenes themselves almost entirely unchanged - the `text` command unwraps its lines, looks up the translation for that dialogue in that cutscene, and re-wraps it. (This possibly results in more or less lines in the translation, but that's not an issue because `position` comes after `text`, not before.)

I have considered a key-based system too (`GAME_PAUSED`), but that also has disadvantages, for example, it becomes less obvious off the bat how to add new features that have text, all keys would have to probably appear at least twice or thrice in the code, the source becomes less self-documenting and it'd be harder to gracefully fall back to English if something wasn't translated or if language files (even the English one) are incomplete somehow.

These are the language files that can be found for each language:

- **meta.xml** - some language-specific configuration
- **strings.xml** - the main list of strings, accessed in the code using `loc::gettext(const char* eng)`
- **numbers.xml** - the list of numbers from Zero to One Hundred, as well as plural form configuration
- **strings_plural.xml** - similar to strings.xml, except strings can have different forms depending on number (1 trinket, 2 trinkets). This can be adapted to how plurals work in the language. Accessed in the code using `loc::gettext_plural(const char* eng_plural, const char* eng_singular, int count)` or the convenience function `gettext_plural_fill(char* buf, size_t buf_len, const char* eng_plural, const char* eng_singular, const char* args_index, ...)` (using VFormat)
- **cutscenes.xml** - the dialogue texts for all cutscenes, and per-dialogue formatting attributes (centering, padding, ...)
- **roomnames.xml** - all regular room names
- **roomnames_special.xml** - special cases for room names, such as Prize for the Reckless, glitch/change, and hidden names/generic area names

There is also a translator menu (only shown in specific cases), which provides options like testing menus, translating room names within the game, syncing all language files with the English template files, and other things.

Lastly, there's a **README.txt** file for translators and maintainers in the languages directory, which explains a lot more than this PR description does (and which can be updated after this is merged).

This PR does not yet aim to include support for bigger fonts as needed for Chinese, Japanese and Korean (and different fonts for different languages in the first place). My plan is to first have this PR merged (that is, squashed; and it's already massive enough as-is!) and then to make a new PR that adds support for different and bigger fonts. I already have that mostly thought out.

After this PR is merged, officially distributed builds (Steam, Itch, Make and Play, etc) should bundle the corresponding versions of the `lang` and `fonts` folders (which are in `/desktop_version/`) as folders next to `data.zip`, and have a folder/text file with C-HashMap's license as discussed below in the comments. Instructions for compiling VVVVVV yourself should not have to change - people should not download these separately like they have to do with `data.zip`, everything will just work in most cases. The language and font files are in the repo, and subject to constant change along with the code, so if they're not found next to `data.zip`, the game will look for them in `/desktop_version/` automatically, or the user can pass `-langdir`. (And if language files are not found for some reason, the game will remain fully functional in English and use the traditional font from `data.zip`.) Better explained by f4b423032bac6f1251107ca12f8a3183601e5959:

> So the better idea is this: VVVVVV should be distributed with the lang folder in the same place where data.zip is (so for example, in the user's Steam folder), and the game will use that if it exists.
> 
> If it doesn't, that probably means someone has compiled VVVVVV and created a build folder for themselves. In that case, the game checks if /desktop_version/ (or \desktop_version\ on Windows) is somewhere up in the path, and if it is (which is likely, since the build folder was probably made inside of desktop_version/), the game will try to use desktop_version/lang/. Which is convenient, because that's the up-to-date version that belongs to the code you compiled, and it's under version control!

## Legal Stuff:

  for these changes


Dav999-v:
I think normal PRs send notifications and draft PRs don't, so... @flibitijibibo, could you give some feedback on this?

InfoTeddy:
> I think normal PRs send notifications and draft PRs don't

I still got emails from this PR.

flibitijibibo:
Seems fine so far. Word wrapping will get really interesting once CJK gets involved but that’s a looong ways off.

Dav999-v:
Alright, thanks. As for CJK, now that I think of it, I read that at least Japanese game developers tend to manually wrap all text because it's hard to write an algorithm that knows where to split sentences and where not to; so maybe CJK languages should just have newlines inserted manually (also, I've been thinking the translation files could have hints that indicate how much room there is)

desktop_version/src/Render.cpp
```diff
@@ -338,6 +343,16 @@ void menurender()
         }
 
 
+        break;
+    case Menu::language:
+        if (loc::languagelist.size() == 0)
```
InfoTeddy, line 348: Can be changed to `loc::languagelist.empty()`. Doesn't really make a difference functionally, but it makes the code a bit clearer.

desktop_version/src/Game.cpp
```diff
@@ -7134,6 +7136,28 @@ void Game::createmenu( enum Menu::MenuName t, bool samemenu/*= false*/ )
         menuxoff = -40;
         menuyoff = 10;
         break;
+    case Menu::language:
+        if (loc::languagelist.size() == 0)
```
InfoTeddy, line 7149: (Same here, can be changed to `loc::languagelist.empty()`).

Jason-serkill:
That’s true
The code is easier to read

desktop_version/src/Render.cpp
Dav999-v: Changed in 57608cdffdf1656cd93c73370784c66a502aed43.

desktop_version/src/Game.cpp
Dav999-v: Changed in 57608cdffdf1656cd93c73370784c66a502aed43.

desktop_version/src/Input.cpp
```diff
@@ -587,12 +587,55 @@ void menuactionpress()
...
+            map.nexttowercolour();
+            break;
+        case 1:
+            // translation statistics
+            // TODO
+            map.nexttowercolour();
+            break;
+        case 2:
+            // return. Also leave the language selection menu
+            game.returnmenu(); game.returnmenu();
```
InfoTeddy, line 625: You can do `game.returntomenu(Menu::options);` instead, if that fits your intent better.

Dav999-v: Sounds good, done in cb2986d7f3c1241be91bd0f81ee65e4b254270d0.

InfoTeddy:
I'm a bit concerned that the `loc::gettext()` system seems to be putting more and more string literals in function arguments. For an example of how much a problem this could be, see the saga of #242, #243, #259, #260, and #261.

The main point is this: whenever the compiler encounters a string literal in a function argument, it has to go completely out of its way to find a location to put that string in. This is wildly inefficient, and only gets worse as you increase the optimization level. Before #242, I couldn't even compile the game with `-O2` because the compiler would keep hanging and dying on `Otherlevel.cpp`, and before #243 it would also hang for an extremely long time on `Scripts.cpp` at `-O2`.

Dav999-v:
> I'm a bit concerned that the `loc::gettext()` system seems to be putting more and more string literals in function arguments. For an example of how much a problem this could be, see the saga of #242, #243, #259, #260, and #261.
> 
> The main point is this: whenever the compiler encounters a string literal in a function argument, it has to go completely out of its way to find a location to put that string in. This is wildly inefficient, and only gets worse as you increase the optimization level. Before #242, I couldn't even compile the game with `-O2` because the compiler would keep hanging and dying on `Otherlevel.cpp`, and before #243 it would also hang for an extremely long time on `Scripts.cpp` at `-O2`.
> 
> Maybe this is only a problem if you have a huge amount of strings, I don't know.

Hm, most of the time I only wrap string literals that are already in function arguments (for example in calls to `graphics.Print`) by a function, right? So, the compiler doesn't suddenly have a lot more string literals to encounter than it did before. In fact, I'm getting rid of string literals wherever I convert manually wrapped text to automatically wrapped text.

InfoTeddy:
Well, I hope it won't be an issue, then.

InfoTeddy:
I've marked this as required for 2.4.

For anyone who's *not* on the VVVVVV Discord server, here's what @Dav999-v had to say regarding the state of the PR himself:

State of localization PR

Dav999-v:
Update, since people may be wondering: I expect to have Ved 1.9.0 done in a couple of days, after that I'll get back to working on this!

Dav999-v:
@flibitijibibo So, I decided that I wanted an alternative to `std::map` for storing all the translations (mainly because storing `std::string`s internally started to cause some annoyances, it seemed less and less convoluted to do it in a non-STL way, plus it'd likely eventually be replaced anyway). So I decided to give this library a go: https://github.com/Mashpoe/c-hashmap

It's a fairly lightweight library, and I'm pretty happy with how it works thus far. The only thing I'm not sure about now is whether it's fine to start using a BSD-licensed library in VVVVVV, since its license must apparently be reproduced in some kind of material provided with the binary distribution. So is that fine? On the one hand, the four other bundled libraries don't have that requirement, on the other hand that might just be a coincidence and not a deliberate choice per se, and all I think it really needs is a text file next to the executable. And maybe it'd be a good thing anyway to have a file like "this is VVVVVV's own license, this is what you can do with its assets, it uses these libraries, these are their licenses..."

flibitijibibo:
We can add the license to the distributed package if needed, `licenses` folders are pretty easy. I'm usually pretty skeptical of dictionary performance so if it doesn't show up in a `perf` graph that's the thing that worries me more. Maybe long after we've finished all the big interface changes and backports we can migrate to generated arrays or something, who knows.

Dav999-v:
Well, I did do a quick benchmark of it myself, and it was already significantly faster than `std::map`.

Dav999-v:
The CentOS CI CMake is too old to set `C_STANDARD` to C99 instead of C90 for C-HashMap. I don't know where it's set to C90, but it'd seem fine to me to set it to C99 (or upgrade the CMake somehow), since this was @InfoTeddy's argumentation for setting it to C90 in the first place:
> All the C third-party dependencies are C90, and all the C files we have are also C90 (well, almost, but that's easily sorted). So we have basically no reason to not go with C90 here.

InfoTeddy:
We are probably going to end up modifying the hashmap anyway to swap out the libc functions with SDL's stdlib. In that case, we can just modify it to not be C99. Looking at it, it's small enough that it probably doesn't deserve the overhead of being a true third-party library in the first place.

Dav999-v:
Even in that case, it's annoying to have the CI fail in the meantime. Is there any reason to restrict ourselves to C90 instead of going with C99, by the way? (Other than the fact that libraries were already written to be compatible with it)

InfoTeddy:
Visual Studio 2010, as @flibitijibibo can confirm. So more technically we are actually going with C90-but-with-some-C99-extensions - in particular, the extensions being `stdint.h` and `long long`. Anything else is out.

If you're really pressed for dev time, then you can just temporarily up the standard in your branch and we'll deal with it later when it comes time to merge this whole thing.

Dav999-v:
Turns out it's just one tiny change (moving the type out of a `for` loop) to make it compatible with C90-with-extensions, so I'll do that instead.

AllyTally:
What does "Level Complete!" and "Game Complete!" being translatable mean for custom levels?

Dav999-v:
> What does "Level Complete!" and "Game Complete!" being translatable mean for custom levels? (or, well, translatable assets in general, I guess)

If a custom levels provides these images as assets, they'll still override the translation. None of the assets can be translated/replaced, these textboxes are simply printed as text, but English will still use the images.

Dav999-v:
> > (this is also why I do not like the direction of removing default function arguments and introducing hundreds of hardcoded `10, 304` all over the place)
> 
> Well, if you want instead, they can be replaced with `-1, -1` to indicate "default" or similar, or we can add a version of the function (with a different name) that fills in the default arguments automatically. I'm not sure that you need default function arguments here.

I'm only seeing this message now, a day later, but apparently we both had the exact same idea: I pushed 366d17d7ebdb005e029bfad97d6bcd7b98fe9cb1 in the same minute you posted that, heh.

It's not that I think we necessarily need default function arguments here or anywhere, it's just that I think getting rid of them requires a bit more effort to determine for what reason they were used, and what the best alternative is on a case-by-case basis. And in this case, I could live with `-1, -1` being repeated a lot better already. To be fair, upstream currently only has 2 calls to `PrintWrap` so the fact that _it_ repeats the same `10, 304` isn't a problem yet, but it kind of says that this decision was made in advance of a mass-switchover to remove default arguments. And the localization branch will, as it stands, add 193 more `PrintWrap` calls, of which only seven need to modify both arguments, and then there's these copy-pasted lines that need to change the linespacing:

```
Render.cpp:                graphics.PrintWrap(0, 135, loc::gettext("Return to main menu?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 142, loc::gettext("Do you want to quit? You will lose any unsaved progress."), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 80, loc::gettext("Return to main menu?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 76, loc::gettext("Do you want to quit? You will lose any unsaved progress."), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 135, loc::gettext("Return to main menu?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 142, loc::gettext("Do you want to quit? You will lose any unsaved progress."), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 80, loc::gettext("Return to main menu?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:                graphics.PrintWrap(0, 76, loc::gettext("Do you want to quit? You will lose any unsaved progress."), 196, 196, 255 - help.glow, true, 12);
Render.cpp:            graphics.PrintWrap(0, 88, loc::gettext("Do you want to return to the secret laboratory?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:            graphics.PrintWrap(0, 76, loc::gettext("Do you want to return to the secret laboratory?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:            graphics.PrintWrap(0, 88, loc::gettext("Do you want to return to the secret laboratory?"), 196, 196, 255 - help.glow, true, 12);
Render.cpp:            graphics.PrintWrap(0, 76, loc::gettext("Do you want to return to the secret laboratory?"), 196, 196, 255 - help.glow, true, 12);
```


Relatedly: I can already embrace the move to C a lot more than I used to - I've been happily writing mostly C-style code for the past three months, and I "get" and sometimes share the dislike of STL/C++ features. So in principle I could live with an eventual full migration, as long as it's thought out carefully and doesn't hurt the code quality/readability much. The main thing I'm now worried about though, is that C90 (in strict ANSI mode/without GNU extensions?) doesn't allow variable declarations anywhere else than at the top of a scope (I didn't even know that until I moved `Textbook` to its own C file). I hate that a lot, and I think it's harmful to code readability: at the top of a scope there's a bunch of mysterious variables that you don't understand yet (they may belong to very small steps right before the final `return`), along with statements that are in an unnatural order just so the one that assigns a variable comes before the others, then throughout the scope, every first use of a variable looks like a variable re-use and the type is out of view - you might as well have used `auto` for every variable. To me, that ought to be a deal-breaker for switching to C90, even with the MSVC 2010 situation. So in my opinion, write C all you want, but use the C++ compiler until it can properly compile C99, and don't make your (new) code harder to read just so it's compatible with this rule.

InfoTeddy:
I had already planned on refactoring all print callsites (including `PrintWrap`) to be better, which would probably have avoided this debacle about default arguments, but you didn't want it because it'd cause too many conflicts. But it's good to know you're okay with doing `-1, -1` for now.

And in my opinion, declarations at the top of blocks don't hurt readability unless you have many variables in the scope. And if that happens, you should probably move things to sub-functions anyway. (See the huge functions that we have; the logic/input/render ones are the most egregious.) Worst-case scenario you can just create a new block.

Dav999-v:
> And in my opinion, declarations at the top of blocks don't hurt readability unless you have many variables in the scope. And if that happens, you should probably move things to sub-functions anyway. (See the huge functions that we have; the logic/input/render ones are the most egregious.) Worst-case scenario you can just create a new block.

I disagree - I found myself spending a disproportionate amount of time thinking of a solution in `textbook_store_raw()` that wouldn't completely break my train of thought and make the function quite "unchronological" to read, or add ugly bare blocks. I ended up deciding to move a few types up (where they were still nowhere near being used yet) and to add one bare block at the end, because [`size_t cursor` and `char* added_text`](https://github.com/Dav999-v/VVVVVV/blob/8c5f99fbac2cf177c70f111c67772834a6f5a269/desktop_version/src/Textbook.c#L80) make no sense at the top of the function. I'm not really as satisfied with the result as I was when I wrote it in C++ territory. It's an advantage to be able to see clearly where a variable is first used and what type it is (and many variables are only assigned once) and it really helps to make clear the train of thought and make the function easier to read. And I still see no reason to need to stick to this declarations-at-the-top rule, other than if you think Microsoft's compiler being in C mode is more "pure" than Microsoft's compiler being in C++ mode, or something like that. And as far as I remember, @flibitijibibo said that he uses MinGW for projects fully in C anyway, which he couldn't do for C++ because of some reason. So whichever way it goes - staying on C++ or eliminating it altogether - should allow us to declare variables anywhere we want, as you can in C99.

Anyway, time to be positive.

> I had already planned on refactoring all print callsites (including `PrintWrap`) to be better, which would probably have avoided this debacle about default arguments, but you didn't want it because it'd cause too many conflicts. But it's good to know you're okay with doing `-1, -1` for now.

I've actually been thinking about this, since I'm gonna need to add a way to tell any print function what font a certain text should use (does it belong to the interface or to the current level) and how to vertically align CJK fonts. So I might as well share what I've been thinking about.

In short, I'd get rid of some of the special print functions (`bigprint`, `bprint`, `bprintalpha`, `bigbprint`, `bigrprint` and `bigbrprint` if I haven't forgotten anything) as well as the optional `cen` argument, and add a bitflag variable to the remaining print functions to replace them. Here's my notes on this:

print_bitflag

So, a simple, left-aligned `Print` or `PrintWrap` would have the flags set to `0`. Instead of controlling text centering with an optional boolean, we'd use `PR_CEN` (or something similar) as a flag. `bigprint` with its scale argument is replaced by `PR_2X`/`PR_3X`/`PR_4X` flags (text shouldn't ever be scaled by non-integer numbers, and it's never done in the game). A border can be added with `PR_BOR` instead of calling `bprint` or `bigbprint`.

As for the font languages 3-63, each language would just be allocated their own number when being loaded, and there'd be some function that takes the language code as a string and returns that number so it can be used as a flag as well (`Print(font_flag("ja"), 0, 0, "日本語", ...)` or something like that)

So we can get rid of at least six of the print functions in the process, and all the `bool cen = false` and `int sc = 2` arguments, and it would solve the future font problem without adding more functions or arguments (or state-changing functions like in LÖVE).

InfoTeddy:
> I disagree - [...] because [`size_t cursor` and `char* added_text`](https://github.com/Dav999-v/VVVVVV/blob/8c5f99fbac2cf177c70f111c67772834a6f5a269/desktop_version/src/Textbook.c#L80) make no sense at the top of the function.

So I reviewed that function, and it looks pretty readable to me either way. I can't vouch for the correctness of the implementation, but it looks okay. Maybe it's just me, but if I saw `size_t cursor; char* added_text;` at the top, I'd naturally read through the whole thing first or at least keep those in mind, and then see those variables being used at the end, and it's just not a big deal if those variables have to be at the top (creating a new block also works too). I know this sounds a lot like "I'm used to it so it's okay", but really, I wouldn't worry about this for a function as small as `textbook_store_raw`.

> @flibitijibibo said that he uses MinGW for projects fully in C anyway, which he couldn't do for C++ because of some reason.

The "some reason" is that he uses Visual Studio 2010 to build the game, and using a later version is prone to breakage from Microsoft's "updates".

I'm not sure about MinGW because [FNA3D](https://github.com/FNA-XNA/FNA3D) still seems to be using VS 2010, but I could understand it if the reason was "I don't want to switch to MinGW if it's an existing project" (which would also apply to VVVVVV too) or something like it also being prone to breakage (though from a cursory glance through the FNA server, compiling with MinGW would apparently let us avoid depending on the C runtime, which might be worth it?).

> I've actually been thinking about this, since I'm gonna need to add a way to tell any print function what font a certain text should use (does it belong to the interface or to the current level) and how to vertically align CJK fonts. So I might as well share what I've been thinking about.

Makes sense to me, and it was the approach I was thinking of too. I was thinking of just adding the align, border, and scale flags, while you can go add the other flags in this branch. Additionally, I was also thinking of unifying the color triplets into one struct (maybe an `SDL_Color`), so they could be easily reusable (this would go for many more functions than just the print ones).

Dav999-v:
> Maybe it's just me, but if I saw `size_t cursor; char* added_text;` at the top, I'd naturally read through the whole thing first or at least keep those in mind, and then see those variables being used at the end, and it's just not a big deal if those variables have to be at the top (creating a new block also works too). I know this sounds a lot like "I'm used to it so it's okay", but really, I wouldn't worry about this for a function as small as `textbook_store_raw`.

Well, it certainly doesn't help with the readability, and in my view it's bad practice actively harmful for readability. Would you voluntarily do it when contributing to an established C99 project that already doesn't do it, and has no plans to downgrade their standard?

And even with how "commonplace" C is, I have often encountered facts like "C standards before C99 need `/* */` comments / don't have a boolean type / require you to write `int i; for (i =` instead of `for (int i =`". But I had never experienced nor heard about _this_ until recently. So I imagine this is going to be a surprise for many other contributors as well, even if they have plenty of C experience.

> > @flibitijibibo said that he uses MinGW for projects fully in C anyway, which he couldn't do for C++ because of some reason.
> 
> The "some reason" is that he uses Visual Studio 2010 to build the game, and using a later version is prone to breakage from Microsoft's "updates".

So, [this](https://github.com/TerryCavanagh/VVVVVV/issues/707#issuecomment-817228780) is the message I was thinking about:

> **2010 is what my Windows build setup uses for C++ builds - I've moved to MinGW for C builds but libstdc++ and its dependencies are a blocker**, and I'm pretty careful about updating Visual Studio to a version Microsoft still touches, since at least for my type of work Microsoft not touching it is a feature and not a bug.

So, we don't have to do it in C++ because C++ supports intermixed declarations and statements, and if there's ever a full move to C, then flibit would probably be able to switch to MinGW, which is not stuck on C90.

---

> Makes sense to me, and it was the approach I was thinking of too. I was thinking of just adding the align, border, and scale flags, while you can go add the other flags in this branch. Additionally, I was also thinking of unifying the color triplets into one struct (maybe an `SDL_Color`), so they could be easily reusable (this would go for many more functions than just the print ones).

Nice! The color thing makes sense to me too.

So, my plan for some time has been to split localization into multiple PRs to make it a little more manageable - so I'm planning to first finish this PR without CJK font support, and then add CJK support in a new PR. So after this gets merged, you/I/someone can overhaul all the print functions, and the CJK PR can then add more flags to that system.

The more I think about it, the more this flags system makes sense. For #859 for example, we could decide the best way involves adding some kind of flag `PR_DRAW_BUTTONS` or `PR_BUTTON1_USE`, currently such a thing would be impossible without adding yet another new print function.

InfoTeddy:
Alright, as this is marked ready for review now...

This PR is 15k lines! That's more than #188 (10k) and #220 (2k) *combined*. In addition, while #188 and #220 were more-or-less net zero lines, with maybe some additions, this PR almost entirely consists of additions (14k lines alone!). Although, I'm told most of the diff consists of placing strings into `loc::gettext`, but it's still a lot.

So it'll probably take a bit to review. In addition, I'd like to livestream me reviewing the entire thing on Twitch. I'll need to set aside large chunks of time in my schedule to do this, which for me means I need to finish up my college homework. No ETA yet on when that will happen, but optimistically, it will be sometime in the next couple weeks.

Finally, I'd like to congratulate Dav999 for putting in this much work. It took almost two years (and CJK is coming in a separate PR!), but I'm sure it will be worth it.

egormanga:
May I ask, where is the translation spreadsheet? Can't find it in the PR files.

Dav999-v:
> May I ask, where is the translation spreadsheet? Can't find it in the PR files.
> 
> My team is going to start translating this game into Russian, and we usually cooperate using Google Spreadsheets.
> 
> Upd. Oh, found it: https://github.com/Dav999-v/TranslationEditor!

@egormanga Yeah, that link is the correct one. As for your issue there (with LibreOffice/Google Docs), I believe the plan will be to give translators prefilled versions of the sheets instead of blank ones, which would mean most (if not all) of the need for macros on the translator side is eliminated.

As a warning though: there are already plans for official translations (which will include Russian as far as I know, just maybe not in the first batch of languages). Maybe planned official translators can be replaced on a case-by-case basis, but that's up to Terry, so you'd have to contact him first about that. It may be worth asking, especially if you know the game well and can provide quality translations, but I can't promise it'll be approved. Otherwise it could still be a fan translation that won't be packaged with the game.

InfoTeddy:
I would just like to say that I have in fact been streaming on Twitch. [You can view the VODs here](https://www.twitch.tv/InfoTeddy/videos). I've also been announcing each stream on [my Twitter](https://twitter.com/Info__Teddy) as well.

InfoTeddy:
Is this ready to be reviewed after the VFormat changes? The last time I reviewed this was [almost a month ago...](https://www.youtube.com/watch?v=fHJyTwn35Mk) :stuck_out_tongue:

Dav999-v:
Certainly! I've done all the VFormat changes I needed to make.

Dav999-v:
It took a bit of work, but I just completed squashing and rebasing the localization branch!

localization is 56 commits ahead

This branch was intermixed with 2.5 years of upstream history, had lots of design decisions that developed and were changed over time, and had lots and lots of merge commits. Merging it as-is would be extremely messy, it'd make the change history of upstream VVVVVV confusing and very misleading, and it'd make bisecting to find the cause of bugs a nightmare. Squashing it down to a single commit would probably overload Github, your browser or you whenever you try to scroll through it, and make it very hard to know why "the localization teracommit" broke something if it turns out it did. So instead, I broke it up into smaller commits with the following general principle:

- First all the language files and all completely new code files, in separate commits, because they're the largest part of the additions, don't depend on anything else, and don't cause compilation errors if not compiled
- Then the "foundation" of the localization system: the entire engine, all the code changes that everything else depends on to make text translatable, while still not touching all the hardcoded text everywhere (except where it was hard to separate them).
- Then all the text itself (when not in the foundation commit) is made translatable, i.e. `loc::gettext` calls are added to strings to look them up in the language files. These are mostly local code changes that depend on the foundation commit, but nothing depends on them.

There are two branches now:

- [`localization-orig`](https://github.com/Dav999-v/VVVVVV/tree/localization-orig), the completely original history, never force-pushed, and can be used to see the state of the localization branch at any time in the past, still has all the rationales for design decisions in commit messages, etc
- `localization`, the squashed history for merging upstream, where all changes are only added as soon as they are actually added in the main repository

With that, I think this is ready to be merged now!

InfoTeddy (approved):
Sure, looks good to me.

AllyTally:
It's been 3,000 years...

mothbeanie:
gratz!!!!!!!

PR #273, opened 2020-06-07
Fix music reloading

leo60228:
## Changes:

This merges the commits missed by @Fussmatte necessary for music to be properly reloaded. Partially fixes #271.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV (for
  example, a 2.3 update on Steam for Windows/macOS/Linux)
- [x] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes

PR #274, opened 2020-06-07
Add C++98/C++03 standards declaration to CMakeFiles.txt

InfoTeddy:
This enforces the C++03 standard for people making pull requests who may not realize their fancy features are too new and shouldn't be used (_\*cough*_, _\*cough*_, @leo60228).

I did some internet searching and this is what I got from this page: https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/

CI will fail because of #273 (i.e. it's Leo's fault).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -18,6 +18,10 @@ IF(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebI
 	SET(GOG ON)
 ENDIF()
 
+# Set standard to C++98/C++03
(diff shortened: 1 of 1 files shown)


leo60228:
> `CMAKE_CXX_STANDARD_REQUIRED`

This doesn't seem like what we want. This will cause an error if a compiler supports C++11 but not C++98, even though that should be allowed.

InfoTeddy:
Fair enough, but removing it does mean there's less chances for someone to discover their mistake of using post-C++03 features.

flibitijibibo:
This'll break the current build but that's fine for the moment...

PR #280, opened 2020-06-13
Put if-guards around usages of entity-getting functions whose return values can be out-of-bounds

InfoTeddy:
There are three functions that return a sentinel value of -1 if the entity it is looking for doesn't exist: `obj.getteleporter()`, `obj.getplayer()`, and `obj.getcompanion()`. There are many places in the code that blindly use the return values from those functions without checking that it isn't -1 first, and could thus end up indexing out-of-bounds, which is bad bad Undefined Behavior.

So to fix the potential for UB, this PR adds if-guards to all of those cases.

## Legal Stuff:

  for these changes


flibitijibibo:
Part of me wonders if we should have asserts for this, but either way this is pretty important to avoid!

PR #293, opened 2020-06-15
Make elephant not be flashy if screen effects are disabled

InfoTeddy:
The flashy color of the elephant can be hard on people's eyes, especially if they're the type who want screen effects disabled because they might have epilepsy. The elephant takes up a good 3/4ths of the screen, you know. If screen effects are disabled, the elephant will use color 22, which is a neutral gray.

I'm only adding this because the VVVVVV speedrun mods (@tzann, @mohoc) invalidate all runs that have the elephant texture removed, even though many people would be looking at a potentially epilepsy-inducing image many times a day grinding 100% speedruns. (Imo, their justification for this is flimsy at best.)

## Legal Stuff:

  for these changes


TerryCavanagh:
Ahh, thanks for this one. I've wanted to do it for years. This really should have been the case since the start.

PR #295, opened 2020-06-16
Support hex entities in metadata

leo60228:
## Changes:

This adds support for hex entities in level metadata.

## Legal Stuff:

  for these changes


TerryCavanagh:
I guess this one is leading towards eventually supporting arbitrary coloured crewmates? 😅 Looks like a good start!

InfoTeddy:
Nah, this is just a fix of a regression from implementing the faster level loading. Previously, TinyXML would've handled decoding XML decimal/hex entities.

PR #304, opened 2020-06-18
Add note to README about "compiler quirks" only applying to 2.2

InfoTeddy:
Only the 2.2 version has these problems, so you don't need to worry about this section when compiling 2.3.

## Legal Stuff:

  for these changes


flibitijibibo:
Will double check this with my crappy VS2010 setup, but I’m pretty sure you’re right...

PR #310, opened 2020-06-21
Add `-playassets` command-line option, and clean up and fix filesystem and command-line arguments code

### LINKED ISSUE #309: Level-specific custom assets with pipes
Piping levels into VVVVVV was added recently, and so was level-specific custom assets. The level-specific custom assets load assets from a folder with the same name as the level files name, but when you pipe a level in, the level filename stored is "special/stdin.vvvvvv". This is an issue because this means you wouldn't be able to have level-specific custom assets for piped-in levels without renaming folders and such every time you want to pipe a different level in.

A good workaround would be adding a new argument, like `-ad`/`--assets-dir` which lets you specify the name of the directory to load custom assets from. This would mean editors that pipe the level data into V6 would still be able to use the custom assets from the level if it wasn't piped in.

InfoTeddy:
If the game was loading a level file from STDIN, and that level file had custom assets, they wouldn't be loaded in. This is because of the fundamental problem where if the game loads a level file from STDIN, it doesn't know the actual filename of the level. To fix this, it needs to be able to be provided the name of the actual filename, so it can load the assets.

After this is merged, a trivial amount of work (compared to this patch) just needs to be done on Ved's end.

Also cleaned up and fixed some filesystem/CLI code (read the commits).

Anyone who wants to test this, but doesn't know how to or want to modify Ved to make it use `-playassets`, but knows how to use the command-line... don't abuse `cat`! I went out of my way to figure out how to avoid abusing `cat`. The command I used was:
```
./VVVVVV -p special/stdin -playassets custom_assets_test < ~/v/levels/open_1.1.3.vvvvvv
```

Fixes #309.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -166,6 +166,44 @@ void FILESYSTEM_mount(const char *fname)
 	}
 }
 
+void FILESYSTEM_mountassets(const char* path)
+{
+	const std::string _path(path);
+
+	std::string zippath = "levels/" + _path.substr(7,_path.size()-14) + ".data.zip";
+	std::string dirpath = "levels/" + _path.substr(7,_path.size()-14) + "/";
+	std::string zip_path;
+	const char* cstr = PHYSFS_getRealDir(_path.c_str());
+
+	if (cstr) {
+		zip_path = cstr;
+	}
+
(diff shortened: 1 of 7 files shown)


flibitijibibo:
Fix makes sense to me - @AllyTally, if this works for you I'll merge this in.

AllyTally:
Yep, this works great! Thank you.

PR #317, opened 2020-06-28
Fix the two-frame-delay when entering a room with an "init" script

InfoTeddy:
This patch is very kludge-y, but at least it fixes a semi-noticeable visual issue in custom levels that use internal scripts to spawn entities when loading a room.

Basically, the problem here is that when the game checks for script boxes and sets `newscript`, `newscript` has already been processed for that frame, and when the game does load a script, `script.run()` has already been processed for that frame.

That issue can be fixed, but it turns out that due to my over-30-FPS game loop changes, there's now *another* visible frame of delay between room load and entity creation, because the render function gets called in between the script being loaded at the end of `gamelogic()` and the script actually getting run.

So... I have to temporary move `script.run()` to the end of `gamelogic()` (in `map.twoframedelayfix()`), and make sure it doesn't get run next frame, because double-evaluations are bad. To do that, I have to introduce the kludge variable `script.dontrunnextframe`, which does exactly as it says.

And with all that work, the two-frame (now three-frame) delay is fixed.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -4780,6 +4780,7 @@ void entityclass::entitycollisioncheck()
         }
     }
 
+    // WARNING: If updating this code, don't forget to update Map.cpp mapclass::twoframedelayfix()
     activetrigger = -1;
     if (checktrigger() > -1)
     {
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -1753,6 +1753,7 @@ void Game::updatestate()
             break;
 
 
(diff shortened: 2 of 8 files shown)


flibitijibibo:
Makes sense to me. You may want to change the warning to explicitly name "Map.cpp mapclass::twoframedelayfix()" so people can Ctrl+F straight to it, other than that this is good to go.

InfoTeddy:
Alright, the warning has been changed to explicitly name "Map.cpp mapclass::twoframedelayfix()".

PR #321, opened 2020-06-28
Add bounds check to textbox functions that use `m`

InfoTeddy:
It seems to be a bit bad to blindly use `m` without checking it. In fact, this has caused a few segfaults already, actually.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -1143,6 +1143,7 @@ void Graphics::textboxtimer( int t )
 {
     if (!INBOUNDS(m, textbox))
     {
+        puts("textboxtimer() out-of-bounds!");
         return;
     }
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Does it happen in normal gameplay? Should we log the error?

InfoTeddy:
Well, not in the main game, no. There's a level I'm working on that uses black magic with the 20 textbox limit and it triggers this OoB indexing.

flibitijibibo:
We should add a quick (one-time?) log if it's only custom data - someone using the editor may want to know about this.

InfoTeddy:
Mk, how do I add a one-time log, then?

flibitijibibo:
I usually just do

```
if (badThing)
{
    static bool whine = true;
    if (whine)
    {
        Log("Bad thing!");
        SDL_assert(0 && "Bad thing!"); // Optional
    }
}

InfoTeddy:
Ah, I see. I'll push something like that shortly.

InfoTeddy:
Actually, after reviewing the black magic trick further, this OoB indexing doesn't happen too often, so I don't need to make the logs one-time.

PR #322, opened 2020-06-29
Make menus automatically centered and narrowed

Dav999-v:
## Changes:

All menus had a hardcoded X position (offset to an arbitrary starting point of 110) and a hardcoded horizontal spacing for the "staircasing" (mostly 30 pixels, but for some specific menus hardcoded to 15, 20 or something else). Not all menus were centered, and seem to have been manually made narrower (with lower horizontal spacing) whenever text ran offscreen during development.

This system may already be hard to work with in an English-only menu system, since you may need to adjust horizontal spacing or positioning when adding an option. The main reason I made this change is that it's even less optimal when menu options have to be translated, since maximum string lengths are hard to determine, and it's easy to have menu options running offscreen, especially when not all menus are checked for all languages and when options could be added in the middle of a menu after translations of that menu are already checked.

Now, menus are automatically centered based on their options, and they are automatically made narrower if they won't fit with the default horizontal spacing of 30 pixels (with some padding). The game.menuxoff variable for the menu X position is now also offset to 0 instead of 110

The _default_ horizontal spacing can be changed on a per-menu basis, and most menus (not all) which already had a narrower spacing set, retain that as a maximum spacing, simply because they looked odd with 30 pixels of spacing (especially the main menu). They will be made even narrower automatically if needed. In the most extreme case, the spacing can go down to 0 and options will be displayed right below each other. This isn't in the usual style of the game, but at least we did the best we could to prevent options running offscreen.

The only exception to automatic menu centering and narrowing is the list of player levels, because it's a special case and existing behavior would be better than automatic centering there.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -7469,7 +7469,7 @@ void Game::createmenu( enum Menu::MenuName t, bool samemenu/*= false*/ )
     // If it doesn't fit onscreen (including square brackets), reduce the horizontal spacing by 5 and retry.
     bool done_once = false;
     int menuwidth = 0;
-    for (; !done_once || (menuwidth > 272 && maxspacing > 0); maxspacing -= 5)
+    for (; !done_once || (menuwidth > 272 && menuspacing > 0); maxspacing -= 5)
     {
         done_once = true;
         menuspacing = maxspacing;
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -7466,7 +7466,9 @@ void Game::createmenu( enum Menu::MenuName t, bool samemenu/*= false*/ )
     }
 
     // Automatically center the menu. We must check the width of the menu with the initial horizontal spacing.
-    // If it doesn't fit onscreen (including square brackets), reduce the horizontal spacing by 5 and retry.
+    // If it's too wide, reduce the horizontal spacing by 5 and retry.
+    // Try to limit the menu width to 272 pixels: 320 minus 16*2 for square brackets, minus 8*2 padding.
+    // The square brackets fall outside the menu width (i.e. selected menu options are printed 16 pixels to the left)
     bool done_once = false;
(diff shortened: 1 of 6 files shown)


desktop_version/src/Game.cpp
```diff
@@ -7486,15 +7457,31 @@ void Game::createmenu( enum Menu::MenuName t, bool samemenu/*= false*/ )
...
+    bool done_once = false;
+    int menuwidth = 0;
+    for (; !done_once || (menuwidth > 272 && menuspacing > 0); maxspacing -= 5)
+    {
+        done_once = true;
+        menuspacing = maxspacing;
+        menuwidth = 0;
+        for (size_t i = 0; i < menuoptions.size(); i++)
+        {
+            int width = i*menuspacing + graphics.len(menuoptions[i].text);
```
InfoTeddy, line 7479: The comment above says this counts square brackets, yet I don't see where in this line it accounts for them.

Dav999-v: Yeah that could have probably been clearer.

The menu is tried to be limited to 272 pixels (see the outer `for` loop). That is 320 (screen width) minus 16\*2 for brackets on each side, minus 8\*2 padding because it looks better to not extend to the borders of the screen. In the functions that draw menu options, the X positions of the menu options are where the text starts, when selected, the option including brackets are displayed 16 pixels more to the left.

InfoTeddy: Yeah, makes sense.

flibitijibibo: Let's document all that in this block, I was going to ask the same thing... other than that this seems fine to me, this will probably make life a lot easier if we ever have to add vertical menu scrolling.

Dav999-v: Done in 803d0f45de320d38cfe24a05595361764ebb5362.

PR #333, opened 2020-06-30
Move some options to a new menu, "advanced options"

Dav999-v:
## Changes:

VVVVVV's menus are kind of packed to the brim, so I thought it was time to recategorize the menus a little bit. There's now a new "advanced options" menu which holds the following options which were moved out of graphic options, game options and especially accessibility options:

- toggle mouse
- unfocus pause
- fake load screen
- room name background
- glitchrunner mode

I also moved the "resize to nearest" option to be grouped more logically and changed "toggle letterbox" to "scaling mode" (further explained in the commit messages).

## Legal Stuff:

  for these changes


flibitijibibo:
The toggle letterbox one in particular was definitely a case of "no, I totally didn't rush this in as soon as my commit went into SDL mainline, what are you talking about..."

This all makes sense to me!

PR #335, opened 2020-06-30
Re-color one-way tiles to match the tileset

InfoTeddy:
One-ways have always had this problem where they're always yellow. That means unless you specifically use yellow, it'll never match the tileset.

The best way to fix this without requiring new graphics or changing existing ones is to simply re-tint the one-way with the given color of the room. That way, the black part of the tile is still black, but the yellow is now some other color.

But existing custom levels already re-texture the one-way tiles and some of them probably don't want the re-color. So if custom assets are mounted, then don't re-color the one-ways.

But new custom levels might want to take advantage of the re-color anyway, so there's now an override switch if a certain tag is present in the XML to use the re-color anyway even if assets are mounted.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -165,6 +165,8 @@ void FILESYSTEM_mount(const char *fname)
 	}
 }
 
+bool FILESYSTEM_assetsmounted = false;
+
 void FILESYSTEM_mountassets(const char* path)
 {
 	const std::string _path(path);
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Makes sense to me - do the inbound checks fail rarely enough to warrant logs?

InfoTeddy:
Yeah, they basically never fail. I don't see any reason to log them.

flibitijibibo:
If they never fail we probably _should_ log them - if this happens we'll want to know about it.

InfoTeddy:
Alright, then I'll add the logs. This happens in rendering code, though, so I'll make them one-time only.

InfoTeddy:
I've added the one-time logs.

PR #337, opened 2020-07-01
Parse all headers from binary blobs

InfoTeddy:
Previously, the game didn't parse all headers from a given binary blob. It would look for headers with very specific names and only parse those.

This patch makes it so that all headers will be parsed and loaded. Headers with said specific names will still get priority and be loaded first, though, in case there happen to be binary blobs where the index of one of those headers isn't in the range 0..16. Then the extra headers are parsed in order according to their index.

`musicclass::play()` has been adjusted accordingly to accomodate for the new tracks, and the `music(-1)` trick still works (it plays the last MMMMMM track if you have MMMMMM and you're on PPPPPP).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BinaryBlob.h
+++ b/desktop_version/src/BinaryBlob.h
@@ -4,6 +4,24 @@
 /* Laaaazyyyyyyy -flibit */
 // #define VVV_COMPILEMUSIC
 
+#define TRACK_NAMES \
+	FOREACH_TRACK("data/music/0levelcomplete.ogg") \
+	FOREACH_TRACK("data/music/1pushingonwards.ogg") \
+	FOREACH_TRACK("data/music/2positiveforce.ogg") \
+	FOREACH_TRACK("data/music/3potentialforanything.ogg") \
+	FOREACH_TRACK("data/music/4passionforexploring.ogg") \
+	FOREACH_TRACK("data/music/5intermission.ogg") \
+	FOREACH_TRACK("data/music/6presentingvvvvvv.ogg") \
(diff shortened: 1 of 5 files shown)


desktop_version/src/BinaryBlob.h
```diff
@@ -1,9 +1,29 @@
 #ifndef BINARYBLOB_H
 #define BINARYBLOB_H
 
+#include <vector>
+
 /* Laaaazyyyyyyy -flibit */
 // #define VVV_COMPILEMUSIC
 
+#define TRACK_NAMES \
+	X("data/music/0levelcomplete.ogg") \
```
flibitijibibo, line 10: Can we rename these macros to indicate a set of actions is being done? I saw the `X()` action in BinaryBlob.cpp first and I had no idea what was going on... something like:

```
#define TRACK_ACTIONS \
    FOREACH_TRACK("etc")
```

InfoTeddy: Done, rebased and force-pushed.

InfoTeddy: Actually, I'll change `TRACK_NAMES` to `TRACK_ACTIONS` (missed it the first time I read your comment!)

InfoTeddy: Nevermind actually, I don't think `TRACK_ACTIONS` is clearer than `TRACK_NAMES`.

InfoTeddy:
Yeah, it works fine with that terrible music file. Just to be sure, though, I made it so that an invalid `valid` value is properly set to false (0), so my code won't be parsing those garbage headers in the first place.

PR #350, opened 2020-07-04
Remove use of STL from things that are always fixed-size, among other things

InfoTeddy:
There are many vectors that always stay the same size throughout the program. There's no need to make them a vector, instead of just a plain array.

Also I did some other cleanups, like removing unnecessary or unused attributes.

### Removed attributes

- `scriptclass::tempword`
- `scriptclass::currentletter`
- `mapclass::temp`
- `mapclass::temp2`
- `editorclass::temp`
- `UtilityClass::temp`
- `UtilityClass::temp2`
- `UtilityClass::globaltemp`
- `Graphics::bfontmask_rect`
- `Graphics::backgrounds`
- `Graphics::bfontmask`
- `GraphicsResources::im_bfontmask`

### De-vectored

- `scriptclass`
  - `tokenize()`: `std::string currentletter` -> `char currentletter`
  - `std::vector<std::string> words` -> `std::string words[40]`

- `Game`
  - `std::vector<int> besttimes` -> `int besttimes[6]`
  - `std::vector<int> besttrinkets` -> `int besttrinkets[6]`
  - `std::vector<int> bestlives` -> `int bestlives[6]`
  - `std::vector<int> bestrank` -> `int bestrank[6]`
  - `std::vector<bool> crewstats` -> `bool crewstats[6]`
  - `std::vector<bool> tele_crewstats` -> `bool tele_crewstats[6]`
  - `std::vector<bool> quick_crewstats` -> `bool quick_crewstats[6]`
  - `std::vector<const char*> superpatrons` -> See below
  - `std::vector<const char*> patrons` -> See below
  - `std::vector<const char*> githubfriends` -> See below

- `MenuOption`
  - `std::string text` -> `char text[41]`

- `entityclass`
  - `std::vector<bool> flags` -> `bool flags[100]`
  - `std::vector<bool> collect` -> `bool collect[100]`
  - `std::vector<bool> customcollect` -> `bool customcollect[100]`

- `mapclass`
  - `std::vector<bool> explored` -> `bool explored[400]`
  - `std::vector<bool> roomdeaths` -> `bool roomdeaths[400]`
  - `std::vector<bool> roomdeathsfinal` -> `bool roomdeathsfinal[400]`
  - `std::vector<int> areamap` -> `bool areamap[400]`
  - `std::vector<int> contents` -> `int contents[1200]`
  - `std::vector<int> vmult` -> `int vmult[30]`
  - `std::vector<std::string> specialnames` -> `std::string specialnames[8]`

- `towerclass`
  - `std::vector<int> back` -> `int back[40 * 120]`
  - `std::vector<int> contents` -> `int contents[40 * 700]`
  - `std::vector<int> minitower` -> `int minitower[40 * 100]`
  - `std::vector<int> vmult` -> `int vmult[40 * 700]`

- `Graphics`
  - `std::vector<SDL_Rect> stars` -> `SDL_Rect stars[50]`
  - `std::vector<int> starsspeed` -> `int starsspeed[50]`
  - `std::vector<SDL_Rect> backboxes` -> `SDL_Rect backboxes[18]`
  - `std::vector<int> backboxvx` -> `int backboxvx[18]`
  - `std::vector<int> backboxvy` -> `int backboxvy[18]`
  - `std::vector<float> backboxint` -> `float backboxint[18]`
  - `std::vector<int> fadebars` -> `int fadebars[15]`

- `UtilityClass`
  - `std::vector<int> splitseconds` -> `int splitseconds[30]`

- `editorclass`
  - `std::vector<int> contents` -> `int contents[40*30 * 20*20]`
  - `std::vector<int> vmult` -> `int vmult[30 * 20]`

### De-vectoring of `superpatrons`, `patrons`, and `githubfriends`

Originally I made them `static const char* []`s, and initialized them in `Game.cpp`, but then the compiler complained that other files won't be able to know what's inside the array. To do that, I'd have to initialize it in `Game.h`, but then that'd mean every time someone got added to the list it'd trigger a re-compile of 20 or so files.

So I just moved the lists (along with `creditmaxposition`) to a new file, `Credits.h`. That way, only the minimal amount of files will need to be recompiled when someone is added to the list.

### Other

I removed all instances where `std::transform()` was used.

The game now uses `SDL_toupper()` and `SDL_tolower()` instead of libc `toupper()` and `tolower()`.

`Graphics::drawmenu()` and `Graphics::drawlevelmenu()` now have less copy-pasted code. In particular, `Graphics::drawlevelmenu()` has been removed and its special-case y-positioning consolidated into `Graphics::drawmenu()`.

## Legal Stuff:

  for these changes


desktop_version/src/Game.h
```diff
@@ -10,7 +10,7 @@
 
 struct MenuOption
 {
-    std::string text;
+    char text[41]; // 40 chars covers the entire screen, + 1 more for null terminator
```
Dav999-v, line 13: You're overlooking UTF-8 not using single bytes per character here, so the buffers would need to be 161 chars to fit the longest possible visible string, especially when menus are translated (ditto the other temporary strings).

InfoTeddy: Fixed, rebased and force-pushed.

Dav999-v: Why not simply add an extra commit instead of rebasing and force pushing, if I may ask? Your commit description is now inconsistent with the action taken in that commit.

InfoTeddy: Because I like having a clean commit history. [Here's an example of what I want to avoid](https://github.com/FNA-XNA/FNA/pull/280#issuecomment-580060023).

Nevertheless, I've rebased and force-pushed to correct that commit description.

PR #354, opened 2020-07-06
Clean up constructors and initializations

InfoTeddy:
Another PR of *code quality improvement* (that's the name of the branch).

This time, I re-added constructors to every class that has member fields, and only the things that need to be strictly initialized in the delayed `init()` function are initialized there (namely, stuff that depends on the filesystem being initialized already). Most notably, `entityclass` no longer has a delayed `init()` function (and it's had said `init()` function since at least 2.2!).

Then I ran the codebase through `cppcheck` and found out every last member field that wasn't initialized, and properly initialized them. And there were a lot of them, actually.

I would've changed the constructors to (mostly) use initialization lists, but the compiler gives off a bunch of warning about the fact that the initialization list isn't the order in which the member fields will be initialized. Apparently this is just to prevent errors with regards to initialization of member fields that depend on other member fields being initialized first, but I don't think the game does that anywhere, so I just didn't bother with the whole thing.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -11,6 +11,7 @@ binaryBlob::binaryBlob()
 {
 	numberofHeaders = 0;
 	SDL_memset(m_headers, 0, sizeof(m_headers));
+	SDL_memset(m_memblocks, 0, sizeof(m_memblocks));
 }
 
 #ifdef VVV_COMPILEMUSIC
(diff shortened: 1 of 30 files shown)


flibitijibibo:
This seems fine but it does change the program flow a fair bit, so I'll save this for 2.4 - if any Valgrind errors pop up from variables that weren't initialized until this patch, the fix should be backported if possible.

InfoTeddy:
You'd be willing to merge everything other than the part where I add proper constructors, though, right?

flibitijibibo:
Yeah, those commits can be cherry-picked if needed.

PR #359, opened 2020-07-09
De-duplicate case 12 (custom levels) in mapclass::loadlevel(), mostly entity creation

InfoTeddy:
Case 12 in `mapclass::loadlevel()` is a hot mess. A lot of it is a bunch of copy-pasted code, especially the part where edentities are converted into actual entities by a bunch of `obj.createentity()` calls. Luckily, I'm here to clean it all up and make it better.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -1611,6 +1611,7 @@ void mapclass::loadlevel(int rx, int ry)
 #endif
 #if !defined(NO_CUSTOM_LEVELS)
 	case 12: //Custom level
+	{
 		const int curlevel = rx-100 + (ry-100) * ed.maxwidth;
 		const edlevelclass* room_ptr = NULL;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
b1c0eed: This adds some more calculations done to all cases, rather than just the 2 cases where the bounding math was present.
512750f7: Same here, this does an assignment for cases that may not be those three values. We would need to assert that no other cases are possible, at which point we could probably inline most of the other statements in the switch block as well.

The other commits look okay to me.

InfoTeddy:
GitHub doesn't seem to work properly right now. I have rebased and force-pushed to resolve conflicts with #361 and address your issues (namely I dropped the `graphics.rcol` commit along with making sure the bounding logic is only calculated for enemies and moving platforms), but it doesn't seem to be updating for this PR.

flibitijibibo:
Rebased manually, again...

PR #364, opened 2020-07-11
Fix variables that weren't being reset when exiting out of playtesting (some of them specifically when you exited playtesting by rescuing the last crewmate in the level), and fix some input relating to teleporters

InfoTeddy:
I intended to fix one bug, but then got carried away and fixed a bunch of others. And so it goes.

In no particular order, the bugs fixed were:

 - The horizontal/vertical warp background is not reset if exiting playtesting via rescuing the last crewmate, leading to the background of the room you rescued the crewmate in scrolling off in the room you re-entered the editor in, instead of the background of the new room being properly drawn.
 - Warp dirs changed by the `warpdir()` command are not properly reset if exiting playtesting via rescuing the last crewmate.
 - Rescuing the last crewmate during playtesting and exiting to editor via pressing Up or Down on the "- Press ACTION to advance text -" prompt moves you to the room above or below the one you started playtesting from. Most noticeable and annoying if the room you started playtesting from is also the room you exited playtesting room.
 - Pressing Esc to exit playtesting immediately brings up settings menu, unless you have frame-perfect skill and only press it for 1 frame.
 - Text boxes are not being properly reset when exiting or entering playtesting, causing a one-frame glitch when entering playtesting if you exited playtesting with text boxes onscreen.
 - Pressing Enter on teleporters during playtesting exits playtesting instead.
 - Pressing Esc in the teleporter menu screen brings you to the "Do you want to quit?" screen instead of the pause menu.
 - The "Game Saved" text box is shown when touching a teleporter in a custom level.
 - The editor note is not reset when exiting playtesting.
 - The editor roomname hiding is not reset when exiting playtesting.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -1878,8 +1878,7 @@ void Game::updatestate()
                 }
                 else
                 {
-                    gamestate = EDITORMODE;
-                    graphics.backgrounddrawn=false;
+                    shouldreturntoeditor = true;
                     if(!muted && ed.levmusic>0) music.fadeMusicVolumeIn(3000);
                     if(ed.levmusic>0) music.fadeout();
                 }
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -7308,6 +7308,8 @@ void Game::returntoeditor()
 
     ed.keydelay = 6;
     ed.settingskey = true;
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Rebased manually since GitHub's merge button is down (lol)

PR #385, opened 2020-07-14
Optimize 30-mode FPS loop to use SDL_Delay()

InfoTeddy:
This patch optimizes the loop used to limit the framerate in 30-FPS-only mode so that it uses `SDL_Delay()` instead of an accumulator. This means that the game will take up less CPU power in 30-FPS-only mode. This also means that the game loop code has been simplified, so there's only two while-loops, and only two places where `game.over30mode` is checked, thus leading to easier-to-understand logic.

Using an accumulator here would essentially mean busywaiting until the 34 millisecond timer was up. (The following is just what @leo60228 told me.) Busywaiting is bad because it's inefficient. The operating system assumes that if you're busywaiting, you're performing a complex calculation and handles your process accordingly. And this is why sleeping was invented, so you could busywait without taking up unnecessary CPU time.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -62,9 +62,7 @@ volatile Uint32 timePrev = 0;
 volatile Uint32 accumulator = 0;
 volatile Uint32 f_time = 0;
 volatile Uint32 f_timePrev = 0;
-volatile Uint32 f_accumulator = 0;
 
-void inline gameloop();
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This will most likely fix Valgrind performance as well, since it's not spinning all the time anymore. Glad sleeping still works!

PR #396, opened 2020-07-19
Change all tilemaps to be short[1200] instead of int[1200]

InfoTeddy:
This removes around megabyte from the binary, so a stripped -Og binary went from 4.0 megabytes to 2.9 megabytes, and an unstripped -O0 binary went from 8.1 megabytes to 7.1 megabytes, which means I can now finally upload an unstripped -O0 binary to Discord without having to give money to Discord for their dumb Nitro thing or whatever.

## Legal Stuff:

  for these changes


flibitijibibo:
As discussed internally, making an exception to the freeze for this rule as it's a pretty straightforward change and is easy to audit, just takes some time...

PR #397, opened 2020-07-19
Clean up includes so other headers are not unnecessarily included in header files, and neatly organize and order them

InfoTeddy:
The game used to include a bunch of header files inside other header files, which would result in a few unnecessary rebuilds of files whenever the inner included header got changed. Now I've "flattened" it so that header includes are mainly done in the `.cpp` file, and the only includes in `.h` files are the ones that are strictly necessary (e.g. if a class has another class inside it as a class member of that class). Except for the TinyXML-2 include, because it's a huge header file and I'd rather just forward declare it instead of including it in a bunch of unnecessary places.

Also, include order has been alphabetized, and files included follow a certain pattern of sections: there's the include of the "main" header file (e.g. `#include "Graphics.h"` for `Graphics.cpp`), then system or third-party includes, then includes from the rest of the project. And additionally sometimes there's a section where you have to do something special, like `#define`s or `#ifdef`s to include some other files (see `FileSystemUtils.cpp` and `editor.cpp` for examples of this).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -1,11 +1,9 @@
 #include "BinaryBlob.h"
 
+#include <physfs.h> /* FIXME: Abstract to FileSystemUtils! */
+#include <SDL.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <SDL.h>
-
-/* FIXME: Abstract to FileSystemUtils! */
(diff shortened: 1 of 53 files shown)


flibitijibibo:
Needs a rebase but otherwise seems fine.

InfoTeddy:
Rebased and force-pushed.

PR #407, opened 2020-07-29
Fix the 'Game Complete' trophy stand so it no longer hovers above the ground

tzann:
## Changes:

The 'Game Complete' Trophy stand hovers above the ground. This change fixes that. Also, I changed the comments in Otherlevel.cpp so they say trophy instead of terminal, since that's what's actually being created.
Here is what it used to look like:
image

Note that the trophy position (line 8953 in Otherlevel.cpp) was altered by 3 pixels originally. This has been removed and instead the position is only altered if the trophy is shown.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -13,6 +13,7 @@ Contributors
 * Daniel Lee (@ddm999)
 * Fredrik Ljungdahl (@FredrIQ)
 * Matt Penny (@mwpenny)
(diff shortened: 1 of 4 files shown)


InfoTeddy:
Also you should add your name to `CONTRIBUTORS.txt` and `githubfriends` in `Credits.h`.

flibitijibibo:
Changes lgtm, as @InfoTeddy mentioned you can add yourself to those files and I'll get everything in together.

PR #408, opened 2020-07-29
Wrap entities

### LINKED ISSUE #404: Vertical entities are not drawn wrapped in rooms that warp on all sides
Vertical entities don't visually wrap around

This seems to be due to the fact that in `Graphics::drawentities()`, the logic goes like `if (map.warpx) { ... } else if (map.warpy) { ... }` instead of `if (map.warpx) { ... } if (map.warpy) { ... }` (`if`-`else if` instead of `if`-`if`).

tzann:
## Changes:

Fix Graphics::drawentities() screen wrapping logic to allow for simultaneous horizontal and vertical wrapping of sprites. Closes #404.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -1659,26 +1659,22 @@ void Graphics::drawentities()
                 wrappedPoint.y -= 230;
             }
 
-            if (map.warpx ||
-                (map.towermode && !map.minitowermode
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Yeah, uh, this doesn't work.

All sides wrap fail

InfoTeddy:
Nevermind, it actually works. @flibitijibibo you should probably do a squash merge though, commit history is a bit messy...

flibitijibibo:
These changes make sense to me as well.

PR #409, opened 2020-07-29
Fix checkpoint starts in editor

### LINKED ISSUE #378: Editor spawns you offset from the checkpoint when starting playtesting
If you spawn at a checkpoint when entering playtesting, the editor won't put you on the exact coordinates of the checkpoint. Instead, it will spawn you a bit to the right of the checkpoint, and a bit further upwards if it's unflipped or a bit further downwards if it's flipped.

This is annoying if you have spikes on the right side of the checkpoint, or have it on the top if you're unflipped or bottom if you're flipped, as Viridian will unnecessarily die once before respawning at the proper coordinates. It is even more annoying if you have tiles on the right side of the checkpoint, as Viridian will clip into the wall and move upwards if unflipped or downwards if flipped and you have to press Left to get out of the wall.

tzann:
## Changes:

This makes starting from a checkpoint in the editor put you in the same position as when respawning from it after a death. Closes #378.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -4711,19 +4711,12 @@ void editorinput()
                             //Checkpoint spawn
                             int tx=(edentity[testeditor].x-(edentity[testeditor].x%40))/40;
                             int ty=(edentity[testeditor].y-(edentity[testeditor].y%30))/30;
-                            game.edsavex = (edentity[testeditor].x%40)*8;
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Oh so that's why the y-offset has been happening... cuz the code swapped around the y-correction for being flipped and unflipped!

Looks good to me.

PR #410, opened 2020-07-29
Improve FileSystemUtils.cpp portability

leo60228:
## Changes:

These tweaks allow VVVVVV to be compiled for a new Unix-like platform with no source changes.

## Legal Stuff:

  for these changes


flibitijibibo:
Will save this for after 2.3 - the plan is to axe save migration after 2.3 is tagged (2.2 was 6 years ago!) so the new CMake option will end up getting cut anyway.

InfoTeddy (changes requested):
Remove the save migration stuff (since it's already removed now), give this PR a rebase, and it's good to go.

leo60228:
Done.

PR #412, opened 2020-07-30
Fallback to loading music from loose files if vvvvvvmusic.vvv is invalid or missing

leo60228:
## Changes:

Files are loaded from paths like `/home/leo60228/.local/share/VVVVVV/music/6presentingvvvvvv.ogg`. This allows the game to have less than 20MB of peak heap usage (though still ~100MB of non-heap RAM usage, which I'm investigating).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -174,6 +174,7 @@ set(PFS_SRC
 	../third_party/physfs/physfs_platform_unix.c
 	../third_party/physfs/physfs_platform_windows.c
 	../third_party/physfs/physfs_platform_haiku.cpp
+	../third_party/physfs/physfsrwops.c
 )
 if(APPLE)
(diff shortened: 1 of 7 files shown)


InfoTeddy (approved):
If @flibitijibibo is fine with the RWops stuff, then this is good to go after a rebase.

flibitijibibo:
Makes sense to me - the only thing that seemed interesting was the way it was added to CMake, but then I remembered that SDL complicates it enough to where it wouldn't be easy to add to the physfs blob, so this is good enough.

leo60228:
Rebased. I can't test at the moment but it compiles fine.

leo60228:
stdout is as expected for both vvvvvvmusic.vvv and loose files, which doesn't necessarily mean anything but is a good sign.

leo60228:
Tested and working.

PR #428, opened 2020-08-04
Crop bottom tower spikes so they don't get drawn behind room name

InfoTeddy:
If you have the translucent room name option enabled, you'd always be seeing the spikes at the bottom of the screen hidden behind the room name. This patch makes it so that the spikes get carefully cropped so they only appear above the room name when the player gets close to the bottom of the screen.

Spikes hidden behind room name

Spikes above room name

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -683,7 +683,7 @@ void Graphics::drawtile2( int x, int y, int t )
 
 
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Merged #427, once this is rebased it's good to go.

InfoTeddy:
Rebased and force-pushed.

PR #429, opened 2020-08-04
Actually finally fix rendering of translucent pixels on tiles (remove OverlaySurfaceKeyed() and set proper foregroundBuffer blend mode)

InfoTeddy:
So, earlier in the development of 2.0, Simon Roth (I presume) encountered a problem: Oh no, all my backgrounds aren't appearing! And this is because my `foregroundBuffer`, which contains all the drawn tiles, is drawing complete black over it!

So he had a solution that seems ingenius, but is actually really really hacky and super 100% *not* the proper solution. Just, take the `foregroundBuffer`, iterate over each pixel, and *don't* draw any pixel that's `0xDEADBEEF`. `0xDEADBEEF` is a special signal meaning "don't draw this pixel". It is called a 'key'.

Unfortunately, this causes a bug where translucent pixels on tiles (pixels between 0% and 100% opacity) didn't get drawn correctly. They would be drawn against this weird blue color.

Now, in #103, I came across this weird constant and decided "hey, this looks awfully like that weird blue color I came across, maybe if I set it to `0x00000000`, i.e. complete and transparent black, the issue will be fixed". And it *did* appear to be fixed. However, I didn't look too closely, nor did I test it that much, and all that ended up doing was drawing the pixels against black, which more subtly disguised the problem with translucent pixels.

So, after some investigation, I noticed that `BlitSurfaceColoured()` was drawing translucent pixels just fine. And I thought at the time that there was something wrong with `BlitSurfaceStandard()`, or something. Further along later I realized that all drawn tiles were passing through this weird `OverlaySurfaceKeyed()` function. And removing it in favor of a straight `SDL_BlitSurface()` produced the bug I mentioned above: Oh no, all the backgrounds don't show up, because my `foregroundBuffer` is drawing pure black over them!

Well... just... set the proper blend mode for `foregroundBuffer`. It should be `SDL_BLENDMODE_BLEND` instead of `SDL_BLENDMODE_NONE`.

Then you don't have to worry about your transparency at all. If you did it right, you won't have to resort this hacky color-keying business.

_\*sigh*_

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -2406,8 +2406,7 @@ void Graphics::drawmap()
         }
         foregrounddrawn = true;
     }
-    OverlaySurfaceKeyed(foregroundBuffer, backBuffer, 0x00000000);
-    //SDL_BlitSurface(foregroundBuffer, NULL, backBuffer, NULL);
+    SDL_BlitSurface(foregroundBuffer, NULL, backBuffer, NULL);
 
 }
 
@@ -2433,7 +2432,7 @@ void Graphics::drawfinalmap()
         foregrounddrawn=true;
     }
 
-    OverlaySurfaceKeyed(foregroundBuffer, backBuffer, 0x00000000);
+    SDL_BlitSurface(foregroundBuffer, NULL, backBuffer, NULL);
 }
 
 void Graphics::drawtowermap()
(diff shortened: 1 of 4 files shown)


flibitijibibo:
This makes total sense to me, but as I remember from the original 2.2 work, SDL 2.0 surface blending makes no sense. This could probably do with a visual diff between older versions and upstream, just to be sure we're not continuing the bizarre whack-a-mole that is VVV and SDL_Surface blending absurdity.

InfoTeddy:
Well, I (and others) distinctly remember in older versions tiles being drawn with the most barely noticeable of transparency, like 90% or so. It hasn't been really reproducible in any builds of 2.3 I've seen, nor have I been able to get it to happen with older versions.

I did some testing just now and did a quick visual diff between older versions and upstream, pretty sure it's fine and the blending is a-ok.

InfoTeddy:
Rebased and force-pushed to resolve conflicts with #430.

PR #436, opened 2020-08-07
Fix potential Undefined Behavior from not checking input passed into atoi()

InfoTeddy:
If an invalid integer is passed into `atoi()`, whether or not the behavior is undefined seems to depend on which libc you're using. So, it's best to have a wrapper function that performs the check for you, to avoid any problems from invalid input.

Additionally, this is an opportunity to replace `atoi()` with `SDL_atoi()`. If libc has `atoi()`, then `SDL_atoi()` will still call that function, so we still need to check the input, but if that function isn't available `SDL_atoi()` will call `SDL_strtol()`, which doesn't have this problem.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -508,7 +508,7 @@ void Game::loadcustomlevelstats()
 
                 if (stat_el->GetText() != NULL)
                 {
-                    stat.score = atoi(stat_el->GetText());
+                    stat.score = help.Int(stat_el->GetText());
                 }
 
                 if (stat_el->Attribute("name"))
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Do we want to use this as an excuse to define the fallback value? A possible addition to Int() would be an "int fallback" parameter that's returned instead of 0 when the string isn't a number. This is pretty important for the configs, so while it somewhat feels like a new feature it may be important for hardening the config parser.

InfoTeddy:
Well... I can't think of any cases off the top of my head where we'd want a fallback value that isn't 0. If an element in a config file is missing, it gets a default value, anyway.

Regardless, I'll add an optional fallback parameter if people want to use it. It can't hurt to have it.

flibitijibibo:
It's mostly for values like the window width/height, but it's definitely a single-digit number of cases at most. That can be covered separately, will go ahead and merge this once the CI is done...

PR #440, opened 2020-08-08
Disable pressing R in No Death Mode

InfoTeddy:
All that pressing R does in No Death Mode is end your run. As a result, it'll only be pressed by accident, so it's better to just disable it instead.

It's not even useful to quick-restart, because it's faster to quit and go through the menu again than it is to wait through the Game Over screen.

Additionally, I removed the `game.deathseq<=0` conditional because it's unnecessary due to the if-statement already being inside a `game.deathseq == -1` conditional.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -1899,7 +1899,7 @@ void gameinput()
                     }
                 }
 
-                if (key.keymap[SDLK_r] && game.deathseq<=0)// && map.custommode) //Have fun glitchrunners!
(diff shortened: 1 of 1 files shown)


Dav999-v:
Technically, people might want to try getting the fastest time on a time trial regardless of the number of deaths (think particularly the Tower which can scroll up faster by dying) so I wouldn't disable it on time trials.

flibitijibibo:
Agreed, if you know you're doomed you may want to hit R to jump to the checkpoint and recover as quickly as possible.

InfoTeddy:
Amended and force-pushed.

PR #441, opened 2020-08-08
Fixed controllers not being able to navigate the map and menu with up and down buttons.

NicholeMattera:
## Changes:

While working on updating my homebrew port on the Switch I noticed the Map and Menu screens couldn't be navigated using the up and down buttons using a controller. Seeing as up and down is allowed for keyboard it seemed controllers should be the same.

## Legal Stuff:

  for these changes


flibitijibibo:
This has bothered me for years, thanks for the fix!

PR #444, opened 2020-08-08
Fix unwinnable save from rescuing Violet out of order

InfoTeddy:
You're intended to rescue Violet first, and not second, third, or fourth, and especially not last.

If you rescue her second, third, or fourth, your crewmate progress will be reset, but you won't be able to re-rescue them again. This is because Vitellary, Verdigris, Victoria, and Vermilion will be temporarily marked as rescued during the `bigopenworld` cutscene, so duplicate versions of them don't spawn during the cutscene, and then will be marked as missing later to undo it.

This first issue can be trivially fixed by simply toggling flags to prevent duplicates of them from spawning during the cutscene instead of fiddling with their rescue statuses.

However, there is still another issue. If you rescue Violet last, then you won't be warped to the Final Level, meaning you can't properly complete the game. This can be fixed by adding a `crewrescued() == 6` check to the Space Station 1 Level Complete cutscene. There is additionally a temporary unrescuing of Violet so she doesn't get duplicated during the `bigopenworld` cutscene, and I've had to move that to the start of the `bigopenworld` and `bigopenworldskip` scripts, otherwise the `crewrescued() == 6` check won't work properly.

I haven't added hooks for Intermission 1 or 2 because you're not really meant to play the intermissions with Violet (but you probably could anyway, there'd just be no dialogue).

Oh, and the pre-Final Level cutscene expects the player to already be hidden before it starts playing, but if you rescue Violet last the player is still visible, so I've fixed that. But there still ends up being two Violets, so I'll probably replace it with a special cutscene later that's not so nonsensical.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -2631,7 +2631,6 @@ void Game::updatestate()
                 state++;
                 statedelay = 30;
                 graphics.textboxremove();
-                crewstats[1] = 0; //Set violet's rescue script to 0 to make the next bit easier
                 teleportscript = "";
             }
             break;
@@ -2644,13 +2643,20 @@ void Game::updatestate()
             if(graphics.fademode==1)
             {
                 startscript = true;
-                if (nocutscenes)
+                if (crewrescued() == 6)
                 {
-                    newscript="bigopenworldskip";
+                    newscript = "startlevel_final";
                 }
                 else
                 {
-                    newscript = "bigopenworld";
+                    if (nocutscenes)
+                    {
+                        newscript="bigopenworldskip";
+                    }
+                    else
+                    {
+                        newscript = "bigopenworld";
+                    }
                 }
                 state = 0;
             }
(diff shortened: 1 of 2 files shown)


Nikostormkilla:
By the way, the easiest way and only way to rescue Violet second or just not first is to hold R during the opening cutscene as you'll spawn in the ship instead of the space station

InfoTeddy:
The *only* way? That's not true. You could use invincibility mode and go through the left exit of Exhaust Chute, then either rescue Vitellary or exit through the entrance of Space Station 2 to go to other zones.

tzann:
I would recommend not adding hooks for Intermission 1 and 2 because that would significantly change how speedruns are done.

InfoTeddy:
Yes, that too. That's another reason I won't be adding those hooks.

flibitijibibo:
Checking on this real quick - are there any remaining concerns with this fix that haven't been addressed yet?

InfoTeddy:
Pretty sure there aren't. Except for being able to have two Violets in a cutscene, but that can be a separate fix and isn't as critical as this one.

flibitijibibo:
Cool, just wanted to be sure - the script changes make sense to me, but you never know with continuity errors like these...

PR #446, opened 2020-08-12
Hotkey text input fix

### LINKED ISSUE #366: Some hotkeys still conflict with taking input
- [ ] Using Alt+F to toggle fullscreen while input is being taken types F.
- [ ] Typing N or M during text input of `ed.getlin()` and level metadata editing will toggle the music and game mutes.

NicholeMattera:
## Changes:

Fixes #366

- The fullscreen issue was solved by checking to see if any modifier buttons (Ctrl, Alt, or Super) were being held down on the key down event, and if so then we ignore the input on the text input event. This allows for future hotkeys to not cause an issue.
- The sound muting issue was because the `getlin` method in editor wasn't properly setting `ed.textentry` to true. The reason why this started happening was because the inEditor check in main.cpp:620 was changed to use just `ed.textentry` whereas before it used to also check `key.textentrymode`.
- I also went ahead and added myself as a contributor. (Hopefully that is okay.)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -12,6 +12,7 @@ Contributors
 * Allison Fleischer (AllisonFleischer)
 * Daniel Lee (@ddm999)
 * Fredrik Ljungdahl (@FredrIQ)
+* Nichole Mattera (@NicholeMattera)
 * Matt Penny (@mwpenny)
 * Tynan Richards (@tzann)
 * Elliott Saltar (@eboyblue3)
--- a/desktop_version/src/Credits.h
+++ b/desktop_version/src/Credits.h
@@ -93,6 +93,7 @@ static const char* githubfriends[] = {
     "Allison Fleischer",
     "Daniel Lee",
     "Fredrik Ljungdahl",
+    "Nichole Mattera",
     "Matt Penny",
     "Tynan Richards",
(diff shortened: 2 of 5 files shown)


InfoTeddy:
You didn't properly link to #366 in the PR description, so it won't be automatically closed if this PR gets merged. You have to say "Fixes #366" or "Closes #366".

InfoTeddy:
This PR has a regression - if I start editing the room name, and then press Enter, it seems like the game still thinks it's taking input from me even though I'm not, and I end up being unable to press E, use the arrow keys to move across the room, place tiles, or do various other things. However, if I press Esc, I get out of this stuck state and can use the editor properly again.

NicholeMattera:
Good catch! :+1:  Didn't pick up on that as I kept instantly previewing the level. :woman_facepalming: 

I also greatly simplified these fixes. The `getlin` method was already setting the `textmod` property, which could have been checked in main to see if it wasn't set to none. Also for the fullscreen fix it seems like SDL text input already ignores ctrl, and super, so I could simplify this fix with using the already existing `altpressed` variable, and just increasing the scope of it so it could be checked in the text input event.

InfoTeddy:
After testing the new commit, it looks good to me!

flibitijibibo:
Ditto, thanks for the review!

PR #447, opened 2020-08-13
Axe manual state trackers and use SDL_IsTextInputActive()

InfoTeddy:
After looking at pull request #446, I got a bit annoyed that we have *two* variables, `key.textentrymode` and `ed.textentry`, that we rolled ourselves to track the state of something SDL already provides us a function to easily query: `SDL_IsTextInputActive()`. We don't need to have either of these two variables, and we shouldn't.

So that's what I do in this patch. Both variables have been axed in favor of using this function, and I just made a wrapper out of it, named `key.textentry()`.

For bonus points, this gets rid of the ugly `NO_CUSTOM_LEVELS` and `NO_EDITOR` ifdef in main.cpp, since text entry is enabled when entering the script list and disabled when exiting it. This makes the code there easier to read, too.

Furthermore, apparently `key.textentrymode` was initialized to *true* instead of false... for whatever reason. But that's gone now, too.

Now, you'd think there wouldn't be any downside to using `SDL_IsTextInputActive()`. After all, it's a function that SDL itself provides, right?

Wrong. For whatever reason, it seems like text input is active *from the start of the program*, meaning that what would happen is I would go into the editor, and find that I can't move around nor place tiles nor anything else. Then I would press Esc, and then suddenly become able to do those things I wanted to do before.

I have no idea why the above happens, but all I can do is to just insert an `SDL_StopTextInput()` immediately after the `SDL_Init()` in `main.cpp`. Of course, I have to surround it with an `SDL_IsTextInputActive()` check to make sure I don't do anything extraneous by stopping input when it's already stopped.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -153,6 +153,10 @@ int main(int argc, char *argv[])
         SDL_INIT_JOYSTICK |
         SDL_INIT_GAMECONTROLLER
     );
+    if (SDL_IsTextInputActive() == SDL_TRUE)
+    {
+        SDL_StopTextInput();
+    }
 
     NETWORK_init();
 
@@ -617,12 +621,7 @@ void inline fixedloop()
     }
 
     //Mute button
-#if !defined(NO_CUSTOM_LEVELS) && !defined(NO_EDITOR)
-    bool inEditor = ed.textentry || ed.textmod != TEXT_NONE || ed.scripthelppage == 1;
-#else
-    bool inEditor = false;
-#endif
-    if (key.isDown(KEYBOARD_m) && game.mutebutton<=0 && !inEditor)
+    if (key.isDown(KEYBOARD_m) && game.mutebutton<=0 && !key.textentry())
     {
         game.mutebutton = 8;
         if (game.muted)
@@ -639,7 +638,7 @@ void inline fixedloop()
         game.mutebutton--;
     }
 
-    if (key.isDown(KEYBOARD_n) && game.musicmutebutton <= 0 && !inEditor)
+    if (key.isDown(KEYBOARD_n) && game.musicmutebutton <= 0 && !key.textentry())
     {
         game.musicmutebutton = 8;
         game.musicmuted = !game.musicmuted;
--- a/desktop_version/src/KeyPoll.cpp
+++ b/desktop_version/src/KeyPoll.cpp
@@ -38,7 +38,6 @@ KeyPoll::KeyPoll()
 	setSensitivity(2);
 
 	quitProgram = 0;
-	textentrymode=true;
 	keybuffer="";
 	leftbutton=0; rightbutton=0; middlebutton=0;
 	mx=0; my=0;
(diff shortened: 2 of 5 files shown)


desktop_version/src/main.cpp
```diff
@@ -153,6 +153,8 @@ int main(int argc, char *argv[])
         SDL_INIT_JOYSTICK |
         SDL_INIT_GAMECONTROLLER
     );
+    // WTF? Why is text input active from program start?!
+    SDL_StopTextInput();
```
flibitijibibo, line 157: This is invalid behavior - SDL makes it perfectly clear that text input is enabled on targets where a keyboard is always present. It's only disabled for devices that have to use on-screen keyboards, where you would have to explicitly start it to know for sure that you will receive text events.

InfoTeddy: Okay, but if I don't do this, then SDL will tell the game text input is being actively taken when it isn't, leading to being unable to do anything in the editor. Is there another way to work around this that I'm not aware of?

flibitijibibo: `SDL_IsTextInputActive()` should be accurate at this point in the program.

InfoTeddy: Well, that's the thing. When I tested it, it was not. If I don't do this, then going to the editor results in my keypresses doing nothing until I press Esc.

flibitijibibo: I find this hard to believe, here is the implementation:

https://hg.libsdl.org/SDL/file/369fea168258/src/video/SDL_video.c#l3836

SDL_InitEvents will start the event state as disabled...

https://hg.libsdl.org/SDL/file/369fea168258/src/events/SDL_events.c#l419

... but VideoInit will enable it based on the backend's screen keyboard support:

https://hg.libsdl.org/SDL/file/369fea168258/src/video/SDL_video.c#l556

We know the event state will be disabled before the above line since the video subsystem requires that the event subsystem be initialized first:

https://hg.libsdl.org/SDL/file/369fea168258/src/video/SDL_video.c#l479

You will want to make absolutely sure that you are disabling it for the right case, as Start/Stop do a lot more than just enable/disable the event type.

InfoTeddy: I don't really see any part of your explanation where it's clear that the event state is disabled at the end of `SDL_Init()`. Especially since line 556 of `SDL_video.c` (in `SDL_VideoInit()`) does `SDL_StartTextInput()` if `SDL_HasScreenKeyboardSupport()` is false.

Indeed, if I open the game up in GDB and set a breakpoint on the first line of `SDL_StartTextInput()`, `SDL_video.c:3757` (it seems like directly setting a breakpoint on the function name doesn't work, I assume this is because of inlining), that's exactly what happens. `SDL_VideoInit()` ends up calling `SDL_StartTextInput()`.

flibitijibibo: Correct, that is why I expect SDL_IsTextInputActive() to return 1 and not 0. You need to check to see if it's active before stopping input, otherwise you enter potentially undefined behavior. If it's returning 0, you need to find out why this is the case.

InfoTeddy: Fair enough, but then I'm confused on how much we actually disagreed in the first place.

Whatever. I can add the check so that it only disables it if it's enabled, but I'm not sure if you want me to do more than that (judging by the sentence "If it's returning 0, you need to find out why this is the case").

desktop_version/src/main.cpp
```diff
@@ -153,6 +153,11 @@ int main(int argc, char *argv[])
         SDL_INIT_JOYSTICK |
         SDL_INIT_GAMECONTROLLER
     );
+    if (SDL_IsTextInputActive() == SDL_TRUE)
+    {
+        // WTF? Why is text input active from program start?!
```
flibitijibibo, line 158: We know why. This line can be cut now.

InfoTeddy: Removed. Amended and force-pushed.

PR #453, opened 2020-08-13
Don't fade music out when returning to the menu if it's Presenting VVVVVV

### LINKED ISSUE #452: Presenting VVVVVV won't play in the menu if it was being played before quitting
If you have Presenting VVVVVV playing in a level, then you quit to the menu, Presenting VVVVVV will fade out, as all music does when returning to the menu, but it wont start playing again for the menu. This is most likely because `music.currentsong` is set to Presenting VVVVVV when you leave the level, so when `musicclass::play()` tries to play it again for the menu, it sees that `music.currentsong` is already Presenting VVVVVV so it doesn't do anything, to avoid restarting the song which is already playing.
This could be fixed in two ways:
- Setting `music.currentsong` to -1 when you exit a level to make sure Presenting VVVVVV always plays, or
- Not fading out the music when you quit to the menu if Presenting VVVVVV is playing, to make it continue playing on the menu instead of it stopping and starting again.

AllyTally:
## Changes:

If Presenting VVVVVV is playing when you exit to the menu, the song will fade out, and then Presenting VVVVVV won't fade (back) in. This fixes that by avoiding fading the music out if it's already playing Presenting VVVVVV for a smooth transition between the level and the menu. The alternative would have been making it fade out and then back in, which wouldn't really be ideal.

Fixes #452.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2310,7 +2310,9 @@ void mapmenuactionpress()
         //This fixes an apparent frame flicker.
         FillRect(graphics.tempBuffer, 0x000000);
         graphics.fademode = 2;
-        music.fadeout();
+        if (music.currentsong != 6) {
+            music.fadeout();
+        }
         map.nexttowercolour();
         if (!game.glitchrunnermode)
         {


flibitijibibo:
This is fine for now, but it may point to a bug in how fades are done. We may have to adjust fading to require the next song as a parameter, so when the song matches it doesn't bother to fade, fixing this issue and others we haven't seen yet. That's probably a 2.4 kind of thing though.

PR #467, opened 2020-09-06
Remove checks to prevent having both warp lines/BGs in editor

InfoTeddy:
It's perfectly acceptable to have both warp lines and a warping background in the same room. Many levels do this exact thing, I would say at least 30 or so levels, many of them popular and played by many, and this has never caused any issues at all.

All that having both warp lines and warp BG does is make it so the warping of the warping background gets overriden by the warp lines, but make it so the background is still a warp background. So in effect, you can have a warp background without any warping. This is perfectly defined behavior. Except, for whatever reason, it's unintentional, and the editor tries to prevent you from doing it.

Key word being "tries". The code to prevent having both warp types is bugged (at least when you change the warp BG. The check when you place warp lines seems to be solid). It compares the `p1` and `p2` attributes of warp lines to the x-coordinate and y-coordinate of the room, despite `p1` and `p2` having nothing to do with room coordinates. `p1` is the type of the warp line and should be treated as an enum, and `p2` is the offset of the warp line from the top/left of the screen. This results in this check sometimes working if you're unlucky, but never actually working properly most of the time. This means people can first place warp lines, and then change the warp background later, to have both warp lines and a warp background.

Having these checks just further complicates the code, makes it more error-prone, and just inconveniences people when they want to do something that's perfectly fine to do. So it's best if we just remove these checks.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -4572,51 +4572,30 @@ void editorinput()
 
             if(key.keymap[SDLK_w])
             {
-                int j=0, tx=0, ty=0;
-                for(size_t i=0; i<edentity.size(); i++)
+                ed.level[ed.levx+(ed.levy*ed.maxwidth)].warpdir=(ed.level[ed.levx+(ed.levy*ed.maxwidth)].warpdir+1)%4;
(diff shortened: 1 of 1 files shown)


TerryCavanagh:
Makes sense to me! Merging.

PR #471, opened 2020-09-08
Prevent same-frame infinite loops in scripts

InfoTeddy:
It's trivially easy to send the scripting system into an infinite loop on the same frame (i.e. without any script delay in between, i.e. within the same execution of `script.run()`). Just take a look at these two scripts:

    a:
    iftrinkets(0,b)
    #

    b:
    iftrinkets(0,a)
    #

The hashes are to prevent the scripting system from parsing `iftrinkets()` using the internal version instead of the simplified version, because after doing a simplified `iftrinkets()`, the parser will (to oversimplify) execute the last line of the script as internal.

Anyway, sending the game into an infinite loop like this will cause the Not Responding dialog on Windows.

So to prevent this from happening, I've added an execution counter to `scriptclass::run()`. If it gets too high, we're in an infinite loop and so we stop running the script.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -9,6 +9,8 @@
 #include "Music.h"
 #include "UtilityClass.h"
 
+#include <limits.h>
+
 scriptclass::scriptclass()
 {
 	//Start SDL
@@ -76,6 +78,8 @@ void scriptclass::tokenize( const std::string& t )
 
 void scriptclass::run()
 {
+	// This counter here will stop the function when it gets too high
+	short execution_counter = 0;
 	while(running && scriptdelay<=0 && !game.pausescript)
 	{
 		if (position < (int) commands.size())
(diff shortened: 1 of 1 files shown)


AllyTally:
Shouldn't `MAX_EXECUTIONS` be a `#define`?

flibitijibibo:
It's not too big of a deal which one, but if we have a solid explanation for why exactly 10,000 that'd be enough for me.

InfoTeddy:
I initially did something like 2<sup>31</sup> at first, but it turns out incrementing to that value is really slow. I then tried 100,000 and it would make the frame take too long and still appear frozen. So I dropped a zero and it didn't seem to freeze at all.

On second thought, I could probably make it so that it would be 2<sup>15</sup>-1 instead, which is just the amount of positive values in a `short` - it'd be more natural that way.

flibitijibibo:
Yeah, something absurdly large but not fatal works for me - we still want to convey to editors that the script is doing something very bad, but we don't want it to be inescapable for the end user if a bad script ships anyway.

PR #472, opened 2020-09-08
Fix destroy(platforms), implement under different names

InfoTeddy:
`destroy(platforms)` has been bugged since 2.0. The problem with it is that it removes the platform entity, but doesn't remove its block. This results in essentially turning the platorm invisible and stopping it from moving.

This error should be fixed, but some levels (including my own) rely on the invisible platform trick. So instead, the fixed version will be implemented under a different name, `destroy(moving)`.

There's also another problem with `destroy(platforms)`, which is that the name is misleading and it doesn't additionally destroy disappearing platforms. I would also fix this, but in order to not run the risk of breakage, it will have to be implemented under a different name, too. So this will be `destroy(disappear)`. As an added benefit, it's also more granular to have platform-destroying functions under different names than it is to consolidate them under the same name.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -169,9 +169,18 @@ void scriptclass::run()
 					for(size_t edi=0; edi<obj.entities.size(); edi++){
 						if(obj.entities[edi].type==11) removeentity_iter(edi);
 					}
-				}else if(words[1]=="platforms"){
+				}else if(words[1]=="platforms"||words[1]=="moving"){
+					bool fixed=words[1]=="moving";
 					for(size_t edi=0; edi<obj.entities.size(); edi++){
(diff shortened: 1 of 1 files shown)


InfoTeddy:
I'm not sure what happened there - somehow I had unpushed commits from an unpushed branch end up in this branch? Whatever, it's fixed now.

InfoTeddy:
Oh, I think I know/remember what happened. I accidentally rebased onto that unpushed branch, then rebased onto master, then didn't realize that my rebasing introduced extraneous commits that I didn't want. Again though, it's fixed now.

InfoTeddy:
This PR has been moved to #494 because GitHub is stupid and won't let you rename branches without closing a PR.

PR #476, opened 2020-09-10
Fix up bounds checks

InfoTeddy:
The game currently has these problems with its bounds checks, all of which this PR fixes:

- It was sometimes using manually-typed bounds checks instead of the macros, which is inconsistent. Also, some of these manually-typed bounds checks are wrong!
- It was falsely assuming it would be sufficient to check `i > -1` and not the other end, *especially* in script commands, where the lack of checking of the other end could lead to potential out-of-bounds indexing given a fuzzer!
- In the same vein, it was also falsely assuming entity getters that returned a default value of 0 would also be fine to not check, leading to the same situation with certain script commands as above.
- And sometimes it just plain flat-out didn't bounds-check certain things.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1434,7 +1434,7 @@ void entityclass::createentity( float xp, float yp, int t, float vx /*= 0*/, flo
 
         //Check if it's already been collected
         entity.para = vx;
-        if (collect[(int) vx]) return;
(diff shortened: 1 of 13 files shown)


flibitijibibo:
This patchset lgtm, just needs a rebase now.

InfoTeddy:
Rebased and force-pushed.

Issue #379, opened 2020-07-12
Script editor should prevent you from typing pipes and leaving a line with a colon on the end


InfoTeddy:
The pipe (`|`) and colon (`:`) characters are very special, because the custom level XML format separates script lines by pipes and denotes a script name line if it has a colon on the end of it, which marks the beginning of another script. Without changing the format, these characters will cause trouble for users who use them without knowing that they're special.

Users are already semi-discouraged from using pipes in scripts, because in the default font they actually show up as `ä`s (instead of adding Unicode support, looks like Terry did this so Andreas Kämper, a superpatron, would have their name show up properly in the credits. He also did it for the `å` in Magnus Pålsson, the game's composer, turning `~` into it, he turned `{` into `ö` for Andreas Jörgensen, a regular patron, and he turned `}` into `Á` for Ángel Louzao Penalva, another regular patron).

It seems fine to put these pipes in and everything. Until they save the level and re-load it, at which point their pipes will properly be parsed as newlines, and then their script lines will be split.

Users have a higher chance of leaving a colon on the end of a script line, because they might want to write dialogue such as this:
```
say(2)
MC-182897 STATUS:
WORKING AS INTENDED
```
(that Working As Intended status is absolutely wrong and stupid btw, thanks Mojang)

But this time, whenever they close the script, the game will immediately re-parse the colon as a new script and break the script in two, which is much worse than a mere line break. At least, in 2.2. In 2.3 this only happens when you save and re-load the level, because after #277 script parsing is done at level load time, instead of having to lug the parsing boilerplate around every time the game needs to do something with scripts.

Anyways, Ved solves these issues by (1) making a newline instead of typing a pipe whenever the user types a pipe, and (2) putting a space after the colon when the user exits out of the script. Putting a space after is fine because the game only checks if the last character is a colon and doesn't strip whitespace when doing so. VVVVVV should do the same.

Edit: #479 solves the colon issue but the pipe issue remains.

MAO3J1m0Op:
It may be possible to allow typing of the pipe in VVVVVV scripts. An option is an escape sequence, but that may cause problems in a few existing scripts, since there's no characters in the ASCII library that don't have a character in the VVVVVV font or another special meaning. Another option is to transition to using the newline character to represent a new line rather than the pipe.

AllyTally:
That sounds hard to do without breaking backwards compatibility with older levels.

PR #479, opened 2020-09-14
Add space after lines terminated by a colon in scripts

MAO3J1m0Op:
## Changes:

This is a small fix that partially addresses #379 by inserting a space after a colon followed by a new line when a script is saved.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -14,6 +14,7 @@ Contributors
 * Fredrik Ljungdahl (@FredrIQ)
 * Nichole Mattera (@NicholeMattera)
 * Matt Penny (@mwpenny)
+* MAO3J1m0Op (@MAO3J1m0Op)
 * Tynan Richards (@tzann)
 * Elliott Saltar (@eboyblue3)
 * Marvin Scholz (@ePirat)
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -2158,7 +2158,7 @@ bool editorclass::save(std::string& _path)
             scriptString += script_.contents[i]
 
             // Inserts a space if the line ends with a :
-            + (script_.contents[i].back() == ':'? " " : "")
+            + (script_.contents[i].length() != 0 && *script_.contents[i].rbegin() == ':' ? " " : "")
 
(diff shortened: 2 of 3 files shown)


flibitijibibo:
Just need the tickboxes in the OP checked and this should be good to review... CC @InfoTeddy, in case this needs a quick test as well.

flibitijibibo:
Looks like there were some build failures, possibly C++ version compliance issues.

InfoTeddy:
On it, I'll be looking at this, too.

desktop_version/CONTRIBUTORS.txt
```diff
@@ -14,6 +14,7 @@ Contributors
 * Fredrik Ljungdahl (@FredrIQ)
 * Nichole Mattera (@NicholeMattera)
 * Matt Penny (@mwpenny)
+* MAO3J1m0Op (@MAO3J1m0Op)
```
InfoTeddy, line 17: You forgot to add yourself to `githubfriends` in `Credits.h`.

desktop_version/src/editor.cpp
```diff
@@ -2155,7 +2155,12 @@ bool editorclass::save(std::string& _path)
         scriptString += script_.name + ":|";
         for (size_t i = 0; i < script_.contents.size(); i++)
         {
-            scriptString += script_.contents[i] + "|";
+            scriptString += script_.contents[i]
+
+            // Inserts a space if the line ends with a :
+            + (script_.contents[i].length() != 0 && *script_.contents[i].rbegin() == ':' ? " " : "")
+
+            + "|";
```
InfoTeddy, line 2163: This works, but I'd rather have this factored out into a full-fledged if-condition instead of being shoved into a ternary. Also, it should just be `script_.contents[i].length()` instead of having that unnecessary `!= 0` at the end. (That way you don't have to think about the difference between `!= 0` and `> 0`.)

InfoTeddy (changes requested):
It works, but there's a few things I want to address.

InfoTeddy:
Alright, looks good to me! @flibitijibibo, you should squash this on merge, in my opinion.

flibitijibibo:
Looks good to me as well, thanks for checking this out!

PR #483, opened 2020-09-21
Allow the character separating script lines to be customizable

MAO3J1m0Op:
## Changes:

In order to allow the typing of the pipe character to fix the second part of #379, an attribute to the `<script>` tag of a level was added. The attribute is titled `linebreak`, and in order to preserve backwards compatibility, the abscense of this attribute (or a value of `""`) tells the game to still parse the level using a pipe as a line separator. However, if the attribute is not present (or, again, has a value of `""`), the save will be written with a `linebreak` value of `\n`.

This PR does change the format of the custom level XML file, but the code handles both the old and new formats.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -1960,11 +1960,11 @@ bool editorclass::load(std::string& _path)
         if (pKey == "script")
         {
             // load in the linebreak attribute
-            linebreak = "default";
+            linebreak = "";
             pElem->QueryStringAttribute("linebreak", &linebreak);
             char splitvalue = linebreak[0];
-            // "default" is the only value where splitvalue and linebreak are different
-            if (std::strcmp(linebreak, "default") == 0)
+            // "" is the only value where splitvalue and linebreak are different
+            if (std::strcmp(linebreak, "") == 0)
             {
                 splitvalue = '|'; // To keep compatibility with older levels
                 linebreak = "\n"; // So this conversion only happens once
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -1959,10 +1959,21 @@ bool editorclass::load(std::string& _path)
 
         if (pKey == "script")
         {
+            // load in the linebreak attribute
+            linebreak = "default";
+            pElem->QueryStringAttribute("linebreak", &linebreak);
+            char splitvalue = linebreak[0];
+            // "default" is the only value where splitvalue and linebreak are different
+            if (std::strcmp(linebreak, "default") == 0)
+            {
+                splitvalue = '|'; // To keep compatibility with older levels
+                linebreak = "\n"; // So this conversion only happens once
+            }
+
             std::string TextString = (pText);
             if(TextString.length())
             {
-                std::vector<std::string> values = split(TextString,'|');
+                std::vector<std::string> values = split(TextString, splitvalue);
                 script.clearcustom();
                 Script script_;
                 bool headerfound = false;
@@ -2152,7 +2163,7 @@ bool editorclass::save(std::string& _path)
     {
         Script& script_ = script.customscripts[i];
 
-        scriptString += script_.name + ":|";
+        scriptString += script_.name + ":" + linebreak[0];
         for (size_t i = 0; i < script_.contents.size(); i++)
         {
             scriptString += script_.contents[i];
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Since this changes behavior it may have to wait until after 2.3, meaning the issue in question will have to be demoted accordingly.

desktop_version/src/editor.cpp
```diff
@@ -1959,10 +1959,21 @@ bool editorclass::load(std::string& _path)
 
         if (pKey == "script")
         {
+            // load in the linebreak attribute
+            linebreak = "";
+            pElem->QueryStringAttribute("linebreak", &linebreak);
+            char splitvalue = linebreak[0];
```
InfoTeddy, line 1965: Why only use the first character of `linebreak`?

desktop_version/src/editor.cpp
```diff
@@ -1959,10 +1959,21 @@ bool editorclass::load(std::string& _path)
...
+            // "" is the only value where splitvalue and linebreak are different
+            if (std::strcmp(linebreak, "") == 0)
```
InfoTeddy, line 1967: 1. We prefer `SDL_strcmp()` instead of `std::strcmp()`.
2. Regardless, this check can be simplified to `linebreak[0] == '\0'`.

desktop_version/src/editor.h
```diff
@@ -113,6 +113,8 @@ class editorclass{
   std::vector<std::string> directoryList;
   std::vector<LevelMetaData> ListOfMetaData;
 
+  const char *linebreak;
```
InfoTeddy, line 116: This attribute should be initialized in `editorclass::reset()`, where all the other attributes of `editorclass` are initialized.

desktop_version/src/editor.cpp
```diff
@@ -1959,10 +1959,21 @@ bool editorclass::load(std::string& _path)
...
+            {
+                splitvalue = '|'; // To keep compatibility with older levels
+                linebreak = "\n"; // So this conversion only happens once
```
InfoTeddy, line 1970: I'm not sure if existing levels should be automatically converted to using newlines. Certainly newly-created levels should be using newlines, but there could be a level out there retexturing newline characters in `font.png` and using them in dialogue. I know for certain I've done it in the title of one of my levels, Alias Space: Checkpoint Abuse.

desktop_version/src/editor.cpp
```diff
@@ -2152,7 +2163,7 @@ bool editorclass::save(std::string& _path)
     {
         Script& script_ = script.customscripts[i];
 
-        scriptString += script_.name + ":|";
+        scriptString += script_.name + ":" + linebreak[0];
```
InfoTeddy, line 2166: See my other comment - if the user creates a new level and saves it, `linebreak` will be uninitialized here! (The best-case scenario is that they load a level beforehand, in which case it will be properly initialized, and then creating a new level will just carry over the `linebreak` from the previous level. Which is still bad, but at least it's not undefined behavior.)

InfoTeddy (changes requested):
I agree with @flibitijibibo that something like this should probably wait for 2.4. However, I feel like #379 should just be fixed by doing what I suggested (preventing you from typing pipes), instead of having that issue demoted.

desktop_version/src/editor.cpp
MAO3J1m0Op: I wanted to use the already existing call to `split`, which took a `char`.

InfoTeddy: You could make the main `split()` take in an `std::string`, and keep the old `char` interface around by using function overloading.

InfoTeddy:
Seems like nothing has changed since my review last year - if you want to fix those and give this PR a rebase, it'd be good to go by me.

Dav999-v:
It's not too late to say this: on current thought, this PR breaks forward compatibility a bit too much in my opinion.

There's not much wrong with _allowing_ the newline character to be customized if the level maker wants, but I disagree with "Certainly newly-created levels should be using newlines". It would mean any new level is immediately, and unexpectedly, broken by default on older versions, for not much practical gain - I think it's rare for anyone to care about being able to use `|` characters in their scripts, or to have problems with it. Plus, if you actually want to use a pipe character, you can just retexture a character in the font, especially with the possibility to make Unicode fonts now.

PR #485, opened 2020-09-22
Split flip logic from player input logic

### LINKED ISSUE #484: Apply flip momentum to duplicate player entities
The player gets a boost of momentum upon flipping, in addition to the gravity change. However, this is not applied to additional players created with `createentity(0,x,y)`. It's debatable whether this is a bug or a feature, but I believe it should be considered:
* The current behavior is, at the very least, unexpected
* #482 was merged, showing that improving support for duplicate player entities is, at the very least, not a *non*-goal
* I don't believe any levels currently released are making use of duplicate player entities
* Outside of contrived scenarios, I doubt this would prevent any new levels using duplicate player entities from being played in 2.2

leo60228:
Closes #484

## Changes:
Flipping only applies momentum to the player entity currently being processed. This normally wouldn't be a problem. However, flipping involves global state, and only one flip can occur per frame. This means that additional player entities don't get this boost of momentum, which feels somewhat unnatural during gameplay.
This commit fixes this by splitting flip logic out of the loop over player entities, and applying the flip momentum to all player entities.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -1909,95 +1909,95 @@ void gameinput()
                 }
             }
         }
+    }
 
-        if (game.press_left)
+    if (game.press_left)
+    {
+        game.tapleft++;
+    }
+    else
+    {
+        if (game.tapleft <= 4 && game.tapleft > 0)
         {
-            game.tapleft++;
-        }
-        else
-        {
-            if (game.tapleft <= 4 && game.tapleft > 0)
+            for (size_t ie = 0; ie < obj.entities.size(); ++ie)
             {
-                for (size_t ie = 0; ie < obj.entities.size(); ++ie)
+                if (obj.entities[ie].rule == 0)
                 {
-                    if (obj.entities[ie].rule == 0)
+                    if (obj.entities[ie].vx < 0.0f)
                     {
-                        if (obj.entities[ie].vx < 0.0f)
-                        {
-                            obj.entities[ie].vx = 0.0f;
-                        }
+                        obj.entities[ie].vx = 0.0f;
                     }
                 }
             }
-            game.tapleft = 0;
         }
-        if (game.press_right)
+        game.tapleft = 0;
+    }
+    if (game.press_right)
+    {
+        game.tapright++;
+    }
+    else
+    {
+        if (game.tapright <= 4 && game.tapright > 0)
         {
-            game.tapright++;
-        }
-        else
-        {
-            if (game.tapright <= 4 && game.tapright > 0)
+            for (size_t ie = 0; ie < obj.entities.size(); ++ie)
             {
-                for (size_t ie = 0; ie < obj.entities.size(); ++ie)
+                if (obj.entities[ie].rule == 0)
                 {
-                    if (obj.entities[ie].rule == 0)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Do we have a before/after sample for this? It makes sense to me (and I'd be interested in knowing of any example that makes use of a behavior this specific) but if it's not noticeable except in a frame-by-frame scenario it may be better to put a big note that says "yup, this is busted alright" to prevent the admittedly narrow case of breaking a level that might require it.

leo60228:
This is a video before the change: https://youtu.be/JkW8Y0hXWk4. First, I mash the right arrow key to demonstrate the second commit, and then I mash the up arrow key to demonstrate the first commit. The top player entity was created by a script, with the bottom being the original.
After these changes, the top half is now the same as the bottom half. I didn't think it was necessary to record another video to demonstrate this.

flibitijibibo:
Ahh okay, so that's a pretty clear difference. If nobody making levels has any objections this seems like a fine fix to me, since it's more likely designers will want all the entities to be properly sync'd anyway.

desktop_version/src/Input.cpp
```diff
@@ -1897,83 +1897,106 @@ void gameinput()
...
+        {
+            game.jumppressed = 5;
+            game.jumpheld = true;
+        }
 
-                if (game.jumppressed > 0)
+        if (game.jumppressed > 0)
+        {
+            game.jumppressed--;
+            if (obj.entities[obj.getplayer()].onground>0 && game.gravitycontrol == 0)
```
InfoTeddy, line 1971: Why the `obj.getplayer()` here? This will only check if the first player entity is on the ground, and none of the other ones.

I tested it just now, and this change breaks mid-air flip levels (such as [this one](https://distractionware.com/forum/index.php?topic=3556.0)), where a duplicate player entity is put inside a box, thus making their `onground` and `onroof` attributes almost always 2, letting the first player entity flip in mid-air. But if you only check the `onground` and `onroof` of the first player entity, then of course they will no longer be able to flip in mid-air.

Also, I'd prefer it if you did an `INBOUNDS()` check on `obj.getplayer()` to uphold the norm of defensive programming.

InfoTeddy (changes requested):
> The game will freeze the player immediately if they release a directional button within 3 frames of pressing it.

It's actually 4 frames. You need to hold an input for 5 frames before releasing it will apply deceleration.

desktop_version/src/Input.cpp
```diff
@@ -1908,6 +1910,12 @@ void gameinput()
                     obj.entities[ie].dir = 1;
                 }
             }
+
+            if (obj.entities[ie].onground > 0) {
+              any_onground = true;
+            } else if (obj.entities[ie].onroof > 0) {
+              any_onroof = true;
+            }
```
InfoTeddy, line 1918: What's up with the 2-space indentation and same-line braces?

InfoTeddy (changes requested):

flibitijibibo:
Modified this to be a draft (so nobody merges it), but also marked it as a 2.4 feature - this may end up being related when we redo the input system to use Action Sets instead of hardcoded keys/buttons.

Issue #373, opened 2020-07-11
Add XML forwards compatibility


InfoTeddy:
There's a semi-annoying issue with the game where if you downgrade to 2.2 and 2.2 writes its settings file, all your 2.3 settings are lost. This is because 2.2 writes only what 2.2 has and nothing else. It has no forwards compatibility. (The same issue also happens with 2.0, i.e. your 2.2 controller mappings and stuff will be lost if 2.0 writes its settings file.)

While there's basically no reason to downgrade to 2.2 after 2.3 is released (if you want credits warp or other glitches, just use glitchrunner mode), and while we can't really rectify this for 2.2, we could stand to introduce XML forwards compatibility in 2.3, which would be applied to all files, not just `unlock.vvv`. And this would mean preserving attributes that 2.3 doesn't have in XML elements that 2.3 does have, as well as preserving XML elements that 2.3 doesn't have, and so on.

Also, as @Dav999-v suggested in the comments below, we should also move all settings to a new `settings.vvv` XML file.

Dav999-v:
Just thinking out loud, but it's a bit odd that settings can be found in the file called "unlock" anyway. Maybe it'd be a good idea to move all configuration to a separate XML file (like `settings.vvv`) where unknown elements are preserved from its inception, and leave `unlock.vvv` for game stuff like unlocks and records. If the new settings file doesn't exist, but `unlock.vvv` does, then it can copy settings from `unlock.vvv` to the new file.

Opening 2.2 won't suddenly destroy new 2.3 options anymore. After opening 2.3, settings common between 2.2 and 2.3 will be configured separately, but that's arguably much better than suddenly losing settings because you wanted to try out something in 2.2 or opened the wrong version accidentally.

InfoTeddy:
I like that `settings.vvv` idea, too, and we should do it.

The main thing though is that in order to not duplicate the file spec a *third* time, we should somehow abstract each element and attribute into an abstract static constant data structure that we can easily read from to load and save data. And we'd still have to do that for moving the configuration settings into `settings.vvv`

Alternatively, I was thinking it was maybe possible to just open an XML document and update it instead of writing an entire new XML document to the file. It seems simple that you should be able to do that, right? I haven't looked into this further but that's maybe a starting point to adding XML forwards compatibility without having to duplicate file specs a third time, or having to abstract and refactor the file specs.

Dav999-v:
[...]

PR #487, opened 2020-09-25
Add XML forwards compatibility

InfoTeddy:
This PR adds XML forwards compatibility, meaning extra XML data will be preserved if the game encounters it while writing an XML file. It assumes that the XML data is from some future version of the game and tries very hard not to throw it away. It does this by loading in the current XML when it saves it, and instead of creating a new XML document from scratch, it takes the existing document and updates it in place.

There are exceptions, of course. The most notable ones are that unknown attributes of `<edLevelClass>` and `<edEntity>` are *not* preserved. This is because they would take extra work to support (you can't just blindly keep the attributes for the same indice between reading and writing, in the meantime the user could have changed the order around or deleted edentities, so then you have to keep the entire XML object attached to the internal edentity object, which just opens up a whole can of worms), and also because the level format, where every single thing is shoved into one file, instead of being made up of multiple files, is terrible, and should be abandoned in the future, and when that happens there won't be any need for preserving attributes on objects like these.

This PR only closes one part of #373.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/XMLUtils.cpp
+++ b/desktop_version/src/XMLUtils.cpp
@@ -1,0 +1,142 @@
+#include <SDL.h>
+#include <tinyxml2.h>
+
+namespace xml
+{
+
+// Helper functions for these utils (not exported)
+
+static inline tinyxml2::XMLDocument& get_document(tinyxml2::XMLNode* parent)
+{
+    return *(parent->GetDocument());
+}
+
+
+// EXPORTED FUNCTIONS
+
+
+// Create a new element if it doesn't exist. Returns the element.
+tinyxml2::XMLElement* update_element(tinyxml2::XMLNode* parent, const char* name)
+{
+    if (parent == NULL)
+    {
+        return NULL;
+    }
+
+    tinyxml2::XMLDocument& doc = get_document(parent);
+
+    tinyxml2::XMLElement* element = parent->FirstChildElement(name);
+    if (element == NULL)
+    {
+        // It doesn't exist, so create a new one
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Was this close intentional? Looks like it was just a force-push...

InfoTeddy:
Oh, no, it wasn't intentional. I have no idea why my force-push would close this.

InfoTeddy:
@flibitijibibo Ready for merge. Didn't think that I wouldn't work on this for so long, but ah well...

desktop_version/src/XMLUtils.cpp
```diff
@@ -0,0 +1,142 @@
+#include <SDL.h>
+#include <tinyxml2.h>
+
+namespace xml
+{
+
+// Helper functions for these utils (not exported)
+
+tinyxml2::XMLDocument& get_document(tinyxml2::XMLNode* parent)
```
flibitijibibo, line 9: May be worth setting this to static (maybe inline too?).

InfoTeddy: What would be the reasoning behind that?

flibitijibibo: Since it's not meant to be exported it would reduce the scope and prevent it from being exported unintentionally, while almost making it easier for the compiler to inline it.

InfoTeddy: Makes sense to me. Added, rebased, and force-pushed.

PR #498, opened 2020-09-28
Clean up and prevent unnecessary qualifiers to self

InfoTeddy:
By "unnecessary qualifiers to self", I mean something like using the `game.` qualifier for a variable on the `Game` class when you're inside a function on the `Game` class itself. This patch is to enforce consistency as most of the code doesn't have these unnecessary qualifiers.

To prevent further unnecessary qualifiers to self, I made it so the extern in each header file can be omitted by using a define. That way, if someone writes something referring to `game.` on a `Game` function, there will be a compile error.

However, if you really need to have a reference to the global name, and you're within the same `.cpp` file as the implementation of that object, you can just do the extern at the function-level. A good example of this is `editorinput()`/`editorrender()`/`editorlogic()` in `editor.cpp`. In my opinion, they should probably be split off into their own separate file because `editor.cpp` is getting way too big, but this will do for now.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1,3 +1,4 @@
+#define OBJ_DEFINITION
 #include "Entity.h"
 
 #include <SDL.h>
--- a/desktop_version/src/Entity.h
+++ b/desktop_version/src/Entity.h
@@ -10,8 +10,22 @@
 #include "BlockV.h"
 #include "Game.h"
 
-#define removeentity_iter(index) { if (obj.removeentity(index)) index--; }
-#define removeblock_iter(index) { obj.removeblock(index); index--; }
+#define removeentity_iter(index) \
+    do \
+    { \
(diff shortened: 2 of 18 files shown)


flibitijibibo:
The change makes sense (and is probably a lot safer, tbh), but i wonder if replacing `NO_GAME_EXTERN` with `GAME_DEFINITION` (for example) would make it clearer - no other build unit would define this other than Game.cpp, so any file _not_ defining the Game class implementation would use the "default" behavior of forward-declaring the variable.

Not a serious issue or anything, but when I saw the defines at the top of the cpp files it reminded me of the [stb](https://github.com/nothings/stb)-style `#ifdef IMPLEMENTATION` checks in the header libraries.

InfoTeddy:
Alright, done. The defines are now e.g. `GAME_DEFINITION`. And yeah, this reminded me of STB, too.

PR #500, opened 2020-09-28
Remove duplicate onroof and onground assignments for vertically-moving platforms

InfoTeddy:
The `onroof`/`onground` attributes are used to determine if the player is standing on a surface and is eligible to flip. Most notably, it is an integer and not a boolean, and it starts at 2, giving the player 2 frames to edge-flip, i.e. they can still flip 2 frames after walking off an edge.

However, these attributes are unnecessarily reassigned in `movingplatformfix()` (which is the function that deals exclusively with vertically-moving platforms; horizontal moving platforms get their own `hormovingplatformfix()`). Whoever wrote this misunderstood what `onroof`/`onground` meant; they thought that they were booleans, and so set them to true, instead of the proper value of 2. This ends up setting `onroof`/`onground` to 1 instead of 2, causing a discrepancy with vertical moving platforms and the rest of the surfaces in the game.

The bigger mistake here is duplicating code that never needed to be duplicated. The `onroof`/`onground` assignment in `gamelogic()` works perfectly fine for vertical moving platforms. Indeed, after testing it with libTAS, I can confirm that removing the duplicate assignments restores being able to edge-flip off of moving platforms with 2 frames of leeway, instead of only 1 frame. It also doesn't change how long it takes for the `onroof`/`onground` to get set when the player is recognized as standing on a vertically-moving platform, either.

And so, it's better to not duplicate this code, because when you duplicate it you run the risk of making a mistake, as I just demonstrated.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -4474,13 +4474,11 @@ void entityclass::movingplatformfix( int t, int j )
                 {
                     entities[j].yp = entities[t].yp + entities[t].h;
                     entities[j].vy = 0;
-                    entities[j].onroof = true;
                 }
                 else
                 {
                     entities[j].yp = entities[t].yp - entities[j].h-entities[j].cy;
                     entities[j].vy = 0;
-                    entities[j].onground = true;
                 }
             }
             else


tzann:
While I have no issue with making the game more internally consistent, I do actually oppose this change. It is an inconsequential change in almost all cases, especially for players who play the game casually. The only room in the main game that _requires_ players to flip between vertically moving platforms is **Down Under**, so the change barely affects anything at all. The argument can be made that there may be custom levels where this change is beneficial, but I would argue that all custom levels so far have been designed with the current game mechanics in mind.

My main argument against this is that changing even tiny mechanisms of the game such as this strongly affects speedrunners. While it seems that adding one frame of leeway to flipping off of a vertically moving platform barely affects anything, there's at least 2 examples of where this would make a substantial difference in a speedrun.

In the room **Stop and Reflect**, a previously impossible strategy is now possible, when coming from **Trench Warfare**. This would affect any speedrun category where the **Trench Warfare** trinket needs to be collected, meaning a faster time is possible on the new version.
Any strategy involving flipping between the platforms in the room **Gordian Knot** now has twice the timing window. The fastest strategy actually requires two flips on the last frame possible, meaning each flip has a 1/30 second window to time. After this change, these windows become twice as large, meaning the strategy becomes substantially easier to execute.

Why is making it possible to have faster times in the speedrun significant? For most games, when a new, faster version is released, the leaderboards can be split into multiple different leaderboards, one for each game version. This would mean that the 'main' leaderboard would suddenly shift from having hundreds of runs to having almost none. Because the VVVVVV speedrunning community is quite old, many of the (top) times were set years ago by runners that do not play the game anymore, meaning it would take a long time before the leaderboards fill up again.
Alternatively, the leaderboard can be maintained, with older runs being at a significant disadvantage. I also believe this is harmful, because it means that previously 'good' times are now much easier to achieve, devaluing them significantly.
I believe both solutions are harmful to the speedrunning scene, especially when the change is so tiny.

If the game were brand new, I would never oppose this change. I just believe that the change is so insignificant that it will be unnoticeable to almost everyone, with the exception of speedrunners, who are negatively impacted.

I have discussed this with @InfoTeddy already, but we were unable to settle our differences, which is why I'm bringing it up here. They also mentioned it may just be up to @TerryCavanagh to settle this type of matter, I don't know.

If necessary I can provide footage of the faster strategy in "Stop and Reflect" that is enabled by this change, or the strategy in "Gordian Knot" that (on v2.2) requires two frame-perfect flips, which would be made significantly easier.

flibitijibibo:
We can reintroduce this with the speedrun flag if it helps - just need to add a bool to those spots (and document why it's there) and we can include it in 2.3.

InfoTeddy:
As I will say again, it seems like the true disagreement between @tzann and I is that I think there's nothing wrong if previously unachievable times are now achievable. If I was a speedrunner, I would say "Hey, new strats!" instead of focusing on the part where previous times are devalued; I don't see how this would negatively impact speedrunning. Bringing up that you put in a lot of time and effort previously is just you committing the sunk cost fallacy.

What confuses me, really, is the part where Tynan said in our earlier discussion that if there was a good reason for this change, he wouldn't oppose it. His argument goes: There is no good reason for this change, and it also negatively impacts speedrunning by adding *new* things, therefore it shouldn't happen; however this brings in the implicit concession that if there *was* a good reason for this change, then it would be okay even if it negatively impacted speedrunning...?

Because it seems weird to me that *if* an addition was bad for speedrunning (which, I disagree with), having a valid reason for the addition would end up negating all of the bad-for-speedrunning that it causes. To me, I don't see how all of the bad-for-speedrunning that you say will happen, would end up getting negated by one single valid good reason. Otherwise, I would point out other 2.3 changes and ask why you don't have a problem with those.

Dav999-v:
I don't have a really strong opinion on this, I'm inclined to say mechanics should remain consistent if possible (who knows how insignificant changes will inexplicably "feel" different for players used to the old version - on the other hand, recently when trying the latest version of 2.3 I could swear something about Viridian's position when touching spikes had changed when it hadn't). But I think eventually minor differences because of code quality improvements are bound to happen either way.

I'm more worried about levels suddenly being much harder in specific circumstances (like how some levels were apparently [incompleteable](https://distractionware.com/forum/index.php?topic=324.msg19447#msg19447) specifically on mobile and 3DS versions) but if this change adds a frame of leeway instead of removing it, that doesn't sound like it would happen.

tzann:
@flibitijibibo I'm not sure what the current state of things is regarding a speedrun flag. I do think it would be a bit odd to have the main game's movement behave differently from the speedrun mode though. But as I said, it may well already be the case and I just don't know it.

It's entirely possible that @InfoTeddy and I won't ever agree on this issue, which is fine. I just think this change should be considered more carefully than it has been, because it does have more significance than just fixing a small oversight in the code.

tzann:
@InfoTeddy it's easy to say what you would think/do if you were a speedrunner, but you're not. I can ask around in the speedrunning community to see what other people think, but I would be quite surprised if most people had a differing opinion to mine.

`however this brings in the implicit concession that if there was a good reason for this change, then it would be okay even if it negatively impacted speedrunning...?`
Yes. If there were a good reason for the change, such as it causing the game to softlock or crash, or just have some other behaviour that damages the user experience I wouldn't have a problem. But it seems to me this change was added for no other purpose than that it seems that the behaviour wasn't originally intentional, not because it significantly improves the gameplay experience.
Had there been a comment in the code saying "it is intentional that the leeway to flip after walking off vertically moving platforms is smaller", then I assume you wouldn't have changed it.

`To me, I don't see how all of the bad-for-speedrunning that you say will happen, would end up getting negated by one single valid good reason.`
I don't understand this argument. Usually, you'd weigh the pros and cons when making a decision. If a good reason outweighs the negative effect on the speedrun community, then the change is worth making.

In general I'm just opposed to gameplay behaviour changes, not changes in general. I haven't read through every pull request, so I don't know every single change made in 2.3. But as far as I know, this is the only one that changes Viridian's movement behaviour. If I'm wrong, feel free to point the other changes out to me.
Under this assumption, I don't see why this specific change should be handled differently from any other quirky behaviour the game has. I'm sure there are many other behaviours that are not intentional, but are still left as is. If it were decided that all unintentional quirks (that can be found) should be fixed, that's fine as well. I just don't see what makes this case different from all the others.

I don't think it's worth arguing about whether you think the change is bad for speedrunning. What matters, in the end, is what most speedrunners think about it. As an experienced speedrunner, I've made my opinion clear. I would imagine that most speedrunners either don't care or are opposed to gameplay changes, but I'm not going to pretend that I represent everyone from the entire community.

I think it would be helpful if you could explain what the _benefits_ of this change are, and why they outweigh the importance of leaving gameplay behaviours the same. I personally just don't think that such a tiny change really has a tangible positive effect on the player experience, which is why I oppose it. Changing gameplay behaviour _does_ have consequences for the speedrunning side of things. So, I believe the arguments against this specific change outweigh the benefits.
I don't mean to say that I think the code should be left as is, with duplicated behaviour. I just mean that the behaviour should be left as is.

I do have a feeling this is something we won't agree on though, so it's probably best to just leave it at this point, and see what other people think.

flibitijibibo:
The speedrun flag has a pretty large amount of changes meant for backward compatibility in exchange for intentionally replicating bugs or inconsistent (but still at least reproducible) behavior. This seems to fall in that category so it would be something I would merge if someone wrote it.

InfoTeddy:
For anyone reading in the future, the disagreement was resolved on Tzann's side after he did more research and found there are more scenarios in 2.2 where Viridian has 2 frames of edge flipping on a vertically moving platform, and then agreed that it's good to fix this inconsistency. That doesn't fully answer what should be done to accommodate 2.3 on the leaderboards, but that's a separate discussion that shouldn't take place here.

PR #507, opened 2020-10-11
Fix "name lookup of ‘i’ changed" warning in editor.cpp

Dav999-v:
## Changes:

I got this warning when compiling VVVVVV again after a while:
```
/home/david/g/github_vvvvvv/dav999fork/VVVVVV/desktop_version/src/editor.cpp: In function ‘void editorrender()’:
/home/david/g/github_vvvvvv/dav999fork/VVVVVV/desktop_version/src/editor.cpp:2697:29: warning: name lookup of ‘i’ changed
                 if(edentity[i].p1<=4)
                             ^
/home/david/g/github_vvvvvv/dav999fork/VVVVVV/desktop_version/src/editor.cpp:2656:14: warning:   matches this ‘i’ under ISO standard rules
     for (int i = edentity.size() - 1; i >= 0; i--)
              ^
/home/david/g/github_vvvvvv/dav999fork/VVVVVV/desktop_version/src/editor.cpp:2691:26: warning:   matches this ‘i’ under old rules
                 for (int i = 0; i < 4; i++) {
                          ^
```

It points out there are two nested for loops both defining `i` in what looks to be the code that draws entities in the editor. I thought it was kind of a worrying warning because it implies that whenever the game is compiled by a compiler using the "old rules" for name lookups, the ID of the entity that is currently being checked would silently be overwritten by `4` somewhere in the loop. To prevent this, I just renamed the inner instances of `i` to `j`.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -2688,7 +2688,7 @@ void editorrender()
                 drawRect = graphics.tiles_rect;
                 drawRect.x += tpoint.x;
                 drawRect.y += tpoint.y;
-                for (int i = 0; i < 4; i++) {
+                for (int j = 0; j < 4; j++) {
                     if (custom_gray) BlitSurfaceTinted(graphics.entcolours[obj.customplatformtile],NULL, graphics.backBuffer, &drawRect, gray_ct);
                     else BlitSurfaceStandard(graphics.entcolours[obj.customplatformtile],NULL, graphics.backBuffer, &drawRect);
                     drawRect.x += 8;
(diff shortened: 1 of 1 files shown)


InfoTeddy:
If you run the codebase through `cppcheck` you get a ton of warnings about variables in an inner scope "shadowing" some variable in an outer scope. But it looks like the compiler specifically warned about this instance as opposed to all the others because the inner `i` is defined in the first statement of a for-loop.

InfoTeddy:
Also, how did you get this error? I don't get it with my GCC 9.3.0.

Dav999-v:
I have GCC 7.5.0, all I did was compile my localization branch after merging `upstream/master` into it. Checking out the branch for this PR (based on `upstream/master`, so lots of files needed recompilation), I believe this was again the only message to pop up.

Shadowing variables doesn't necessarily cause ambiguity, it's just in this case the outer variable was used after (and outside the scope of) the inner for-loop.

PR #510, opened 2020-11-01
Reload window icon when mounting and unmounting assets

InfoTeddy:
The window icon is simply another asset that can be customized by level makers. And in fact, one of my levels changes the window icon. It's simply named `VVVVVV.png`, but it doesn't sit in the `graphics` folder, rather it sits in the root VVVVVV directory.

I noticed that this asset was missed when per-level assets loading was added, so I decided to add it in.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -62,27 +62,7 @@ void Screen::init(
 	);
 	SDL_SetWindowTitle(m_window, "VVVVVV");
 
-	unsigned char *fileIn = NULL;
-	size_t length = 0;
-	unsigned char *data;
-	unsigned int width, height;
-	FILESYSTEM_loadFileToMemory("VVVVVV.png", &fileIn, &length);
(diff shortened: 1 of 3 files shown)


desktop_version/src/Screen.cpp
```diff
@@ -109,6 +89,31 @@ void Screen::init(
...
+		24,
+		width * 3,
+		0x000000FF,
+		0x0000FF00,
+		0x00FF0000,
+		0x00000000
+	);
+	SDL_SetWindowIcon(m_window, icon);
+	SDL_FreeSurface(icon);
+	SDL_free(data);
```
flibitijibibo, line 114: In this one case it does have to be free - we would need to macro the malloc call in lodepng to get this working with SDL_stdlib.

InfoTeddy: Fixed, rebased, and forced-pushed.

PR #513, opened 2020-11-01
Fix crewmates being drawn behind other entities

### LINKED ISSUE #426: Supercrewmates are rendered behind other entities
A supercrewmate behind a checkpoint

This is weird, given that Viridian is drawn in front of entities, and thus it looks unusual for other crewmates to not do the same thing.

This happens because the checkpoint gets created first and then the supercrewmate gets created afterwards. Entities are drawn backwards, so the checkpoint ends up being drawn on top.

InfoTeddy:
This fixes the draw order by drawing all other entities first, before then drawing all humanoids[1] after, including the player afterwards.

This is actually a regression fix from #191. When I was testing this, I was thinking about where get a crewmate in front of another entity in the main game, other than the checkpoints in Intermission 1. And then I thought about the teleporters, because I remember the pre-Deep Space cutscene in Dimension Open looking funny because Vita ended up being behind the teleporter. (Actually, a lot of the cutscenes of Dimension Open look funny because of crewmates standing behind terminals.)

So then I tried to get crewmates in front of teleporters. It actually turns out that you can't do it for most of them... except for Verdigris. And then that's what I realized why there was an oddity in `WarpClass.cpp` when I was removing the `active` system from the game - for some reason, the game put a hole in `obj.entities` between the teleporter and the player when loading the room Murdering Twinmaker. In a violation of Chesterton's Fence (the principle that you should understand something before removing it), I shrugged it off and decided "there's no way to support having holes with my new system, and having holes is probably bad anyway, so I'm going to remove this and move on". The fact that there wasn't any comments clarifying the mysterious code didn't help (but, this *was* 2.2 code after all; have you *seen* 2.2 code?!).

And it turns out that this maneuver was done so Verdigris would fill that hole when he got created, and Verdigris being first before the teleporter would mean he would be drawn in front of the teleporter, instead of being behind it. So ever since b1b1474b7bbc3ceddea24f689a7ddb050cfe4490 got merged, there has actually been a regression from 2.2 where Verdigris got drawn behind the teleporter in Murdering Twinmaker, instead of properly being in front of it like in 2.2 and previous.

This patch fixes that regression, but it actually properly fixes it instead of hacking around with the `active` system.

Closes #426.

[1]: I'm going to go on a rant here, so hear me out. It's not explicitly stated that the characters in VVVVVV are human. So, given this information, what do we call them? Well, the VVVVVV community (at least the custom levels one, I don't think the speedrunning community does this or is preoccupied with lore in the first place) decided to call them "villis", because of the roomname "The Villi People" - which is only one blunder in a series of awful headcanons based off of the assumption that the intent of Bennett Foddy (who named the roomnames) was to decree some sort of lore to the game. Another one being "Verdigris can't flip" because of "Green Dudes Can't Flip". Then an OC (original character) got named based off of "The Voon Show" too. And  so on and so forth.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -1591,26 +1591,18 @@ void Graphics::drawentities()
 
 #if !defined(NO_CUSTOM_LEVELS)
     // Special case for gray Warp Zone tileset!
-    int room = game.roomx-100 + (game.roomy-100) * ed.maxwidth;
-    bool custom_gray = INBOUNDS_ARR(room, ed.level)
+    const int room = game.roomx-100 + (game.roomy-100) * ed.maxwidth;
+    const bool custom_gray = INBOUNDS_ARR(room, ed.level)
     && ed.level[room].tileset == 3 && ed.level[room].tilecol == 6;
 #else
-    bool custom_gray = false;
+    const bool custom_gray = false;
 #endif
 
     std::vector<SDL_Surface*>& tilesvec = (map.custommode && !map.finalmode) ? entcolours : tiles;
 
     std::vector<SDL_Surface*>& spritesvec = flipmode ? flipsprites : sprites;
 
-    int yoff;
-    if (map.towermode)
-    {
-        yoff = lerp(map.oldypos, map.ypos);
-    }
-    else
-    {
-        yoff = 0;
-    }
+    const int yoff = map.towermode ? lerp(map.oldypos, map.ypos) : 0;
 
     for (int i = obj.entities.size() - 1; i >= 0; i--)
     {
@@ -1619,8 +1611,8 @@ void Graphics::drawentities()
             continue;
         }
 
-        int xp = lerp(obj.entities[i].lerpoldxp, obj.entities[i].xp);
-        int yp = lerp(obj.entities[i].lerpoldyp, obj.entities[i].yp);
+        const int xp = lerp(obj.entities[i].lerpoldxp, obj.entities[i].xp);
+        const int yp = lerp(obj.entities[i].lerpoldyp, obj.entities[i].yp);
 
         switch (obj.entities[i].size)
         {
@@ -1669,7 +1661,7 @@ void Graphics::drawentities()
                 wrappedPoint.y -= 230;
             }
 
-            bool isInWrappingAreaOfTower = map.towermode && !map.minitowermode && map.ypos >= 500 && map.ypos <= 5000;
+            const bool isInWrappingAreaOfTower = map.towermode && !map.minitowermode && map.ypos >= 500 && map.ypos <= 5000;
             if (wrapX && (map.warpx || isInWrappingAreaOfTower))
             {
                 drawRect = sprites_rect;
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -1598,25 +1598,9 @@ void Graphics::drawentities()
     bool custom_gray = false;
 #endif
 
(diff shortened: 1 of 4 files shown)


InfoTeddy:
Actually, I'm going to put this PR on hold, because I just realized something. (The entity refactor changes will be cherry-picked; I'll open a separate PR for those.)

So, actually, it's not as if it's impossible for custom levels to fix their draw order in such a way that their crewmates will be on top. All they have to do is create the crewmate before whatever else they want to create behind them. And actually, it's sometimes useful to take advantage of the draw order to intentionally draw a crewmate behind an entity - I even used it in my VVVVVV Discord Halloween 2020 contest submission:

Cyan man headstand

With my patch, this would be impossible because the crewmate would always be drawn on top, and there'd be no way to change the draw order to get it any other way.

On the other hand, though, the issue with the draw order in custom levels (and this applies to the main game, too) is that entities specific to a room, such as checkpoints and terminals, get created first upon room load, before crewmates do. After a room is loaded, the main game has to create a crewmate using companion logic or the special scripting logic in `mapclass::loadlevel()`, and a custom level has to create a crewmate by using an internal script that calls `createentity()`. Edentity crewmates can be placed in front of other entities, because the game adheres to edentity order - but edentity crewmates are also only rescuable, and you can't really do much with them, and you might not want to use a rescuable crewmate for a cutscene (most people don't). So, it's impossible to get a crewmate to be drawn in front of a terminal or checkpoint that is created immediately upon room load.

But I would like that Verdigris regression to be fixed, and for the draw order to be fixed for at least the main game. I could add a `custommode` check, but I'd like to head in a direction where custommode and non-custommode are the same thing and there's no specific main-game-code or custom-level-code, if at all possible. So I'm sort of torn on this one.

InfoTeddy:
Well, I've decided to compromise and just throw in the `custommode` check. A proper solution for the main game that doesn't have a `custommode` check would be one where crewmates get created before entities get loaded in a room - or some function that rearranges the entity order around to put crewmates first. But either of those will have to be added in 2.4, this will do fine for now.

desktop_version/src/Graphics.cpp
```diff
@@ -1585,356 +1591,364 @@ void Graphics::drawtrophytext()
...
+#if !defined(NO_CUSTOM_LEVELS)
+    // Special case for gray Warp Zone tileset!
+    const int room = game.roomx-100 + (game.roomy-100) * ed.maxwidth;
+    const bool custom_gray = INBOUNDS_ARR(room, ed.level)
+    && ed.level[room].tileset == 3 && ed.level[room].tilecol == 6;
+#else
+    const bool custom_gray = false;
+#endif
+
+    std::vector<SDL_Surface*>& tilesvec = map.custommode && !map.finalmode ? entcolours : tiles;
```
flibitijibibo, line 1649: Had a bit of trouble reading this one line:

```suggestion
    std::vector<SDL_Surface*>& tilesvec = (map.custommode && !map.finalmode) ? entcolours : tiles;
```

InfoTeddy: Fixed, rebased, and force-pushed. (Not using GitHub's thing because that adds an unnecessary commit.)

flibitijibibo:
Did the force-push go through? Looks like it's still the push from an hour ago.

flibitijibibo:
Whuh, nevermind, just took a second to refresh.

Thanks for all the quick fixes!

InfoTeddy:
Yeah, I rebased on the wrong branch so then I quickly did it for this branch.

PR #527, opened 2020-11-04
Move all settings to settings.vvv

### LINKED ISSUE #373: Add XML forwards compatibility

InfoTeddy:
The game previously did this dumb thing where it lumped in all its settings with its file that tracked your records and unlocks, `unlock.vvv`. It wasn't really an issue, until 2.3 came along and added a few settings, suddenly making a problem where 2.3 settings would be reset by chance if you decided to touch 2.2.

The solution to this is to move all settings to a new file, `settings.vvv`. However, for compatibility with 2.2, settings will still be written to `unlock.vvv`.

The game will prioritize reading from `settings.vvv` instead of `unlock.vvv`, so if there's a setting that's missing from `unlock.vvv`, no worries there. But if `settings.vvv` is missing, then it'll read settings from `unlock.vvv`. As well, if `unlock.vvv` is missing, then `settings.vvv` will be read from instead (I explicitly tested for this, and found that I had to write special code to handle this case, otherwise the game would overwrite the existing `settings.vvv` before reading from it; kids, make sure to always test your code!).

Closes #373 fully.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -4436,6 +4436,14 @@ void Game::deletestats()
     }
 }
 
+void Game::deletesettings()
+{
+    if (!FILESYSTEM_delete("saves/settings.vvv"))
+    {
+        puts("Error deleting saves/settings.vvv");
+    }
+}
+
 void Game::unlocknum( int t )
 {
 #if !defined(MAKEANDPLAY)
@@ -4471,7 +4479,10 @@ void Game::loadstats(int *width, int *height, bool *vsync)
     tinyxml2::XMLDocument doc;
     if (!FILESYSTEM_loadTiXml2Document("saves/unlock.vvv", doc))
     {
-        savestats();
+        // Save unlock.vvv only. Maybe we have a settings.vvv laying around too,
+        // and we don't want to overwrite that!
+        savestats(true);
+
         printf("No Stats found. Assuming a new player\n");
     }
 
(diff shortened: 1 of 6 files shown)


desktop_version/src/Game.cpp
```diff
@@ -4926,8 +4973,58 @@ void Game::savestats()
...
+    {
+        savesettings();
+        puts("No settings.vvv found");
+    }
+
+    tinyxml2::XMLHandle hDoc(&doc);
+    tinyxml2::XMLElement* pElem;
+    tinyxml2::XMLHandle hRoot(NULL);
+
+    {
```
flibitijibibo, line 4991: What's up with this block? Seems like it's missing a return and has some weird indentation... otherwise the rest seems okay.

InfoTeddy: That's just how the other XML files are handled by the game (go look at `Game::loadstats()`, for instance). It's bad because of the weird indentation and the fact that it does error checking only to throw it away, but I'm keeping it for consistency with the rest of the code.

It should probably be fixed, though. And all the XML parsing should probably have error handling (immediate segfault if any element in an XML document is missing!!!).

flibitijibibo: Ah, got it - this is one of those things your brain blocks out after you see it and ship it.

PR #535, opened 2020-11-07
Fix game loop order to what it was in 2.2

### LINKED ISSUE #464: Revert game loop order to what it was in 2.2
2.3 added over-30-FPS mode (#220), and as part of the changes required, I had to move all render functions to the end of the frame. For the game-gamestate GAMEMODE, 2.2's game loop order was `gameinput()`, then `gamerender()`, then `gamelogic()`, while 2.3's was `gameinput()`, then `gamelogic()`, then `gamerender()`. This was due to the fact that all render functions needed special code to render more than 30 frames a second.

This has caused a few issues, due to the fact that some parts of the game expect some other code to be ran before the next `gamerender()` gets ran, some of them most notably expecting `script.run()` to get ran before the next render. These issues include but are not limited to:
- Violet appearing 1 frame after the Ship teleporter room appears (#402, fixed by #463)
- Gotorooming while player stands inside gravity lines coloring those gravity lines gray instead of white (fixed by #488)

But funnily enough, this new order also removes the 1-frame input delay between pressing left/right/ACTION and Viridian moving. That's because in 2.2, input is taken in `gameinput()`, then `gamerender()` gets called. Viridian moves in `gamelogic()`, but the player won't be able to see that they moved until the next `gamerender()` is called, and there was already a `gamerender()` after their input where they saw Viridian staying in the same place they were previously.

In hindsight, I now think there's a way to fix the order while still having that special rendering code for each render function. However, while fixing the order may fix rendering glitches, it might also introduce new ones we don't know about.

~~After discussing this in Discord DMs with @flibitijibibo, we agreed that the order should be fixed in 2.4, as it would need a separate pass.~~ Well, @tzann asked why we couldn't just fix the order now, since the tag date is on January 10th anyway, so after some discussion with @flibitijibibo we agreed that we could fix the order in 2.3 now.

InfoTeddy:
Okay, so the reason why all render functions were moved to the end of the frame in #220 is because it's simpler to call two fixed functions and then a delta function instead of one fixed function, then a delta function, and then another fixed function.

This is because fixed functions need special handling inside `deltaloop()`, and you can't simply duplicate this handling after calling a delta function. Oh, and to make matters worse, it's not always fixed-delta-fixed, sometimes (like in MAPMODE and TELEPORTERMODE) it's delta-fixed-fixed, so we'd need to handle that somehow too.

The solution here is to generalize the game loop and factor out each function, instead of hardcoding it. Instead of having hardcoded case-switches directly in the loop, I made a function that returns an array of functions for a given gamestate, along with the number of functions, then the game loop processes it accordingly. In `fixedloop()`, it iterates over the array and executes each function until it reaches a delta function, at which point it stops. And when it reaches the end of the array, it goes back to the start of the array.

But anyway, if it gets to a delta function, it'll stop the loop and finish `fixedloop()`. Then `deltaloop()` will call the delta function. And then on the next frame, the function index will be incremented again, so `fixedloop()` will call the fixed functions again.

Actually, the previous game loop was actually made up of one big loop, with a gamestate function loop nested inside it, flanked with code that ran at the start and end of the "big loop". This would be easy to handle with one loop (just include the beginning and end functions with the gamestate functions in the array), except that the gamestate functions could suddenly be swapped out with unfocused functions (the ones that run when you unfocus the window) at any time (well, on frame boundaries, since `key.isActive` only got checked once, guarding the entire "inner loop" - and I made sure that changing `key.isActive` wouldn't immediately apply, just like the previous game loop order) - so I had to add yet another layer of indirection, where the gamestate functions could immediately be swapped out with the unfocused functions (while still running the beginning and end code, because that was how the previous loop order worked, after all).

This also fixes a regression that the game loop that #220 introduced had, where if the fixed functions switched the gamestate, the game would prematurely start rendering the gamestate function of the new gamestate in the deltaframes, which was a source of some deltaframe glitches. But fixing this is likely to just as well cause deltaframe glitches, so it'd be better to fix this along with fixing the loop order, and only have one round of QA to do in the end, instead of doing one round after each change separately.

This PR fixes a bunch of regressions, including backgrounds taking 1 frame to update, Viridian's respawn animation variables being 1 frame out of sync with each other, and 1-frame glitches when going to and from in-game options from editor settings; it also removes a bunch of kludge, like the `createentity()` deltaframe kludge, or the gravity line constant-`gotoroom` kludge fix, or the fix to update entity colors whenever a room is entered in finalmode.

Fixes #464.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -125,6 +125,7 @@ SET(VVV_SRC
 	src/WarpClass.cpp
 	src/XMLUtils.cpp
 	src/main.cpp
+	src/DeferCallbacks.c
 	src/Network.c
 	src/ThirdPartyDeps.c
 )
--- a/desktop_version/src/DeferCallbacks.c
+++ b/desktop_version/src/DeferCallbacks.c
@@ -1,0 +1,84 @@
+#include "DeferCallbacks.h"
+
+#include <SDL.h>
+
+/* Callbacks to be deferred to the end of each sequence of gamestate functions
+ * in main. Useful for fixing frame-flicker glitches when doing a state
+ * transition in a function that gets executed before the render function.
+ *
+ * We store a linked list of callbacks, to allow for the possibility of having
+ * more than one callback active at a time (otherwise we could easily just
+ * have a single pointer here and the header would only be 1 line and an
+ * include guard) and to do it without having to allocate memory at runtime.
+ */
+
+static struct DEFER_Callback* head = NULL;
+
+/* Add a callback. Don't call this directly; use the DEFER_CALLBACK macro. */
+void DEFER_add_callback(struct DEFER_Callback* callback)
+{
+    struct DEFER_Callback* node;
+
+    /* Are we adding the first node? */
+    if (head == NULL)
+    {
+        head = callback;
+        return;
+    }
+
+    /* Time to walk the linked list */
+    node = head;
+
+    if (node == callback)
+    {
+        goto fail;
+    }
+
+    while (node->next != NULL)
+    {
+        node = node->next;
+
+        if (node == callback)
+        {
+            goto fail;
+        }
+    }
+
+    /* We're at the end */
+    node->next = callback;
+
+    /* Success! */
+    return;
+
+fail:
+    /* Having multiple instances of a callback isn't well-defined
+     * and is a bit complicated to reason about */
+    SDL_assert(0 && "Duplicate callback added!");
+}
+
+/* Call each callback in the list, along with deleting the entire list. */
+void DEFER_execute_callbacks(void)
+{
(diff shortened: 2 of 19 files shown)


flibitijibibo:
This and #565 are the last major blockers, so these should be the absolute top priority for all development until they're finished - the rest of the 2.3 items are either ready to review already or are looking to be bumped to 2.4 instead.

flibitijibibo:
Man that deferred system is smart. Makes me want to double check FAudio's linked lists to see if I can stop mallocing so much...

Anyway, this looks good as far as I can tell. It does teeter on over-engineered with the loop-struct-framework thing, but for a game loop this wacky I suppose it'll be worth having in the long run. Well done overall!

PR #539, opened 2020-11-07
Move all fixed-timestep render updates to new file RenderFixed.cpp

InfoTeddy:
As part of fixing #464, I'll need to move these pieces of code around easily. In #220 I just kind of shoved them awkwardly in whatever fixed function would be last called in the gamestate loop, which I shouldn't have done as I've now had to make formal fixed-render functions anyway. Because these fixed functions need to be called directly before a render function, and I'm fixing the order to put render functions in their proper place, so I need to be able to move these around easily, and making them function calls instead of inlined makes them easier to manipulate.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/preloader.cpp
+++ b/desktop_version/src/preloader.cpp
@@ -10,7 +10,7 @@ int pre_darkcol=0, pre_lightcol=0, pre_curcol=0, pre_coltimer=0, pre_offset=0;
 int pre_frontrectx=30, pre_frontrecty=20, pre_frontrectw=260, pre_frontrecth=200;
 int pre_temprectx=0, pre_temprecty=0, pre_temprectw=320, pre_temprecth=240;
 
-void preloaderlogic()
+void preloaderrenderfixed()
 {
   if (pre_transition < 30) pre_transition--;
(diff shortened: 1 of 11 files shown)


InfoTeddy:
Just a small update to put the `game.gamestate = TITLEMODE` in `preloaderrenderfixed()` instead of `preloaderrender()` because we shouldn't be mutating state during render functions... (_\*cough\* \*cough\*_ fixing gravity lines during `editorrender()` _\*cough\* \*cough\*_)

desktop_version/src/preloader.cpp
```diff
@@ -88,7 +92,7 @@ void preloaderrender()
       pre_transition = 29;
     }
   }else if (pre_transition <= -10) {
-    game.gamestate=TITLEMODE;
+
```
flibitijibibo, line 95: Might be worth just commenting this, to avoid the empty block.

InfoTeddy: Added a comment, rebased, and force-pushed.

flibitijibibo:
Perfect, thanks! (Also sorry it took over a month to look at all this)

PR #544, opened 2020-11-12
Fix Prize for the Reckless moving platform losing its block when you die and the platform's x-coordinate is to the left of x=152

### LINKED ISSUE #543: Falling through moving platform on Prize for the Reckless
Title. I've tried getting help with this in the v6cord and nobody else can seem to reproduce my issue. I compiled the game with cmake on Arch Linux, game runs fine, no errors at all, but no matter what options I change / if I have 60 or 30fps on I will always fall through the moving platform leading to the trinket on prize for the reckless. It doesn't matter what direction it moves, if it's touching the wall at the time I get on the platform, I will always phase right through it, making 100% completion impossible. It's compiled from the latest commit.

InfoTeddy:
This bug would happen if you die and the platform isn't completely clear of the space of its adjacent disappearing platform.

The block needs to be updated accordingly with calls to `obj.nocollisionat()` and `obj.moveblockto()`, else the block will simply be left behind and the platform will no longer have any collision. This is in contrast to 2.2 behavior, where the platform would simply unconditionally create a new block, which would actually end up with a duplicate block since the previous block didn't get cleaned up, but this didn't cause any problems because the room was carefully designed so you would never be able to touch that previous block after you died and respawned at the checkpoint. But it's still there.

I also added comments to document what this kludge code did, because otherwise it would be mysterious to readers who are unfamiliar with it.

Fixes #543.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Logic.cpp
+++ b/desktop_version/src/Logic.cpp
@@ -465,8 +465,17 @@ void gamelogic()
                 {
                     if (obj.entities[i].xp < 152)
                     {
+                        //Move the platform to the right side of the disappearing platform,
+                        //otherwise it will get stuck on the kludge 18,9 tile we placed
+                        //(and if the tile wasn't there it would pass straight through again)
+                        int prevx = obj.entities[i].xp;
+                        int prevy = obj.entities[i].yp;
+                        obj.nocollisionat(prevx, prevy);
+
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Kludgetastic! Thanks to everyone who checked this out.

PR #547, opened 2020-11-14
Fix Secret Lab Time Trial trophies having wrong colors

InfoTeddy:
Ever since 2.0, the colors of some of the Time Trial trophies in the Secret Lab don't correspond to the crewmate of the given level. The trophy for the Tower uses Victoria's color, and the Lab trophy uses Vermilion's color. The Space Station 2 trophy uses Viridian's color, and the Final Level trophy uses Vitellary's color.

This doesn't appear to be intentional, and it would be odd if it was, since this game matches the colors everywhere else (each zone on the map is colored with their respective crewmate in mind, for instance). Also, the Lab trophy has the sad expression, which is Victoria's trait - it would be weird if this was intended for Vermilion instead.

But the biggest piece of evidence that this was unintentional is the corresponding comment for each color in `Graphics::setcol()`. It mislabels yellow as cyan, cyan as yellow, blue as red, and red as blue.

To fix this, I simply have to set the correct color for each trophy in case 25 of `entityclass::createentity()`. I could fix it in `Graphics::setcol()` itself, but custom levels might depend on those certain colors being the way they are, so it's a safer bet to just fix it in the trophy creation case itself.

The diff of this might look weird. Even though all I'm doing is changing some value assignments around, it looks like the "patience" algorithm thinks I'm moving a whole case of the trophy switch-case around.

Here are the screenshots showing before and after, with each trophy labeled since you have to stand on them to show their text:

Before:
Trophies before

After:
Trophies after

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1745,21 +1745,21 @@ void entityclass::createentity( float xp, float yp, int t, float vx /*= 0*/, flo
             if(game.bestrank[1]>=3)
             {
                 entity.tile = 186 + vx;
-                entity.colour = 35;
+                entity.colour = 33;
             }
             break;
         case 3:
             if(game.bestrank[2]>=3)
             {
                 entity.tile = 184 + vx;
-                entity.colour = 33;
+                entity.colour = 35;
             }
             break;
         case 4:
             if(game.bestrank[3]>=3)
             {
                 entity.tile = 184 + vx;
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Makes sense to me, but just in case I'll defer to @TerryCavanagh...

TerryCavanagh:
Yep, this is correct! Thanks for the fix!

PR #548, opened 2020-11-16
Restrict the ENTER menu to saving while in a cutscene

Dav999-v:
## Changes:

PR #468 made it so you can use the menus while in a cutscene, in order to counteract softlocks. However, this has resulted in more unintentional behavior:
- `gamemode(teleporter)` breaks when opening the ENTER menu (Misa mentioned this)
- The player can now interrupt shakes and walks, and have their timers run out before resuming the cutscene
Hurry hurry
- After completing the game, the player can warp to the ship while a dialogue is active, and prevent themselves from advancing text (plus it's always rude to just teleport away while someone's talking)
"Ok whatever, bye!"
- The player can peek at the map before hidecoordinates is run, and can also peek at what the game does with missing/rescued crewmates during cutscenes
Evacuation map
Everyone's location

This PR fixes the latter two issues. While a script is running, only the SAVE tab is now available. Therefore the player can still get themselves out of softlocks as intended in #468, but they can't do things like looking at the map or teleporting away during a cutscene.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2044,7 +2044,14 @@ void gameinput()
         game.gamesaved = false;
         game.gamesavefailed = false;
         graphics.resumegamemode = false;
-        game.menupage = 0; // The Map Page
+        if (script.running)
+        {
+            game.menupage = 3; // Only allow saving
+        }
+        else
+        {
+            game.menupage = 0; // The Map Page
+        }
         BlitSurfaceStandard(graphics.menubuffer,NULL,graphics.backBuffer, NULL);
         graphics.menuoffset = 240; //actually this should count the roomname
         graphics.oldmenuoffset = 240;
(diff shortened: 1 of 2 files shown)


InfoTeddy:
The rendering isn't changed, so when you're on the SAVE screen it looks like all the other options are available at the bottom, when they aren't.

Dav999-v:
> The rendering isn't changed, so when you're on the SAVE screen it looks like all the other options are available at the bottom, when they aren't.

I was thinking of not showing them, but they're already "grayed out", and not being able to use left/right to switch between them felt like enough feedback that the other tabs are locked out to me. Should the other options be hidden completely? Without having tested it out, it kinda sounds like that would also wrongly set the in-cutscene menu apart from the regular menu with all the options available, instead of just the same menu with some options unavailable.

InfoTeddy:
I see your point, but I was thinking that someone might be confused and think that there's some sort of bug preventing you from moving over to the other options, instead of it being intentional. Yeah, it'd be better if all the other options were gone, and "[SAVE]" would be re-centered in the middle of the screen.

Dav999-v:
> I see your point, but I was thinking that someone might be confused and think that there's some sort of bug preventing you from moving over to the other options, instead of it being intentional. Yeah, it'd be better if all the other options were gone, and "[SAVE]" would be re-centered in the middle of the screen.

Alright, done in ec60c2395cd1f8b60cfdf006003dcd1d0d9534d3.

saveonly

PR #554, opened 2020-12-20
Allow pressing ACTION to skip the fake loading screen

InfoTeddy:
While there already exists an option to skip the fake loading screen entirely (without requiring an ACTION press), there are several reasons for including this option as well:

 * So people upgrading from 2.2 won't have to sit through the fake loading screen the first time they open 2.3.

 * So if people are too lazy to use the existing option, they can use this one instead.

 * So tool-assisted speedruns (TASes) of this game can skip the fake loading screen without requiring an existing `settings.vvv` beforehand.

   This last one is the biggest reason for me, since I'm not sure what TASVideos.org rules are regarding existing save files, but with this change nobody has to worry about their rules and can safely just press ACTION to skip the fake loading screen automatically.

Currently, when you press ACTION, the game will draw a solid black frame before drawing the ACTION prompt of `TITLEMODE`. This is because the current game loop code immediately switches which rendering functions it uses if `game.gamestate` changes in the middle of the execution of the loop, but this is a bug that will be fixed when #535 gets merged, which prevents switching rendering functions in the middle of the game loop.

Also, I discussed with @flibitijibibo in Discord DMs about adding this change, and his response was that he would defer to @TerryCavanagh whether or not this should be merged, because he doesn't know how important the fake loading screen is to Terry.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -527,6 +527,7 @@ void inline fixedloop()
         switch(game.gamestate)
         {
         case PRELOADER:
+            preloaderinput();
             preloaderrenderfixed();
             break;
 #if !defined(NO_CUSTOM_LEVELS) && !defined(NO_EDITOR)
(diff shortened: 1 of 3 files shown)


TerryCavanagh:
I'm fine with this change 👍

Issue #556, opened 2020-12-21
Segfault on startup if unlock.vvv or settings.vvv is missing


leo60228:
https://github.com/TerryCavanagh/VVVVVV/commit/e73c14d1f373905f3d16c77947b6fba7474d678e removes the "unnecessary" Screen options from Game. However, they're there for a reason: Screen is created after settings are loaded, and settings are saved automatically when loading them without an existing save. This leads to a segfault due to dereferencing the null `screenbuffer` pointer.

InfoTeddy:
> However, they're there for a reason: [reason]

I don't think that's quite the full story.

These duplicate variables already existed in 2.2, where `Screen` was created (initialized) *before* `game.loadstats()` got called. With the merging of #220 and the addition of the VSync option, we ran into a bug with SDL2 where toggling the VSync flag wouldn't apply until the renderer got destroyed and re-created. So after working around that, we also made it so `Screen` got created *after* `game.loadstats()` got called, so we wouldn't unnecessarily destroy the renderer and re-create it again. And this was necessary to fix, because the reset renderer workaround doesn't work on Haiku; you need to close the game and open it again, but if the game does the workaround on startup, then there's no way for the VSync option to properly apply on startup.

All of that is to say, it's completely anachronistic to say those duplicate variables were there because of 2.3 startup order, when they already existed previously in 2.2. And I still stand by my thinking that those duplicate variables shouldn't exist.

PR #557, opened 2020-12-22
Refactor Game::savestats() to not use a default argument

InfoTeddy:
In order to be able to fix the bug #556, I'm planning on adding `ScreenSettings*` to the `settings.vvv` write function. However, that entails adding another argument to `Game::savesettings()`, which is going to be really messy given the default argument of `Game::savestats()`. That, combined with the fact that the code comment at the site of the implementation of `Game::savestats()` being wrong (!!!), leads me to believe that using default function arguments here isn't worth it.

Instead, what I've done is made it so callers are explicit about whether or not they're calling `savestats()`, `savesettings()`, or both at the same time. If they are calling both at the same time, then they will be using a new function named `savestatsandsettings()`.

In short, these are the interface changes:
 * `bool Game::savestats(bool)` has been removed
 * `bool Game::savestatsandsettings()` has been added
 * `void Game::savestats_menu()` has been renamed to `void Game::savestatsandsettings_menu()`
 * All previous callers of `bool Game::savestats()` are now using `bool Game::savestatsandsettings()`
 * The one caller of `bool Game::savestats(bool)` is now using `bool Game::savestats()`

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -1334,7 +1334,7 @@ void Game::updatestate()
                 }
             }
 
-            savestats();
+            savestatsandsettings();
 
             graphics.fademode = 2;
             music.fadeout();
@@ -3133,7 +3133,7 @@ void Game::updatestate()
         }
 
 
-            savestats();
+            savestatsandsettings();
             if (nodeathmode)
             {
                 unlockAchievement("vvvvvvmaster"); //bloody hell
@@ -4447,7 +4447,7 @@ void Game::unlocknum( int t )
     }
 
     unlock[t] = true;
-    savestats();
+    savestatsandsettings();
 #endif
 }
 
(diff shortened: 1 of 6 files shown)


desktop_version/src/Game.cpp
```diff
@@ -4812,20 +4812,18 @@ bool Game::savestats(const bool stats_only /*= true*/)
...
-    {
-        success &= savesettings();
-    }
+    return FILESYSTEM_saveTiXml2Document("saves/unlock.vvv", doc);
+}
 
-    return success;
+bool Game::savestatsandsettings()
+{
+    return savestats() && savesettings();
```
flibitijibibo, line 4820: These may need to be or'd so settings aren't skipped when stats fail to save.

InfoTeddy: Ah, good point. Fixed and rebased.

PR #560, opened 2020-12-24
De-duplicate script.running checks

InfoTeddy:
While I was working on #535, I noticed that all the call sites of `script.run()` have the exact same code, namely:

```c++
if (script.running)
{
    script.run();
}
```

I figured, why not move the `script.running` check into the function itself? That way, we won't have to duplicate the check every single time, and we don't risk forgetting to add the check and causing a bug because of that.

The check was already duplicated once since 2.0 (it's used in both GAMEMODE and TELEPORTERMODE), and with the fix of the two-frame delay in 2.3, it's now duplicated twice, leading to *three* instances of this check in the code, when there should be only one.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -2101,9 +2101,6 @@ void mapclass::twoframedelayfix()
 	game.state = 0;
 	game.statedelay = 0;
 	script.load(game.newscript);
-	if (script.running)
-	{
-		script.run();
-		script.dontrunnextframe = true;
-	}
+	script.run();
+	script.dontrunnextframe = true;
 }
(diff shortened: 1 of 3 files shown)


desktop_version/src/Map.cpp
```diff
@@ -2101,9 +2101,6 @@ void mapclass::twoframedelayfix()
 	game.state = 0;
 	game.statedelay = 0;
 	script.load(game.newscript);
-	if (script.running)
-	{
-		script.run();
-		script.dontrunnextframe = true;
-	}
+	script.run();
+	script.dontrunnextframe = true;
```
flibitijibibo, line 2105: Do we care that this now gets unconditionally set to true?

InfoTeddy: It shouldn't matter. It'll be removed when #535 gets merged anyway.

PR #567, opened 2020-12-28
Consistently set game.lifeseq to 0 when script.startgamemode() is called

InfoTeddy:
There's a small inconsistency where the first time you load in to the game, `game.lifeseq` is at 0, but when you exit to the menu and load in again, `game.lifeseq` becomes 10. This is visible as Viridian blinking when loading in only after the first time you load in, and this also means that after the first time you load in, you also have to wait 5 frames before being able to move Viridian.

The reason for this inconsistency is because on the first time you load in to the game, there are no entities loaded in `obj.entities` yet, so the game creates a player entity, and doesn't mess with `game.lifeseq`. When you exit and then load in for the second time, `obj.entities` contains at least one entity (all the entities from the room you just exited out of - `map.gotoroom()` hasn't even been called yet, so it doesn't even check for only the player entity), so the game calls `map.resetplayer()` instead, and `map.resetplayer()` sets `game.lifeseq` to 10.

There's even an inconsistency to this inconsistency - when you die in No Death Mode, all entities will be removed from `obj.entities`, so the next time you load in to the game, `game.lifeseq` will be 0.

This inconsistency is pretty minor in the grand scheme of things, but it still bothers me, so I'm going to fix it.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -2675,6 +2675,7 @@ void scriptclass::startgamemode( int t )
 		{
 			map.resetplayer();
 		}
+		game.lifeseq = 0;
 		map.gotoroom(game.saverx, game.savery);
 		map.initmapdata();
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Qualifies as an uninitialized memory bug in my books... the worst kind, too.

PR #573, opened 2021-01-01
Set game as DPI-aware

leo60228:
## Changes:

This prevents the OS from performing bilinear scaling in addition to whatever the game is doing. SDL_WINDOW_ALLOW_HIGHDPI should do this on all platforms, but it's only implemented for macOS, Wayland (X11 doesn't have a concept of DPI-awareness), and a few more esoteric ones. Windows has a few ways of doing it, but SetProcessDPIAware is the easiest. The docs say that this can cause unexpected behavior, but my understanding is that this relates to multithreading, and in this case nothing should be able to go wrong.

I don't currently have a way to do Windows builds locally, so I'm relying on CI to make sure this compiles. It'd be nice if someone could test it on HiDPI Windows/Wayland/macOS, but I'm pretty confident that this should work if it builds.
## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -60,7 +60,7 @@ void Screen::init(const ScreenSettings& settings)
 	SDL_CreateWindowAndRenderer(
 		640,
 		480,
-		SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE,
+		SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI,
 		&m_window,
 		&m_renderer
 	);
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -170,7 +170,7 @@ void Screen::ResizeScreen(int x, int y)
 	if (stretchMode == 1)
 	{
 		int winX, winY;
-		SDL_GetRendererOutputSize(m_renderer, &winX, &winY);
+		GetWindowSize(&winX, &winY);
 		int result = SDL_RenderSetLogicalSize(m_renderer, winX, winY);
 		if (result != 0)
 		{
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -170,7 +170,7 @@ void Screen::ResizeScreen(int x, int y)
 	if (stretchMode == 1)
 	{
 		int winX, winY;
-		SDL_GetWindowSize(m_window, &winX, &winY);
+		SDL_GetRendererOutputSize(m_renderer, &winX, &winY);
 		int result = SDL_RenderSetLogicalSize(m_renderer, winX, winY);
 		if (result != 0)
 		{
@@ -256,7 +256,7 @@ void Screen::ResizeToNearestMultiple()
 
 void Screen::GetWindowSize(int* x, int* y)
 {
-	SDL_GetWindowSize(m_window, x, y);
+	SDL_GetRendererOutputSize(m_renderer, x, y);
 }
 
 void Screen::UpdateScreen(SDL_Surface* buffer, SDL_Rect* rect )
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -22,6 +22,10 @@
 #include "SoundSystem.h"
 #include "UtilityClass.h"
 
+#ifdef _WIN32
+#include <windows.h>
+#endif
+
 scriptclass script;
 
 #if !defined(NO_CUSTOM_LEVELS)
@@ -158,6 +162,10 @@ int main(int argc, char *argv[])
             return 1;
         }
     }
+
+#ifdef _WIN32
+    SetProcessDPIAware();
+#endif
 
     if(!FILESYSTEM_init(argv[0], baseDir, assetsPath))
     {


desktop_version/src/Screen.cpp
```diff
@@ -170,7 +170,7 @@ void Screen::ResizeScreen(int x, int y)
 	if (stretchMode == 1)
 	{
 		int winX, winY;
-		SDL_GetWindowSize(m_window, &winX, &winY);
+		SDL_GetRendererOutputSize(m_renderer, &winX, &winY);
```
InfoTeddy, line 173: This call should use `Screen::GetWindowSize()` like all other instances of this call.

InfoTeddy (changes requested):

sylveon:
The method used for Windows doesn't make the app per-monitor DPI aware, which means that if the game is played on a monitor other than the primary one, upscaling or downscaling of the application window may take place.

The documentation for the API also mentions that using the API instead of app manifests isn't recommended: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiaware

InfoTeddy:
Yep, the Microsoft docs you linked to were already discussed on the VVVVVV Discord server. Leo's reaction to the "unexpected application behavior" remark: "thanks for the detail there".

However, I did not know about the detail of per-monitor DPI awareness. Does using manifests fix this? The docs don't say.

leo60228:
Well, VVVVVV *isn't* per-monitor DPI aware.

* No DPI awareness (default): From VVVVVV's perspective, it's always running on a standard DPI monitor, and Windows will upscale it to the correct DPI.
* System DPI awareness (this PR): From VVVVVV's perspective, it's always running on the same DPI monitor for the duration of the program, and Windows will upscale it to the current DPI if necessary. Since VVVVVV doesn't care what DPI the monitor is (it does its own upscaling no matter what), this is fine.
* Per-monitor DPI aware: VVVVVV sees the actual DPI of the monitor it's being displayed on at all times, and is expected to resize the window automatically upon a DPI change. This isn't implemented, and I'm not fully convinced that it should be.

There's also an additional issue with per-monitor DPI awareness: it wasn't introduced until Windows 8.1. While that's the oldest currently supported version of Windows, with how conservative the game is with minimum requirements in other regards, I assumed that keeping compatibility with Vista (or at least 7) was desired.

sylveon:
Yeah, [manifests allow to set per-monitor DPI awareness](https://docs.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process#setting-default-awareness-with-the-application-manifest).

Usually, programs would use a combination of both `dpiAware` and `dpiAwareness`, something like this:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
  <asmv3:application>
    <asmv3:windowsSettings>
      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true/pm</dpiAware>
      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
    </asmv3:windowsSettings>
  </asmv3:application>
</assembly>
```

> There's also an additional issue with per-monitor DPI awareness: it wasn't introduced until Windows 8.1. While that's the oldest currently supported version of Windows, with how conservative the game is with minimum requirements in other regards, I assumed that keeping compatibility with Vista (or at least 7) was desired.

Manifests also makes handling the cases where per-monitor isn't available (on older OS for example) simpler, since the OS will automatically fallback to system-aware, instead of the program needing to try setting the DPI awareness until it succeeds.

To test the V1 per monitor DPI awareness on a machine with a recent OS version, one can simply remove the per monitor V2 entry as well.

> Per-monitor DPI aware: VVVVVV sees the actual DPI of the monitor it's being displayed on at all times, and is expected to resize the window automatically upon a DPI change. This isn't implemented, and I'm not fully convinced that it should be.

Unless `WM_DPICHANGED` has custom handling, Windows will automatically resize the window for you when changing DPIs. The game will see this size change as if the user grabbed the window and resized it manually. As you've said, the game does its own upscaling no matter what so this isn't an issue.

leo60228:
The CMake version used by VVVVVV is too old to automatically handle app manifests, and I'm not sure of a good way to handle it without native CMake support.

InfoTeddy:
I talked with @flibitijibibo in Discord DMs about this earlier. He said that it's completely fine to just drag the manifest file alongside the executable, because it would at least be exposed to end users.

flibitijibibo:
Yeah, I'm okay with just having the manifest be manual - worst case we can just emit a warning at runtime if the file isn't found.

The alternative is to try and add programmatic support by dlopening user32.dll, but who knows how much further Windows hi-dpi is going to break... for FNA games we generally just ignore it and let people shove their own mainifest into the game folder. I don't think we can add it to SDL either, even though logically it should be there... Microsoft _really_ blew it with this API compared to macOS and Wayland.

flibitijibibo:
I've pulled in all the high-DPI changes in this patchset except for the Windows commit - the manifest is probably the way to go with this unless someone who _really_ likes Windows wants to write the dynamically loaded user32 stuff.

PR #574, opened 2021-01-01
Run CI on CentOS 7

leo60228:
## Changes:

This adds a Dockerfile for a CentOS 7-based build environment, and then has Linux CI use a build of that Dockerfile. This means that Linux CI matches the environment releases get built in. However, this adds quite a bit of complexity, and introduces a reliance on my built Docker image. The image could easily be changed to be pulled from another source, however.

## Legal Stuff:

  for these changes


desktop_version/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /tmp
+RUN curl -O https://www.libsdl.org/release/SDL2-2.0.14.tar.gz
+RUN tar -xf SDL2-2.0.14.tar.gz
+RUN mkdir SDL2-2.0.14/build
+RUN curl -O https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-2.0.4.tar.gz
+RUN tar -xf SDL2_mixer-2.0.4.tar.gz
+RUN mkdir SDL2_mixer-2.0.4/build
+
+RUN yum -y install \
+    # SDL2 dependencies, adapted from https://hg.libsdl.org/SDL/file/default/docs/README-linux.md
```
flibitijibibo, line 13: A simplification might be what I do for flibitBuild:

```
yum -y install yum-utils
yum -y builddep SDL2
```

Full build setup is here: https://icculus.org/finger/flibitijibibo

desktop_version/src/Entity.cpp
```diff
@@ -11,6 +11,8 @@
 #include "Script.h"
 #include "UtilityClass.h"
 
+#include <cmath>
```
flibitijibibo, line 14: The cmath/fabs patches make sense to me, this can be submitted separately.

flibitijibibo (commented):
Do we have a sample of this running somewhere? I'd love to start using CentOS 7 for Linux testing, but I don't have much experience with Docker beyond what cybik did for ANGLE-Builds.

leo60228:
> Do we have a sample of this running somewhere?

The CI for this PR, in fact :P

desktop_version/Dockerfile
leo60228: I was avoiding using EPEL. That might make more sense, though.

desktop_version/src/Entity.cpp
leo60228: The game doesn't compile on CentOS 7 without this. I wasn't sure how to split this into a second PR and have the Git history still make sense.

InfoTeddy: What I do is make one PR with the separate change, and branch off of that PR to submit a second PR, and label that the second PR depends on the first one.

Anyways, we should be using `SDL_abs()` in `Entity.cpp` instead of `std::abs()`.

desktop_version/src/Entity.cpp
```diff
@@ -4389,8 +4391,8 @@ void entityclass::fixfriction( int t, float xfix, float xrate, float yrate )
     if (entities[t].vx > 6) entities[t].vx = 6.0f;
     if (entities[t].vx < -6) entities[t].vx = -6.0f;
 
-    if (std::abs(entities[t].vx-xfix) <= xrate) entities[t].vx = xfix;
-    if (std::abs(entities[t].vy) < yrate) entities[t].vy = 0;
+    if (std::fabs(entities[t].vx-xfix) <= xrate) entities[t].vx = xfix;
+    if (std::fabs(entities[t].vy) < yrate) entities[t].vy = 0;
```
InfoTeddy, line 4395: Why change this? Are you sure that we won't be taking out any rounding errors that resulted from the previous implicit conversion to int? (This also applies to the other places where you switched to using `fabs()`.)

InfoTeddy (changes requested):

desktop_version/src/Entity.cpp
leo60228: I assumed that any previous behavior was a bug.

desktop_version/src/Entity.cpp
leo60228: The correct function is SDL_fabsf, for what it's worth. std::fabs is overloaded.

desktop_version/src/Entity.cpp
InfoTeddy: So let's think each case through...

- `entityclass::fixfriction()` is unused, so we can ignore that. That function should probably be removed.
- `entityclass::applyfriction()` is used for entities that have a true `gravity` attribute, which just means humanoids (player and crewmates), so it's most easily tested by checking player movement.

  The code here checks if the absolute value of the player's x-velocity, after rounding towards 0, is less than 1.1 (`xrate` will always be 1.1 here), and if so, it will set their x-velocity to 0. It also checks if the absolute value of the player's y-velocity, after rounding towards 0, is less than 0.25 (`yrate` will always be 0.25), and if so, it will set their y-velocity to 0 as well. What changing this to `fabs` or `fabsf` does is take out the rounding towards 0.

  The player's x-velocity depends upon if they're holding left or right. If they're holding left or right for at least 5 frames, they will undergo deceleration when they release the directional key; else they will simply stop in place and have their x-velocity set to 0.

  So on the first frame of holding left or right, the initial value of the player's x-velocity is 3 - then `applyfriction()` will subtract 1.1 from that, leaving their x-velocity at 1.9. 1.9 is not less than 1.1, but 1.9 rounded to 1 *is* less than 1.1, so this is a bit worrying. However, their x-velocity would have been set to 0 anyways because they held the directional key for less than 5 frames, so this doesn't matter.

  So what if the player has been holding their directional key for 5 frames, and then lets go, starting to decelerate? Well, they are guaranteed to have an x-velocity of 6 at this point (because their acceleration x-velocities per frame after holding a directional key would have been 1.9 -> 3.8 -> 5.7 -> 6, so they would be guaranteed to have full speed at this point), and here is where `applyfriction()` comes in to decelerate them, subtracting 1.1 from their x-velocity every frame. This means their speed on each frame will be 6 -> 4.9 -> 3.8 -> 2.7 -> 1.6 -> 0.5 -> 0.

  But wait, this is where the rounding comes in. The function performs the check in question *after* subtracting from their x-velocity, so when the function subtracts from their x-velocity to make it 1.6, it will check: is 1.6, after rounding towards 0, less than 1.1? And 1.6 rounded towards 0 is 1, which *is* less than 1.1.

  So that means their velocity each frame on the current build is 6 -> 4.9 -> 3.8 -> 2.7 -> 0.

  But if you change it to `fabs`/`fabsf` and take out the rounding, then the function will no longer zero out their velocity when it gets to 1.6. This PR makes it so that each frame is now 6 -> 4.9 -> 3.8 -> 2.7 -> 1.6 -> 0.

  This essentially results in having an extra frame of deceleration when letting go of a directional key after holding it for 5 frames. This means you have added one extra frame of moving 2 pixels to the left, if decelerating to the left, and added one extra frame of moving 1 pixel to the right, if decelerating to the right. (The discrepancy between 2 and 1 is because of a further round towards zero when computing the `newxp` and `newyp` of the player.) Which means...

  _This is a physics change!_ **In an unrelated PR!** _**Why did you do this, Leo?!**_
- The activity zone and teleporter prompt checks both check if the player x-velocity is less than or equal to 1, after rounding it towards 0.

  As covered above, on the first frame of the player's acceleration, their x-velocity is 1.9. 1.9, after rounding this towards 0, is equal to 1. But if you take out the rounding, then 1.9 is *not* less than or equal to 1. So taking out the rounding means that you'd no longer be able to activate activity zones or teleporter prompts on the first frame of acceleration, which is yet another physics change, in a PR that's primarily about something else.

In conclusion, swapping these functions to `fabs`/`fabsf` is most likely going to break some of my TASes that synced fine on 2.2. And also, don't put unrelated changes in your PR.

flibitijibibo: Oh, huh... we should probably document that it's abs for a reason (an unintended one, admittedly). In any case, it'd be fine to just submit the compile fix separately, then this PR can pull from master and I'll squash it all down when it's ready.

InfoTeddy: Is this even a compile fix? There's tons of implicit conversions like this between float and int all around the codebase, and the most they do is emit warnings on MSVC (which should be disabled, I can probably grab a snippet from LodePNG that does that). That's why I'm so baffled Leo included this.

flibitijibibo: It might be a CFLAG - it builds on flibitBuild but I have a habit of instinctively adding -std flags, and Cent GCC default's to a very old, very strict standard.

InfoTeddy: I see. All the more reason to switch to using `SDL_abs()` instead, then. But I agree with submitting the switch to `SDL_abs()` separately.

leo60228: #575

leo60228: I initially resolved this, but it'd make more sense to wait for #575 to be merged (at which point I can do a rebase).

flibitijibibo: Merged that, thanks!

leo60228:
> It might be a CFLAG - it builds on flibitBuild but I have a habit of instinctively adding -std flags, and Cent GCC defaults to a very old, very strict standard.

This isn't directly related to this PR, but I've been doing some testing. As far as I can tell, this didn't build until GCC 7. This doesn't seem to be affected by -std either, so I'm pretty confused.

It's also worth noting that this is a hard error, not a `-Werror`-by-default warning. See https://godbolt.org/z/oxjsWP.

leo60228:
This is *probably* not a glibc change, as the relevant lines in stdlib.h have sat unchanged for over 15 years.

InfoTeddy:
Yeah, it doesn't seem like this is the end of it. Since `std::abs()` is a template overload, and I don't exactly know which version it's using, I'm doing more investigating to figure out exactly which behavior was in 2.2. I wrote my previous comment and reasoned that `std::abs()` in this case chose the int overload.

leo60228:
`std::abs` doesn't have any floating-point overloads, and there should be no difference in behavior between `int` and `long`.

leo60228:
Never mind, C is weird.

InfoTeddy:
Hm? According to [cplusplus.com](https://www.cplusplus.com/reference/cmath/abs/), the definitions of `abs()` in C++98 are:

```c++
     double abs (double x);
      float abs (float x);
long double abs (long double x);
```

InfoTeddy:
Also, if #575 was wrong, then we'll have to re-audit all the other replacements of `std::abs()` that I did.

leo60228:
It's definitely `float`: https://godbolt.org/z/cKdvWv

Oops.

flibitijibibo:
This seems good now, thanks for checking this out!

PR #576, opened 2021-01-02
Use SDL_fabsf where std::abs was previously used 

leo60228:
## Changes:

Despite @InfoTeddy having written [a short essay](https://github.com/TerryCavanagh/VVVVVV/pull/574#discussion_r550822082) about how I should've used `SDL_abs` in #575, it turns out that [this isn't actually true](https://godbolt.org/z/Pnh3Gj).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -4389,8 +4389,8 @@ void entityclass::fixfriction( int t, float xfix, float xrate, float yrate )
     if (entities[t].vx > 6) entities[t].vx = 6.0f;
     if (entities[t].vx < -6) entities[t].vx = -6.0f;
 
-    if (SDL_abs(entities[t].vx-xfix) <= xrate) entities[t].vx = xfix;
-    if (SDL_abs(entities[t].vy) < yrate) entities[t].vy = 0;
+    if (SDL_fabsf(entities[t].vx-xfix) <= xrate) entities[t].vx = xfix;
+    if (SDL_fabsf(entities[t].vy) < yrate) entities[t].vy = 0;
 }
 
 void entityclass::applyfriction( int t, float xrate, float yrate )
@@ -4410,8 +4410,8 @@ void entityclass::applyfriction( int t, float xrate, float yrate )
     if (entities[t].vx > 6.00f) entities[t].vx = 6.0f;
     if (entities[t].vx < -6.00f) entities[t].vx = -6.0f;
 
-    if (SDL_abs(entities[t].vx) < xrate) entities[t].vx = 0.0f;
-    if (SDL_abs(entities[t].vy) < yrate) entities[t].vy = 0.0f;
+    if (SDL_fabsf(entities[t].vx) < xrate) entities[t].vx = 0.0f;
+    if (SDL_fabsf(entities[t].vy) < yrate) entities[t].vy = 0.0f;
 }
 
 void entityclass::updateentitylogic( int t )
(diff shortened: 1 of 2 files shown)


flibitijibibo:
It was right in a stdlib that makes any sense! Will CC @InfoTeddy for approval juuust in case but this should be okay... I hope...

InfoTeddy:
I am looking at this now, still doing my testing...

InfoTeddy:
Okay, so here were my results:
- VVVVVV 2.0 and 2.2: 4 frames of deceleration
- VVVVVV 2.3 before #575: 5 frames of deceleration somehow?
- VVVVVV 2.3 after #575: Viridian can't even move left or right
- This PR: 5 frames of deceleration

So the problem with #575 is that it essentially disabled Viridian's horizontal movement, i.e. being really wrong. But for some reason, 2.3 has an extra deceleration frame.

I counted by using libTAS and counting the amount of frames Viridian kept moving after releasing a held direction for 5 frames - accounting for the 1-frame input delay, if there was one (this applies to 2.0 and 2.2, but not 2.3, and I also checked #535 because it had the 1-frame input delay, but I'm planning on re-removing the 1-frame input delay from that PR because it feels really awful in over-30 mode). If there was a 1-frame input delay, I would wait 1 extra frame after releasing the key before counting. After double-checking all my tests, those are the results I got above.

In any event, this PR should be merged because #575 made Viridian simply unable to move horizontally at all, but I'd like to reiterate that we need to review all instances of `std::abs()` being replaced with an SDL stdlib function, so this is still not quite the end of it.

leo60228:
The `std::abs` in 2.2's `applyfriction` are compiled as `(float)((uint)x & 0x7fffffff)` (the cast is likely due to the comparison immediately afterwards, but I suppose that this might not be the case). This should be the same as `SDL_abs`.

leo60228:
This code, however:
```c
#include <cstdlib>

int SDL_abs(int x) {
    return abs(x);
}

float abs_test(float x) {
    return SDL_abs(x);  
}
```
becomes this after optimization/decompilation (with latest g++, which may be the cause):
```c
ulong abs_test(float x) {
  uint y = (int)x >> 0x1f;
  return (ulong)(((int)x ^ y) - y);
}
```

leo60228:
`(float) (unsigned int) x` is `(float)((long)x & 0xffffffff)`. The assembly produced differs slightly depending on g++ version, but this does not affect the decompilation.

leo60228:
The 2.2 results seem to be a misdecompilation from Ghidra. No casts between integers and floats appear in the disassembly, only the decompilation. I'm guessing that Ghidra is decompiling an `&` on the binary representation of a float as an `&` on that float casted to an integer, despite that not being correct.
Anyway, looking only at the disassembly, 2.2 has the same behavior as `SDL_fabsf`, implying that the deceleration differences are unrelated. A bisection may be necessary.

InfoTeddy:
Back to auditing all replacements of `SDL_abs()`: there are currently only 4 uses of `SDL_abs()` in the codebase, and all of them previously used libc `abs()` (which is `int abs(int)`), so there are no further problems here.

leo60228:
It's worth noting that while libc's `abs()` is not overloaded, `fabs()` is defined in both `tgmath.h` and `math.h`. The `math.h` version is not overloaded, while the `tgmath.h`.

Technically, this is actually a "type-generic macro," which is subtly different than an overload. While the details of this were compiler-specific in C99, in C11 this is implemented via `_Generic`.

PR #579, opened 2021-01-04
Clean up whitespace and control flow of musicclass::play()

InfoTeddy:
While working on #580, I noticed that `musicclass::play()` had some inconsistent whitespace, as well as having control flow that meant nesting the rest of the function in two unnecessary levels of indentation. And that other PR would be much more readable if this function was improved. So I thought I'd fix it up first.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [X] My changes may be used in a future commercial release of VVVVVV (for
  example, a 2.3 update on Steam for Windows/macOS/Linux)
- [X] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes

PR #580, opened 2021-01-04
Fix regression with playing the same track after it fades out

InfoTeddy:
With commit 48313169b68f8ef8e6f83f29830b789061219ffd (PR #453), @AllyTally added a single-case patch for a regression, instead of fixing it at its root cause.

In fact, that commit only fixes the music if Presenting VVVVVV is playing while exiting to the menu, not if you enter a level that plays Presenting VVVVVV - so it only fixes it going one way, and not going the other way around; neither fixing also all the other cases this could happen.

It doesn't, say, fix the case where you are exited to the menu automatically after collecting the last crewmate in the level (or if the level calls gamestate 1013 itself), which is what happens in my MIRA-VIU TAS video at the end, and which I noted in the description of that video ([link](https://www.youtube.com/watch?v=OYQO4ePbYW4&t=111)).

So, the problem here is that when `musicclass::play()` is called, it sees that `currentsong` is the same as its input, and decides that since the music is already playing, it shouldn't play the music again. Thus, the music fades out, and we get silence instead of the music playing again.

But I said this was a regression. Why didn't this happen in 2.2? Well, it's because of the fact that 2.2 sets `currentsong` to -1 (no music playing at all) immediately when starting a fadeout, and not when the fadeout completes (commit facb079b3597b380f876537523cd351a0e637b62, PR #316). As you can imagine, this discrepancy could lead to bugs, given that the game would think that music wasn't playing when in actuality it was, but fixing this bug could also break code that expected this wrong behavior. And in this case, it has.

So to properly fix the root cause of this, instead of naïvely single-case patching out every case that comes up randomly, in `musicclass::play()`, the function will now ignore if the input given is the same as `currentsong` if the music is currently fading out.

Depends on #579.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2319,9 +2319,7 @@ void mapmenuactionpress()
         //This fixes an apparent frame flicker.
         FillRect(graphics.tempBuffer, 0x000000);
         graphics.fademode = 2;
-        if (music.currentsong != 6) {
-            music.fadeout();
-        }
+        music.fadeout();
         map.nexttowercolour();
         if (!game.glitchrunnermode)
         {
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -188,7 +188,7 @@ void musicclass::play(int t, const double position_sec /*= 0.0*/, const int fade
 	safeToProcessMusic = true;
 	musicVolume = MIX_MAX_VOLUME;
 
-	if (currentsong == t)
+	if (currentsong == t && Mix_PlayingMusic() != MIX_FADING_OUT)
 	{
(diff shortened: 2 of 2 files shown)


flibitijibibo:
Fair enough... for the record, I place 0.00% blame on Ally for doing an isolated patch, since the music system is so absurdly fragile - SDL_mixer definitely plays into that a bit, considering the music implementation has some really bizarre behavior internally (grep SDL_Delay). I still have vivid memories of fighting SDL_mixer for Proteus, an admittedly far more complex audio engine... yech.

InfoTeddy:
(Rebased and force-pushed to fix the merge conflict.)

PR #583, opened 2021-01-08
Move mapclass r/g/b variables off onto TowerBG

InfoTeddy:
This fixes a bug where if you entered a tower before watching the credits sequence, the credits sequence would have mismatched text and background colors.

Mismatched text and background colors on the credits screen

This bug happened because entering a tower modified the `r`/`g`/`b` attributes of mapclass, and updated `graphics.towerbg`, without updating `graphics.titlebg` too. Then `gamecompleterender()` uses the `r`/`g`/`b` attributes of mapclass, which were meant to work with `graphics.towerbg`, but draws `graphics.titlebg` instead.

The solution is to put the `r`/`g`/`b` attributes on TowerBG instead. That way, entering a tower will only modify the `r`/`g`/`b` attributes used to render towers, and won't affect the `r`/`g`/`b` attributes used to render the credits sequence.

Additionally, I also de-duplicated the case-switch that updated the `r`/`g`/`b` attributes based off of the current `colstate`, because it got copy-pasted twice, leading to three instances of one piece of code.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Map.cpp
+++ b/desktop_version/src/Map.cpp
@@ -13,9 +13,6 @@
 mapclass::mapclass()
 {
 	//Start here!
-	r = 196;
-	g = 196;
-	b = 196;
 	colstatedelay = 0;
 	colsuperstate = 0;
 	spikeleveltop = 0;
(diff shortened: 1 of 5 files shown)


TerryCavanagh:
Ah, cool! Always wondered why this happened!

InfoTeddy:
@TerryCavanagh Actually, the specific case mentioned in this PR here is a regression from another PR I did that fixed some tower background/title screen background rendering glitches (#522). This specific case didn't happen in 2.2, or previously as far as I know (I tested 2.2 and confirmed it didn't happen in that version).

TerryCavanagh:
Oh, hmm - I guess I’m thinking of a different colour mismatch bug 🤔

InfoTeddy:
Maybe the one that #111 fixed? Btw, #522 was meant as a generalized fix for every possible color mismatch bug (it's the PR that introduced the TowerBG object).

PR #589, opened 2021-01-10
Clean up unnecessary exports and add static keywords

InfoTeddy:
This patch cleans up unnecessary exports from header files (there were only a few), as well as adds the `static` keyword to all symbols that aren't exported and are specific to a file. This helps the linker out in not doing any unnecessary work, speeding it up and avoiding silent symbol conflicts (otherwise two symbols with the same name (and type/signature in C++) would quietly resolve as okay by the linker).

## Legal Stuff:

  for these changes


flibitijibibo:
Man I really blew it with the filesystem and controller declarations, didn't I...

PR #591, opened 2021-01-11
Remove game.savemystats

InfoTeddy:
This variable seems to have been intended to make sure `game.savestatsandsettings()` was called at the end of the frame, or make sure that it didn't get called more than once per frame. I don't see any frame ordering-related reason why it needs to be called specifically at the end of the frame (the function doesn't modify any state), so it's more plausible that it was added to make sure it didn't get called more than one per frame.

However, upon further analysis, none of the code paths where `game.savemystats` is used ever calls or sets `game.savemystats` more than once, and a majority of the code directly calls `game.savestatsandsettings()` anyway, so there's no reason for this variable to exist. If we ever need to make sure it doesn't get called more than once, and there's no way to change the code paths around to prevent it otherwise, we can use the defer callbacks system that I added to #535, when it gets merged.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -130,7 +130,6 @@ void Game::init(void)
     roomchange = false;
 
 
-    savemystats = false;
     quickrestartkludge = false;
 
     tapleft = 0;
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Makes sense, we do need to be careful with frequent writes as excessive disk activity can be a really bad cert issue, but this probably didn't come close to covering everything.

PR #595, opened 2021-01-11
Use pragma once everywhere

N00byKing:
## Changes:

Replace header guards with pragma once.
All "reasonable" compilers should support it (GCC support goes back as far as 3.4)

However, I didn't touch anything inside third_party

## Legal Stuff:

  for these changes


flibitijibibo:
This doesn't really fix any bugs and only adds potential to _add_ bugs (particularly for those building VVV for really old systems), so unfortunately we cannot merge this in.

PR #596, opened 2021-01-12
Reduce dependency on libc functions

InfoTeddy:
During 2.3 development, there's been a gradual shift to using SDL stdlib functions instead of libc functions, but there are still some libc functions (or the same libc function but from the STL) in the code.

Well, this patch replaces all the rest of them in one fell swoop.

SDL's stdlib can replace most of these, but its `SDL_min()` and `SDL_max()` are inadequate - they aren't really functions, they're more like macros with a nasty penchant for double-evaluation. So I just made my own `VVV_min()` and `VVV_max()` functions and placed them in `Maths.h` instead, then replaced all the previous usages of `min()`, `max()`, `std::min()`, `std::max()`, `SDL_min()`, and `SDL_max()` with `VVV_min()` and `VVV_max()`.

Additionally, there's no `SDL_isxdigit()`, so I just implemented my own `VVV_isxdigit()`.

There's one usage of `min()` that I simply removed - it was the generation of a random number between 0 and 12 that got capped to 11 with `min()` (because otherwise 12 would index a vector out-of-bounds, or at least before I added bounds checks to all tile drawing functions). I opted to simply remove the `min()` and change the range to be between 0 and 11 instead. The random range is effectively the same, but the distribution of random numbers won't be - however, I don't think the distribution of randomness matters here.

SDL has `SDL_malloc()` and `SDL_free()`, but they have some refcounting built in to them, so in order to use them with LodePNG, I have to replace the `malloc()` and `free()` that LodePNG uses. Which isn't too hard, I did it in a new file called `ThirdPartyDeps.c`, and LodePNG is now compiled with the `LODEPNG_NO_COMPILE_ALLOCATORS` definition.

Lastly, I also refactored the awful `strcpy()` and `strcat()` usages in `PLATFORM_migrateSaveData()` to use `SDL_snprintf()` instead. I know save migration is getting axed in 2.4, but it still bothers me to have something like that in the codebase otherwise.

Without further ado, here is the full list of functions that the codebase now uses:

- `SDL_strlcpy()` instead of `strcpy()`
- `SDL_strlcat()` instead of `strcat()`
- `SDL_snprintf()` instead of `sprintf()`, `strcpy()`, or `strcat()` (see above)
- `VVV_min()` instead of `min()`, `std::min()`, or `SDL_min()`
- `VVV_max()` instead of `max()`, `std::max()`, or `SDL_max()`
- `VVV_isxdigit()` instead of `isxdigit()`
- `SDL_strcmp()` instead of `strcmp()`
- `SDL_strcasecmp()` instead of `strcasecmp()` or Win32 `strcmpi()`
- `SDL_strstr()` instead of `strstr()`
- `SDL_strlen()` instead of `strlen()`
- `SDL_sscanf()` instead of `sscanf()`
- `SDL_getenv()` instead of `getenv()`
- `SDL_malloc()` instead of `malloc()` (replacing in LodePNG as well)
- `SDL_free()` instead of `free()` (replacing in LodePNG as well)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/ThirdPartyDeps.c
+++ b/desktop_version/src/ThirdPartyDeps.c
@@ -1,0 +1,18 @@
+#include <SDL_stdinc.h>
+
+// Handle third-party dependencies' needs here
+
+void* lodepng_malloc(size_t size)
+{
+    return SDL_malloc(size);
+}
+
+void* lodepng_realloc(void* ptr, size_t new_size)
+{
+    return SDL_realloc(ptr, new_size);
+}
+
+void lodepng_free(void* ptr)
+{
+    SDL_free(ptr);
+}
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -117,6 +117,7 @@ SET(VVV_SRC
 	src/XMLUtils.cpp
 	src/main.cpp
 	src/Network.c
+	src/ThirdPartyDeps.c
 )
 IF(NOT CUSTOM_LEVEL_SUPPORT STREQUAL "DISABLED")
 	LIST(APPEND VVV_SRC src/editor.cpp)
@@ -231,6 +232,8 @@ SET_PROPERTY(TARGET VVVVVV PROPERTY CXX_EXTENSIONS FALSE)
 ADD_LIBRARY(tinyxml2-static STATIC ${XML2_SRC})
 ADD_LIBRARY(physfs-static STATIC ${PFS_SRC} ${PFSP_SRC})
 ADD_LIBRARY(lodepng-static STATIC ${PNG_SRC})
+
+TARGET_COMPILE_DEFINITIONS(lodepng-static PRIVATE -DLODEPNG_NO_COMPILE_ALLOCATORS)
 
 # Static Dependencies
 TARGET_LINK_LIBRARIES(VVVVVV physfs-static tinyxml2-static lodepng-static)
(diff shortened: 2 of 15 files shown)


desktop_version/src/ThirdPartyDeps.c
```diff
@@ -0,0 +1,24 @@
+#include <stddef.h>
+
+// Handle third-party dependencies' needs here
+
+void* SDL_malloc(size_t);
```
flibitijibibo, line 5: I believe this will have issues due to calling conventions - it probably works coincidentally as a result of it being __cdecl implicitly for C. If we include SDL_stdinc.h it should bring these declarations in without being in too much (still a lot, but it shouldn't add to the build time).

flibitijibibo (changes requested):
Looks good other than the one little thing.

desktop_version/src/ThirdPartyDeps.c
InfoTeddy: Good point - initially I had included the entirety of `SDL.h` here, but moved it to forward declarations to reduce build time. But you do raise a good point. However, from what I understand, isn't it not recommended to include only one SDL sub-header, and you should instead just include all of `SDL.h`?

flibitijibibo: I think for some subheaders they depend on the monolithic SDL.h, but stdinc happens to be a core file that doesn't care about the others, so it should be safe to include directly (FAudio does this, for example).

InfoTeddy: Alright, fair enough. Rebased, fixed, and force-pushed.

PR #598, opened 2021-01-12
Update Version Check Method

N00byKing:
## Changes:

Using this Patch, there is no need for duplicating Version Information. Official (tagged) Builds will remain as-is, but no longer require hardcoding the Version String. Unofficial/Development builds will additionally Display:
- Base Version upon which Patches (may) have been applied
- Current Commit hash
- Whether or not the tree is dirty

If no changes have been made, the Build Date will still be shown in unofficial Builds to detect Compile-Time discrepancies when Troubleshooting.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -158,37 +158,31 @@ ELSE()
 	ADD_EXECUTABLE(VVVVVV ${VVV_SRC})
 ENDIF()
 
-IF(NOT OFFICIAL_BUILD)
-	# Add interim commit hash and its date to the build
+FIND_PACKAGE(Git REQUIRED)
 
-	# FIND_PACKAGE sets GIT_FOUND and GIT_EXECUTABLE
-	FIND_PACKAGE(Git)
+# These filenames have to be qualified, because when we run
+# the CMake script, its work dir gets set to the build folder
+SET(VERSION_INPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/Version.h.in)
+SET(VERSION_OUTPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/Version.h)
 
-	IF(GIT_FOUND)
-		# These filenames have to be qualified, because when we run
-		# the CMake script, its work dir gets set to the build folder
-		SET(VERSION_INPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/Version.h.in)
-		SET(VERSION_OUTPUT_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/Version.h.out)
+ADD_CUSTOM_TARGET(
+	GenerateVersion ALL
+	# This BYPRODUCTS line is required for this to be ran every time
+	BYPRODUCTS ${VERSION_OUTPUT_FILE}
+	COMMAND ${CMAKE_COMMAND}
+	# These args have to be passed through, otherwise the script can't see them
+	# Also, these args have to come BEFORE `-P`! (Otherwise it fails with an unclear error)
+	-DGIT_EXECUTABLE=${GIT_EXECUTABLE}
+	-DINPUT_FILE=${VERSION_INPUT_FILE}
+	-DOUTPUT_FILE=${VERSION_OUTPUT_FILE}
+	-P ${CMAKE_CURRENT_SOURCE_DIR}/version.cmake
+)
 
-		ADD_CUSTOM_TARGET(
-			GenerateVersion ALL
-			# This BYPRODUCTS line is required for this to be ran every time
-			BYPRODUCTS ${VERSION_OUTPUT_FILE}
-			COMMAND ${CMAKE_COMMAND}
-			# These args have to be passed through, otherwise the script can't see them
-			# Also, these args have to come BEFORE `-P`! (Otherwise it fails with an unclear error)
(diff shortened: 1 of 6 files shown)


N00byKing:
Looks like the CentOS target doesn't have Git installed...
If this Patch is useful i'll fix it, so just give me a nod to let me if it is/n't ;)

desktop_version/CMakeLists.txt
```diff
@@ -158,38 +158,32 @@ ELSE()
...
+	# These args have to be passed through, otherwise the script can't see them
+	# Also, these args have to come BEFORE `-P`! (Otherwise it fails with an unclear error)
+	-DGIT_EXECUTABLE=${GIT_EXECUTABLE}
+	-DINPUT_FILE=${VERSION_INPUT_FILE}
+	-DOUTPUT_FILE=${VERSION_OUTPUT_FILE}
+	-P ${CMAKE_CURRENT_SOURCE_DIR}/version.cmake
+)
+
+IF(OFFICIAL_BUILD)
+	ADD_DEFINITIONS(-DOFFICIAL_BUILD)
```
InfoTeddy, line 182: You completely misunderstand the `OFFICIAL_BUILD` CMake variable - it's not a compile-time definition. Nowhere in the code is there an `#ifdef OFFICIAL_BUILD`.

InfoTeddy (commented):
I'm not a fan of this PR - the version number is completely fine as it is (it only gets used once), and if we ever need to use it more than once, there are better ways than relying on the current Git tag. Maybe something like this in a header file:
```c++
const int VERSION_MAJOR = 2;
const int VERSION_MINOR = 3;
```
But anyways, this PR introduces a hard dependency on Git for anyone using the CMake file. There's no reason to do that - partly because if in the future we need to move to another version control software, we'd be able to more easily do so without requiring Git, but mostly because I see no point in doing so. And indeed, the CentOS target failed because it doesn't have Git installed (why would it need to?).

Also, it seems like you're German, and I just want to say that we don't capitalize every noun in English like you would in German. Furthermore, just like your last PR (#595), you forgot to include yourself in `CONTRIBUTORS.txt` and `Credits.h`.

desktop_version/CMakeLists.txt
N00byKing: The PR introduces one to decide whether or not to include the Commit Date; see Version.h.in

InfoTeddy: My bad then, I didn't see that.

N00byKing:
The PR does introduce a dependency on git, but I didn't think too much of it. Anyone who is going to compile it will be using git anyway, imo, and it won't be needed for running the Game. Switching the VCS is also unrealistic imho. Additionally, as you said it is only one place where a change would be needed if a different VCS would be used. In that regard it would be equivalent to the current situation, which also uses git.

Regarding the credit, I didn't want to be presumptuous, so in any project with a CONTRIBUTORS.md or similar I usually add the changes only when I know the patch will be accepted. Seems like it'll be unrealistic in this case, anyway

And yeah, my english is very much a work in progress :/ 
Always had problems with capitalization, even in German haha
Appreciate the help!

flibitijibibo:
The Git dependency is common but not 100% always there - the CentOS builder pretty much nails it. There are cases where builds are done from zipped tags which do not have the Git log available, and some may use alternative revision control locally and so they may not have the Git information (I run into this all the time, as MojoShader really prefers you have Hg which we do not use for FNA's copy of the repository).

N00byKing:
Hm, true.
I could add a Fallback Option which just prints "Unofficial Build" or something, would that help?

flibitijibibo:
Something like `2.3-UNKNOWN` is probably enough to go on - whatever's enough to make it clear that it's not my build nor a build that can be trivially identified (unless someone is intentionally malicious).

InfoTeddy:
I forgot to say this and test this earlier, but the version number printed would end up being something like `v2.2-1295-g769f99f`, not `v2.3`, since the printed number depends on the last tag in the repo.

And like I said earlier, I don't see any need to replace the hardcoded `v2.3` with anything else.

N00byKing:
That is intended. It says the last Tag, and commits built on top of it, instead of working version. After all, the build wasn't performed on v2.3, at least that is my reasoning

flibitijibibo:
For the moment we're going on a system where it's `2.3-${hash}` for indev builds, with final builds being simply `2.3`. It's a model vaguely similar to SDL's where they increment the version right after tagging, though instead of doing releases with every other version we just use the git hash to indicate that it's in-development. What we might be able to do is move to the alternating number system (so 2.4 would be the next indev while 2.5 would be the next tag) but for now all indev builds are `2.3-${hash}`

InfoTeddy:
> What we might be able to do is move to the alternating number system

That would actually fit with the pattern so far: 2.0 was a release, 2.1 was a bunch of beta builds that never got officially released, and 2.2 was a full release. Shame I didn't recognize it sooner and suggest it.

flibitijibibo:
Hypothetically it's still possible since we've yet to tag a release... if this is what we want to do I'm in favor of it since we don't know how long tags are going to be (this one's timeline was a freak case since it was the first after the source release).

flibitijibibo:
Apparently the feedback we got from users was that they absolutely did not want an alternating version number for indev and stable, so that pretty much killed this patch... it's not as automated as this patch tries but the system we have will do for now.

PR #600, opened 2021-01-14
Disable MSVC warnings about implicit conversion

InfoTeddy:
Implicit conversion warnings happen all over the codebase, but there's no reason to warn on all of them, and adding casts everywhere is annoying to read and patently unnecessary.

MSVC is the only compiler that has this warnings (GCC even on `-Wall -Wextra` doesn't warn about implicit conversions), so disable it for MSVC.

## Legal Stuff:

  for these changes


flibitijibibo:
Might be worth replacing fopen with SDL_RWops, but yeah the conversion warnings can get a bit excessive. Sometimes it finds good precision issues but it's pretty rare.

InfoTeddy:
Hm, yeah, we could use SDL_RWops instead. It would remove another libc dependency. That'll happen later though, so for now, I've just dropped the `fopen()` warning disabling from the PR (it only warns twice anyways).

PR #607, opened 2021-02-02
Fix physics regression with immediately flipping off of horizontally-moving platforms and conveyors, and re-fix glitchy y-position when spawning onto a conveyor

### LINKED ISSUE #606: Physics regression with y-position after flipping off of horizontally-moving platforms and conveyors
@anna328p on the VVVVVV Discord server reported being unable to do the Trench Warfare trick where you flip over the entire room in one flip. I was able to confirm her findings by creating a libTAS movie file that worked on 2.2 but desynced in 2.3.

The problem seems to be that when Viridian flips off of the horizontally-moving platform, their y-position is closer to the platform in 2.3 than in 2.2, and so they hit the wall and are unable to complete the flip in 2.3, while they completely miss it and complete the flip in 2.2.

2.2:
Trench Warfare trick in 2.2

2.3:
Trench Warfare trick in 2.3 (regression)

I heavily suspect this is caused by 942217f871bf1182ded3a915b8c26b90c04d561a (#509).

InfoTeddy:
This is a re-do of 942217f871bf1182ded3a915b8c26b90c04d561a (#509), but with a more conservative fix that only resets the player's `newxp` and `newyp` when they respawn from a checkpoint or spawn in to the map. The previous fix (of a regression of a fix) has a regression where immediately flipping off of horizontally-moving platforms or conveyors will no longer provide you with a "boost" given certain vertical pixel alignments.

Unlike the previous patch, if the player were to suddenly collide with a conveyor or horizontally-moving platform during gameplay, their y-position would revert back to the intended next y-position of the previous frame. But this is the same behavior as before, I haven't ever seen such a contrived situation come up, and this behavior is probably more preferable for gameplay than actually going to the conveyor, so it's fine.

I also decided to reset `newxp` here, and not just `newyp`, because while resetting `newyp` seems to be enough, it's safer to also reset `newxp` (and so future readers won't question why only `newyp` is reset but not `newxp`).

I tested this and it once again fixes the death loop issue from earlier, while also still allowing for that Trench Warfare trick to be possible (I tested it with the libTAS movie I mentioned in #606; it syncs fine). There are no other known regressions resulting from this fix (hopefully).

Fixes #606.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Logic.cpp
+++ b/desktop_version/src/Logic.cpp
@@ -826,7 +826,6 @@ void gamelogic()
                 if (INBOUNDS_VEC(i, obj.entities) && j > -1000)
                 {
                     obj.entities[i].newxp = obj.entities[i].xp + j;
-                    obj.entities[i].newyp = obj.entities[i].yp;
                     obj.entitymapcollision(i);
                 }
                 else
@@ -835,7 +834,6 @@ void gamelogic()
                     if (INBOUNDS_VEC(i, obj.entities) && j > -1000)
                     {
                         obj.entities[i].newxp = obj.entities[i].xp + j;
-                        obj.entities[i].newyp = obj.entities[i].yp;
                         obj.entitymapcollision(i);
                     }
                 }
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Excellent work on this, glad we were able to keep compat as well as stability!

PR #609, opened 2021-02-05
Split flip logic from player input logic

### LINKED ISSUE #484: Apply flip momentum to duplicate player entities

leo60228:
(this is a replacement for #485 since GitHub doesn't allow editing the repo of an existing PR)

Closes #484

## Changes:
Flipping only applies momentum to the player entity currently being processed. This normally wouldn't be a problem. However, flipping involves global state, and only one flip can occur per frame. This means that additional player entities don't get this boost of momentum, which feels somewhat unnatural during gameplay.
This commit fixes this by splitting flip logic out of the loop over player entities, and applying the flip momentum to all player entities.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2182,10 +2182,13 @@ void gameinput(void)
                 }
             }
 
-            if (obj.entities[ie].onground > 0) {
-              any_onground = true;
-            } else if (obj.entities[ie].onroof > 0) {
-              any_onroof = true;
+            if (obj.entities[ie].onground > 0)
+            {
+                any_onground = true;
+            }
+            else if (obj.entities[ie].onroof > 0)
+            {
+                any_onroof = true;
             }
         }
     }
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2180,95 +2180,95 @@ void gameinput(void)
                 }
             }
         }
+    }
 
-        if (game.press_left)
+    if (game.press_left)
+    {
+        game.tapleft++;
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Seems fine to me... Any reason this is marked as draft (other than being deferred to 2.4)?

leo60228:
https://github.com/TerryCavanagh/VVVVVV/pull/485#issuecomment-762403932

InfoTeddy:
Got it - so it's okay to merge, then?

leo60228:
I assume so, but flibit was the one who marked it as a draft in the first place.

leo60228:
Actually, let me at least fix the Input.cpp indentation.

flibitijibibo:
Oh, I might have done that when we started the feature freeze. Reviewing now...

flibitijibibo (approved):
Still makes sense, so as long as it's tested we're good to go.

leo60228:
Haven't tested the rebase, but it builds and it worked before.

leo60228:
Tested and working.

leo60228:
Wait, I meant to send that on a different PR. I was going to test this one next...

leo60228:
Tested and working.

PR #610, opened 2021-02-07
Miscellaneous cleanups with UtilityClass.cpp and some of editor.cpp

InfoTeddy:
This is just a general grab-bag of cleanups I did after refactoring `ss_toi()` to not use stringstreams.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/UtilityClass.cpp
+++ b/desktop_version/src/UtilityClass.cpp
@@ -36,12 +36,41 @@ static const char* GCChar(const SDL_GameControllerButton button)
 	}
 }
 
-int ss_toi( std::string _s )
+int ss_toi(const std::string& str)
 {
-	std::istringstream i(_s);
-	int x = 0;
-	i >> x;
-	return x;
+	int retval = 0;
+	int place = 1;
+	bool negative = false;
+	const int radix = 10;
+
+	for (size_t i = 0; i < str.size(); ++i)
+	{
+		const char chr = str[i];
+
+		if (i == 0 && chr == '-')
+		{
+			negative = true;
+			continue;
+		}
+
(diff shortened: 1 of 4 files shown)


desktop_version/src/UtilityClass.cpp
```diff
@@ -1,67 +1,63 @@
...
+	default:
+		SDL_assert(0 && "Unhandled button!");
+		return NULL;
 	}
-	SDL_assert(0 && "Unhandled button!");
-	return NULL;
 }
 
-int ss_toi( std::string _s )
+int ss_toi(const std::string& str)
```
flibitijibibo, line 39: This implementation seems incomplete - should we worry about `-` as well?

desktop_version/src/UtilityClass.cpp
```diff
@@ -1,67 +1,63 @@
 #define HELP_DEFINITION
 #include "UtilityClass.h"
 
-#include <cctype>
 #include <SDL.h>
 #include <sstream>
 
-/* Used by UtilityClass::GCString to generate a button list */
```
flibitijibibo, line 8: Might be a good comment to keep, but we're probably dumping a lot of this when input gets redone anyway...

desktop_version/src/UtilityClass.cpp
InfoTeddy: Good catch - almost forgot to add that. Added, rebased, and force-pushed.

desktop_version/src/UtilityClass.cpp
InfoTeddy: Yeah, it'll probably get axed anyway. If you really want me to keep this, I can; otherwise I don't see the point.

PR #613, opened 2021-02-12
Clean up is_number(), is_positive_num(), and VVV_isxdigit(), and fix some of their bugs

InfoTeddy:
While working on an unrelated patchset, I noticed that these two functions could use some work. Then I worked on that other patchset further and it turned out that it didn't need to depend on any of the changes I was planning to make to these functions. So I've split these changes off into their own PR, which is this PR.

`is_positive_num()` no longer uses an `std::string` and instead uses a `const char*`, along with its `hex` argument becoming const. Code style of both `is_positive_num()` and `is_number()` has also been cleaned up.

But the biggest thing is probably the bug fixes; `is_positive_num()` and `is_number()` no longer accept empty strings as numbers, and `is_number()` no longer accepts the string `"-"` as a number, either. `VVV_isxdigit()` no longer accepts the entire alphabet as being hexadecimal digits.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/UtilityClass.cpp
+++ b/desktop_version/src/UtilityClass.cpp
@@ -272,7 +272,7 @@ bool is_number(const char* str)
 		return false;
 	}
 
-	for (size_t i = 1; str[i] != '\0'; i++)
+	for (size_t i = 1; str[i] != '\0'; ++i)
 	{
 		if (!SDL_isdigit(str[i]))
 		{
--- a/desktop_version/src/UtilityClass.cpp
+++ b/desktop_version/src/UtilityClass.cpp
@@ -272,7 +272,7 @@ bool is_number(const char* str)
 		return false;
 	}
 
-	for (int i = 1; str[i] != '\0'; i++)
+	for (size_t i = 1; str[i] != '\0'; i++)
 	{
 		if (!SDL_isdigit(str[i]))
 		{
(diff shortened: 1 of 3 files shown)


desktop_version/src/UtilityClass.cpp
```diff
@@ -225,13 +225,20 @@ void UtilityClass::updateglow()
...
+	if (str[0] == '\0')
 	{
-		if (!SDL_isdigit(static_cast<unsigned char>(str[i])) && (i != 0 || str[0] != '-'))
+		return false;
+	}
+
+	for (size_t i = 0; str[i] != '\0'; ++i)
+	{
+		if (!SDL_isdigit(str[i])
+		&& (i != 0 || str[0] != '-' || str[1] == '\0'))
```
flibitijibibo, line 236: Do we just want to unroll iterations 0 and 1? It took me a few tries to figure out even the original loop check...

flibitijibibo (changes requested):
Otherwise good, that loop though... it wasn't good to begin with so I can't blame you on that one >_>

desktop_version/src/UtilityClass.cpp
InfoTeddy: Yeah, we can unroll it. Just give me a second...

InfoTeddy: Took me a bit, but I think I've figured it out. Rebased and force-pushed. (Also fixed a bug with `VVV_isxdigit()`.)

flibitijibibo:
Latest is good! Also I just realized isxdigit was base 36 until just now, lolll

PR #614, opened 2021-02-13
Refactor loading arrays from XML to not use the STL

InfoTeddy:
The current way "arrays" from XML files are loaded (before this PR is applied) goes something like this:

1. Read the buffer of the contents of the tag using TinyXML-2.

2. Allocate a buffer on the heap of the same size, and copy the existing buffer to it. (This is what the statement `std::string TextString = pText;` does.)

3. For each delimiter in the heap-allocated buffer...

   a. Allocate another buffer on the heap, and copy the characters from the previous delimiter to the delimiter you just hit.

   b. Then allocate the buffer *again*, to copy it into an std::vector.

4. Then re-allocate every single buffer *yet again*, because you need to make a copy of the `std::vector` in `split()` to return it to the caller.

As you can see, the existing way uses a lot of memory allocations and data marshalling, just to split some text.

The problem here is mostly making a temporary `std::vector` of split text, before doing any actual useful work (most likely, putting it into an array or *another* `std::vector` - if the latter, then that's yet another memory allocation on top of the memory allocation you already did; this memory allocation is unavoidable, unlike the ones mentioned earlier, which should be removed).

So I noticed that since we're iterating over the entire string once (just to shove its contents into a temporary `std::vector`), and then basically iterating over it again - why can't the whole thing just be more immediate, and just be iterated over once?

So that's what I've done here. I've axed the `split()` function (both of them, actually), and made `next_split()` and `next_split_s()`.

`next_split()` will take an existing string and a starting index, and it will find the next occurrence of the given delimiter in the string. Once it does so, it will return the length from the previous starting index, and modify your starting index as well. The price for immediateness is that you're supposed to handle keeping the index of the previous starting index around in order to be able to use the function; updating it after each iteration is also your responsibility.

(By the way, `next_split()` doesn't use `SDL_strchr()`, because we can't get the length of the substring for the last substring. We could handle this special case specifically, but it'd be uglier; it also introduces iterating over the last substring twice, when we only need to do it once.)

`next_split_s()` does the same thing as `next_split()`, except it will copy the resulting substring into a buffer that you provide (along with its size). Useful if you don't particularly care about the length of the substring.

All callers have been updated accordingly. This new system does not make *any* heap allocations at all; at worst, it allocates a temporary buffer on the stack, but that's only if you use `next_split_s()`; plus, it'd be a fixed-size buffer, and stack allocations are negligible anyway.

This improves performance when loading any sort of XML file, especially loading custom levels - which, on my system at least, I can noticeably tell (there's less of a freeze when I load in to a custom level with lots of scripts). It also decreases memory usage, because the heap isn't being used just to iterate over some delimiters when XML files are loaded.

Also some cleanups, because there's lots of related code to tidy up that I've decided to tidy up in this PR.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -442,6 +442,22 @@ void Game::updatecustomlevelstats(std::string clevel, int cscore)
     savecustomlevelstats();
 }
 
+#define LOAD_ARRAY_RENAME(ARRAY_NAME, DEST) \
+    if (pKey == #ARRAY_NAME) \
+    { \
+        std::string TextString = pText; \
+        if (TextString.length()) \
+        { \
+            std::vector<std::string> values = split(TextString, ','); \
+            for (int i = 0; i < VVV_min(SDL_arraysize(DEST), values.size()); i++) \
+            { \
+                DEST[i] = help.Int(values[i].c_str()); \
+            } \
+        } \
+    }
+
+#define LOAD_ARRAY(ARRAY_NAME) LOAD_ARRAY_RENAME(ARRAY_NAME, ARRAY_NAME)
+
 void Game::loadcustomlevelstats()
 {
     //testing
@@ -4420,22 +4436,6 @@ void Game::unlocknum( int t )
     savestatsandsettings();
 #endif
 }
-
-#define LOAD_ARRAY_RENAME(ARRAY_NAME, DEST) \
-    if (pKey == #ARRAY_NAME) \
-    { \
-        std::string TextString = pText; \
-        if (TextString.length()) \
-        { \
-            std::vector<std::string> values = split(TextString, ','); \
-            for (int i = 0; i < VVV_min(SDL_arraysize(DEST), values.size()); i++) \
-            { \
-                DEST[i] = help.Int(values[i].c_str()); \
-            } \
-        } \
-    }
-
-#define LOAD_ARRAY(ARRAY_NAME) LOAD_ARRAY_RENAME(ARRAY_NAME, ARRAY_NAME)
 
 void Game::loadstats(ScreenSettings* screen_settings)
 {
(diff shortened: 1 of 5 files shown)


desktop_version/src/Game.cpp
```diff
@@ -442,6 +442,27 @@ void Game::updatecustomlevelstats(std::string clevel, int cscore)
     savecustomlevelstats();
 }
 
+#define LOAD_ARRAY_RENAME(ARRAY_NAME, DEST) \
+    if (pKey == #ARRAY_NAME && pText[0] != '\0') \
+    { \
+        char buffer[16]; \
```
flibitijibibo, line 448: Could use a comment to explain why 16 is used

desktop_version/src/UtilityClass.cpp
```diff
@@ -71,21 +73,61 @@ int ss_toi(const std::string& str)
...
+		{
+			++*start;
+			return true;
+		}
+		else if (str[idx] == '\0')
+		{
+			return true;
+		}
+
+		++idx;
```
flibitijibibo, line 102: The ++ use here and slightly above threw me off for a second... I know it's against style everywhere else but += 1 might be easier to read for the pointer derefs in particular.

flibitijibibo (changes requested):
Otherwise looks good, the load time improvement is great!

desktop_version/src/Game.cpp
InfoTeddy: Done. Rebased and force-pushed.

desktop_version/src/UtilityClass.cpp
InfoTeddy: Done. Rebased and force-pushed.

PR #616, opened 2021-02-16
Fix all memory errors reported by Valgrind (including some 2.3 regressions)

InfoTeddy:
The other day, I ran this game under Valgrind and it reported a ton of stuff. Some of it was memory that wasn't really leaked but which we should clean up on program exit anyway just so Valgrind is useful, and most of the rest of it is 2.3 regressions that didn't exist in 2.2.

Here are all the resources which weren't technically being leaked (because they can't get leaked more than once), but which Valgrind reports, and is now properly cleaned up by the time of the game exiting:

- All `Screen` resources
- All `GraphicsResources` resources
- All `Graphics` resources (tile vectors and buffers)
- All `musicclass` resources
- The `musicWriteBlob` that writes a `BinaryMusic.vvv` to disk (if `VVV_COMPILEMUSIC` is defined)

Here are all the memory leaks that are fixed:

- Leaking memory if an `mmmmmm.vvv` file exists while assets are mounted or unmounted, which is a 2.3 regression

  (@leo60228 earlier fixed a memory leak for the case of only PPPPPP existing, but not if *both* PPPPPP and MMMMMM exist - which is slightly disappointing for someone who primarily programs in a language whose entire point of existing at all is based around a compiler that forces you to correctly manage your memory; you would expect this to facilitate better reasoning about memory management, but I guess single-case patching can affect anyone, even if they are a Rust developer; single-case patching strikes again!)

- Leaking memory every single time a file is loaded when the levels list is loaded, another 2.3 regression

- Leaking memory for every `SDL_RWops` allocated in `MusicTrack::MusicTrack()` (this is only really a leak if custom assets are mounted or unmounted)

There is also a reading of uninitialized memory when the levels list gets created, which Valgrind reported, and which I've also fixed.

All exit paths of the game have been cleaned up to use common code (so it is guaranteed that the game will attempt to write `unlock.vvv` and `settings.vvv` if it managed to exit gracefully). Lastly, there are cleanups in other various parts of the code; view the commits for more details. Actually, you should always view the commits for more details; I'm the only one around here who actually writes something useful in their commit descriptions.

After this PR is merged, the only thing that `valgrind --leak-check=full ./VVVVVV` will ever complain about (on my system, at least) is something about 8 bytes of memory being leaked by PulseAudio. This is obviously not us, it's only 8 bytes, and PulseAudio is stupid anyways; so we don't need to worry about that.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -183,8 +183,8 @@ std::vector<int> binaryBlob::getExtra()
 	for (size_t i = 0; i < SDL_arraysize(m_headers); i += 1)
 	{
 		if (m_headers[i].valid
-#define FOREACH_TRACK(track_name) && SDL_strcmp(m_headers[i].name, track_name) != 0
-		TRACK_NAMES
+#define FOREACH_TRACK(_, track_name) && SDL_strcmp(m_headers[i].name, track_name) != 0
+		TRACK_NAMES(_)
 #undef FOREACH_TRACK
 		) {
 			result.push_back(i);
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -3,8 +3,8 @@
 #include <physfs.h> /* FIXME: Abstract to FileSystemUtils! */
 #include <SDL.h>
 #include <stdio.h>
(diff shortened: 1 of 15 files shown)


flibitijibibo:
Here's hoping Pipewire is better!

PR #623, opened 2021-02-20
Use SDL_OpenURL in FILESYSTEM_openDirectory

leo60228:
## Changes:
The recently released SDL 2.0.14 adds a native function for opening URIs from the host system, superseding the OS-specific implementations of FILESYSTEM_openDirectory.

This has only been tested on Linux. It should be tested on other platforms too.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -565,8 +565,9 @@ bool FILESYSTEM_openDirectory(const char *dname)
 {
 	char url[MAX_PATH];
 	SDL_snprintf(url, sizeof(url), "file://%s", dname);
-	if (SDL_OpenURL(url))
+	if (SDL_OpenURL(url) == -1)
 	{
+		printf("Error opening directory: %s\n", SDL_GetError());
 		return false;
 	}
 	return true;
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -16,15 +16,11 @@
 #if defined(_WIN32)
 #include <windows.h>
 #include <shlobj.h>
-#include <shellapi.h>
 #elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__HAIKU__) || defined(__DragonFly__)
 #include <unistd.h>
 #include <dirent.h>
 #include <limits.h>
 #include <sys/stat.h>
-#include <sys/types.h>
-#include <sys/wait.h>
-#include <spawn.h>
 #define MAX_PATH PATH_MAX
 #endif
 
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -561,42 +561,16 @@ bool FILESYSTEM_openDirectoryEnabled()
 	return !SDL_GetHintBoolean("SteamTenfoot", SDL_FALSE);
 }
 
-#ifdef _WIN32
 bool FILESYSTEM_openDirectory(const char *dname)
 {
-	ShellExecute(NULL, "open", dname, NULL, NULL, SW_SHOWMINIMIZED);
+	char url[MAX_PATH];
+	SDL_snprintf(url, sizeof(url), "file://%s", dname);
+	if (SDL_OpenURL(url))
+	{
+		return false;
+	}
 	return true;
 }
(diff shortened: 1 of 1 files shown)


InfoTeddy (changes requested):
The includes at the top of the file should be cleaned up, now that we no longer do the implementation ourselves.

In 6847eb3a8706f11d9bd9651e1e43b47db3da4f1c I added Win32 `<shellapi.h>`, and *nix `<sys/types.h>`, `<sys/wait.h>`, and `<spawn.h>`. Ethan for some reason added *nix `<limits.h>` and `<sys/stat.h>` in f815b1ee62101a803a526f136ad4e13a92e4b273 even though that commit ought to have *removed* includes from the file, not add them?

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -561,42 +561,16 @@ bool FILESYSTEM_openDirectoryEnabled()
 	return !SDL_GetHintBoolean("SteamTenfoot", SDL_FALSE);
 }
 
-#ifdef _WIN32
 bool FILESYSTEM_openDirectory(const char *dname)
 {
-	ShellExecute(NULL, "open", dname, NULL, NULL, SW_SHOWMINIMIZED);
+	char url[MAX_PATH];
+	SDL_snprintf(url, sizeof(url), "file://%s", dname);
```
InfoTeddy, line 567: Why is there only one slash? Shouldn't it be `file:///`?

I tried `xdg-open file://` (which should be what `SDL_OpenURL()` ultimately ends up doing on my system), it said `xdg-open: file 'file://' does not exist`; but `xdg-open file:///` worked.

InfoTeddy (changes requested):

desktop_version/src/FileSystemUtils.cpp
leo60228: `FILESYSTEM_openDirectory` is called with a path. With your change, it would be `file:////home/leo60228/.local/share/VVVVVV/levels` on my system, which is one too many slashes.

leo60228: On Windows, `file://C:\Users\leo60228\Documents\VVVVVV` is *presumably* correct.

InfoTeddy: Having duplicate slashes on *nix is completely fine (`xdg-open file:////home/` works for me). I'm not sure about how many slashes Windows accepts, though.

leo60228: I'm not sure what your point is, then. Two slashes in the format string seems to be correct in all cases, even if more might work.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -561,42 +561,16 @@ bool FILESYSTEM_openDirectoryEnabled()
...
+	if (SDL_OpenURL(url))
+	{
+		return false;
+	}
 	return true;
```
InfoTeddy, line 572: 1. It is true that `SDL_OpenURL()` returns -1 (a nonzero value) upon failure; however treating it directly like a boolean in a conditional is a bit obtuse. So, it ought to be `SDL_OpenURL(url) == -1`. However...
2. This construct can be simplified to `return SDL_OpenURL(url) == -1`. However...
3. We ought to print what `SDL_GetError()` gives us if `SDL_OpenURL()` fails.

InfoTeddy (changes requested):

desktop_version/src/FileSystemUtils.cpp
InfoTeddy: Well, I don't know if it's correct on Windows. But I guess that's already addressed by you yourself admitting this needs to be tested on other platforms.

desktop_version/src/FileSystemUtils.cpp
leo60228: I considered that, but I was mainly trying to be consistent with the rest of `FileSystemUtils.cpp`. I'll change it, though.

leo60228:
> Ethan for some reason added *nix `<limits.h>` and `<sys/stat.h>` in f815b1e even though that commit ought to have removed includes from the file, not add them?

They were there before, that commit just reordered them.

InfoTeddy:
> They were there before, that commit just reordered them.

I see, I misunderstood, then.

InfoTeddy:
I can confirm that this works on Windows.

PR #624, opened 2021-02-20
Decouple editor code and custom level code

InfoTeddy:
So, it seems like what happened in 2.0 was that Terry or Simon wanted to add custom level support, and started by working on an editor, in `editor.cpp`. Unfortunately, they kind of just kept bolting things on to the file, and eventually that file not only contained code for the editor, but also the entirety of custom level support.

While I'd rather have it so there's no difference between main game levels and custom levels, it's pretty clear that the current state of affairs is bad. And to that end, I've renamed `editorclass` to `customlevelclass` and separated the editor-specific attributes to a new `editorclass`. (The name of the instance of `customlevelclass` is `cl`.)

I've also gone ahead and renamed some other things - the `edentities` class is now named `CustomEntity`, the `edentity` array of *instances* of the class is now named `customentities` (yeah, reversing the singular/plural names is really stupid); the `edlevelclass` class is now named `RoomProperty`, the `level` array of instances of the class is now named `roomproperties`. I've also fixed up some other things here and there, such as ifdefs of `NO_CUSTOM_LEVELS` and `NO_EDITOR`.

## Legal Stuff:

  for these changes


flibitijibibo:
The heck... apparently there's still some algorithm stuff in editor. But, a broad review:

The reasoning makes sense to me - though I will admit, 5000 isn't massive to me, but that may come from a combination of commercial game experience (current record is 45,000 lines (no, seriously)) and the Vulkan renderer I've been working on where the file has been hanging around 11,000 lines for quite a while.

Personal bias aside, the part where I'm not totally sure is the naming itself. The file split-up itself? Totally got it. The function names are what have me wondering... should we fully commit and fix the names to refer to custom level logic? When I saw the patchset I thought "oh, it's like [FAudio's tool UI](https://github.com/FNA-XNA/FAudio/blob/master/utils/uicommon/FAudioUI_main.cpp#L48-L55) where we have a common UI loop and let tools implement logic separately," but in this case it seems like the issue was that custom levels and the _editor_ for custom levels were overly coupled, both in compile units _and_ what runs when. If that's the real issue, we should rip this band-aid off straight away and rename that part of the code accordingly. I could be misreading this though; much like the gameplay logic I didn't see very much of the editor when working on 2.2.

flibitijibibo:
Oh also: I'm not too worried about git history... we can always run git blame on a truncated history (which I actually do a lot with FNA, due to its decade+ history).

InfoTeddy:
> apparently there's still some algorithm stuff in editor

Oh, yeah there is. After you merged that other PR, I rebased this one and pre-emptively removed it from `editor.cpp` again. But apparently it's still used there.

> it seems like the issue was that custom levels and the *editor* for custom levels were overly coupled, both in compile units *and* what runs when

Yep, you hit the nail on the head here. That's exactly what's going on. So... what exactly is your recommendation? Is it enough to just change this PR to rename the original file `CustomLevels.cpp` and the new file `Editor.cpp`?

flibitijibibo:
Let's also change the function prefixes as well - something like `s/editor/customlevel/g` for functions is fine with me. (If it feels cleaner to make separate commits that's fine too!)

InfoTeddy:
:+1:

It's a bit more than just the function prefixes, there's several classes and such (and I forgot about the header files too!); this might take a while...

PR #625, opened 2021-02-20
Remove unneeded <algorithm> includes

InfoTeddy:
One of these days, I need to get around to running Include What You Use on this codebase. Until then, while I was working on #624, I noticed these; I'm removing them now.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -1,4 +1,3 @@
-#include <algorithm>
 #include <iostream>
 #include <iterator>
 #include <physfs.h>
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Looks like this quietly does something on CentOS, we'll see how the other builds do here...

InfoTeddy:
Right - it was one of those evil transitive includes. I added it to `editor.cpp` (it'll be removed from there and put in `editorloop.cpp` in #624 anyway).

flibitijibibo:
Cool, we'll start with this because any chance to nuke `<algorithm>` is a good commit in my book...

Issue #618, opened 2021-02-16
SDL in Ubuntu stable releases is too old


Dav999-v:
These commits added the use of `SDL_zeroa()`:
70b9ffe6a58b71122aa60280bae4194326d026a0
a113662050d29a01819a2f606c9bb1d55d941abb
bd97378862c6b01d1de59f59a98a8a53a74fb696

The problem is that `SDL_zeroa()` was [added in SDL 2.0.12](https://github.com/libsdl-org/SDL/blob/983bbf9ef3e572a073a6f5877faf1c0b4803527c/WhatsNew.txt), and the earliest version of Ubuntu that can easily install SDL 2.0.12 via `apt` is the very latest, Ubuntu 20.10 (see [20.04](https://packages.ubuntu.com/focal/libsdl2-dev) and [20.10](https://packages.ubuntu.com/groovy/libsdl2-dev)), and 20.10 is not an LTS version.

So now anyone on most versions of Ubuntu or related distros (or anywhere else an earlier SDL is used than March 2020) will get a compilation error and will have to be told their SDL is too old. This makes "just" compiling VVVVVV more difficult on those systems.

flibitijibibo:
Has a report been filed to Canonical to update to 2.0.14? If we can fix it on that end instead that'd be preferred, as we'll eventually be moving to 2.0.16 for some stdlib functionality anyhow.

Dav999-v:
I don't know about any method to file a report to Canonical or where to find out whether anyone has done it, but in my experience it's pretty standard that they keep packages on whichever version was latest before the original date of that Ubuntu release (unless required for security), and only newer releases get newer versions.

leo60228:
Packages will only be backported into Ubuntu itself due to major bugs, however users can opt-in to updating specific packages through the Ubuntu Backports Project.

`requestbackport` in the `ubuntu-dev-tools` package is an interactive tool for submitting backport requests with the required metadata.

dsalt:
I have SDL 2.0.14 installed here, and I doubt that I'll be updating past that before Debian's next release (though I'm actually using Devuan) – and if I do it'll most likely be via backports.

The only times that I've seen newer versions being pulled into an existing distribution release are to pull in some security bug fixes but the fixes are too big or too invasive for the maintainers to handle as patches.

(Incidentally, Debian's previous release, buster, has SDL 2.0.9.)

[...]

PR #626, opened 2021-02-20
Clarify build requirements

leo60228:
## Changes:

This clarifies the build requirements, as a partial fix for #618.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/README.md
+++ b/desktop_version/README.md
@@ -1,21 +1,32 @@
-How to Build
-------------
-VVVVVV's official desktop versions are built with the following environments:
+# How to Build
 
-- Windows: Visual Studio 2010
+## Requirements
+
+The current version of VVVVVV should be buildable with most modern compilers.
+Official releases use the following:
+
+- Windows: Visual Studio 2010 on Windows 7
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Woops, meant to review this earlier...

This changes the whole file instead of adding the version number to the SDL2 line. Adding the version is an okay fix in my book, but the formatting is a whole other issue.

InfoTeddy:
#662 is the same thing as this PR but without all the other changes.

This PR (#626) also removes the "A Word About Compiler Quirks" section. Ethan, the only question I have for you is if you're okay with that section being completely removed. Me personally, I'm in favor of it, or at least it should be done when 2.3 is tagged; I don't think anyone's going to fiddle with 2.2 code anytime soon. However at the very least it ought to be done in a separate merge and not lumped in together with any other ones.

flibitijibibo:
I suppose for 2.3 we can cut that section, now that all the undefined behavior has been fixed at this point. When we fix the loop I'll do a VS2010 build and as long as it behaves even in release mode we can cut that text out.

PR #628, opened 2021-02-25
Explicitly declare void for all void parameter functions

InfoTeddy:
Apparently in C, if you have `void test();`, it's completely okay to do `test(2);`. The function will take in the argument, but just discard it and throw it away. It's like a trash can, and a rude one at that. If you declare it like `void test(void);`, this is prevented.

This is not a problem in C++ - doing `void test();` and `test(2);` is guaranteed to result in a compile error (this also means that right now, at least in all `.cpp` files, nobody is ever calling a void parameter function with arguments and having their arguments be thrown away). However, we may not be using C++ in the future, so I just want to lay down the precedent that if a function takes in no arguments, you must explicitly declare it as such.

And in fact, actually, doing this revealed that the forward-declares for `unlockAchievement()` in `Network.c` were mistakenly being declared with no arguments, even though that function takes in a `const char*` as an argument. I've fixed that in this PR, too.

I would've added `-Wstrict-prototypes`, but it produces an annoying warning message saying it doesn't work in C++ mode if you're compiling in C++ mode. So it can be added later.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BinaryBlob.cpp
+++ b/desktop_version/src/BinaryBlob.cpp
@@ -7,7 +7,7 @@
 #include "Exit.h"
 #include "UtilityClass.h"
 
-binaryBlob::binaryBlob()
+binaryBlob::binaryBlob(void)
 {
 	numberofHeaders = 0;
 	SDL_zeroa(m_headers);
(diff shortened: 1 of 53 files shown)


NogNoa:
yeah that's like totaly a backward compability feature for a 1978 version of the language where the arguments were declared between the () and the { for some reason (and they're probably still allowed to be).

PR #632, opened 2021-02-27
Refactor assets mounting to not use the STL, and clean up some code as well

InfoTeddy:
`FILESYSTEM_mountassets()`, `FILESYSTEM_assets()`, and `endsWith()` no longer use `std::string`s, since they don't need to. Along the way, I also cleaned up some other code that interacted with them, because they needed it too.

As always, look at the commits for more info.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -208,6 +208,14 @@ void FILESYSTEM_mountassets(const char* path)
 	const char* zip_normal;
 	char dir[MAX_PATH];
 
+	/* path is going to look like "levels/LEVELNAME.vvvvvv".
+	 * We want LEVELNAME, which entails starting from index 7
+	 * (which is how long "levels/" is)
+	 * and then grabbing path_size-14 characters
+	 * (14 chars because "levels/" and ".vvvvvv" are both 7 chars).
+	 * We also add 1 when calculating the amount of bytes to grab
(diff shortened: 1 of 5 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -164,66 +164,132 @@ char *FILESYSTEM_getUserLevelDirectory(void)
...
+	const size_t path_size = SDL_strlen(path);
+	char filename[MAX_PATH];
+	char zip_data[MAX_PATH];
+	const char* zip_normal;
+	char dir[MAX_PATH];
+
+	SDL_strlcpy(
+		filename,
+		&path[7],
+		VVV_min((path_size - 14) + 1, sizeof(filename))
```
flibitijibibo, line 214: A couple magic numbers that need comments (mostly stemming from me not commenting them originally)...

flibitijibibo (changes requested):
Just needs a couple comments, can be done with a new commit on top of the existing patchset.

desktop_version/src/FileSystemUtils.cpp
InfoTeddy: You weren't the person that added this code (unless you're talking about being the person who merged it in without asking for it to be commented). But yeah, I'll comment it.

InfoTeddy: Done.

PR #635, opened 2021-03-03
Add bounds checks to fix undefined behavior with levels bigger than the size limit

InfoTeddy:
This adds bounds checks so levels bigger than 20x20 don't run into undefined behavior.

Most of the work done is creating setter and getter funcs and replacing all existing raw accesses with those functions. Then the rest is just adding bounds checks to those functions.

Anyways, with this PR, it should no longer be possible to cause undefined behavior by going past the size limit. I tested it with Valgrind, to be sure.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1250,9 +1250,8 @@ void entityclass::createentity( float xp, float yp, int t, float vx /*= 0*/, flo
 
 #if !defined(NO_CUSTOM_LEVELS)
     // Special case for gray Warp Zone tileset!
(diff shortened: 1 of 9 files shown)


flibitijibibo:
This all makes sense to me. Do we want to log any cases where `INBOUNDS` fails?

InfoTeddy:
Unless you want me to, I don't think we need to log any of those cases.

flibitijibibo:
In the cases where they'd be extra spammy (once-per-frame type stuff), probably not - if there are any one-offs that would point to lurking danger later on, those would be good to print.

InfoTeddy:
There's not really any good one-offs I can think of - you are either in-bounds, or you are not.

PR #636, opened 2021-03-03
Fix regression with background of level minimaps not being black

InfoTeddy:
I had misread this line in #629 and thought that it was just clearing the entire surface, when really it was filling the surface with opaque black. `ClearSurface()` would instead make it transparent, which would mean when it got drawn, it would get drawn against blue, and not black. Whoops.

## Legal Stuff:

  for these changes


flibitijibibo:
Made the same mistake myself during the review...

PR #642, opened 2021-03-05
Untabify every single file

InfoTeddy:
Yeah, so, all files are indented with 4-column spaces now.

I did a simple `rg -l '\t' | xargs -n 1 sed -i -e 's/\t/    /g'` in `desktop_version/` to produce this commit. So even though it is a 58592-line diff, you can check for yourself that I didn't sneak in anything bad or anything.

## Legal Stuff:

  for these changes


flibitijibibo:
Ehhh this goes too far. I can believe that some code will just never be touched and therefore always be ugly, but a lot of it probably _will_ get fixed eventually.

InfoTeddy:
Eventually? Not if you adhere to the rule (as I have) that you indent with the same indentation as the surrounding indentation. That rule doesn't make the mixed indentation any *worse*, but it still leaves an inconsistency throughout the codebase. I don't see how it'll be fixed eventually without just saying "screw it" and reformatting everything in one go.

flibitijibibo:
If you want to maintain another branch that's fine, but I really _really_ don't want to merge this until we know for sure it's required to ship. We have blockers to worry about right now.

InfoTeddy:
Alright, fair enough. Closing because it's annoying to rebase and to not be distracting.

PR #647, opened 2021-03-06
Copy blend mode to recreated surfaces, fixing room name background during menu animation in Flip Mode

InfoTeddy:
This fixes a "root cause" bug (that's existed since 2.2 and below) where recreated surfaces wouldn't preserve the blend mode of their original surface.

The surface-level (pun genuinely unintended) bug that this root bug fixes is the one where there's no background to the room name during the map menu animation in Flip Mode.

This is because the room name background relies on `graphics.backBuffer` being filled with complete black. This is achieved by a call to `ClearSurface()` - however, `ClearSurface()` actually fills it with transparent black (this is not a regression; in 2.2 and previous, this was an "inlined" `FillRect(backBuffer, 0x00000000)`). This would be okay, and indeed the room name background renders fine in unflipped mode - but it suddenly breaks in Flip Mode.

Why? Because `backBuffer` gets fed through `FlipSurfaceVerticle()`, and `FlipSurfaceVerticle()` creates a temporary surface with the same dimensions and color masks as `backBuffer` - it, however, does *not* create it with the same blend mode, and kind of sort of just forgets that the original was `SDL_BLENDMODE_NONE`; the new surface is `SDL_BLENDMODE_BLEND`. Thus, transparency applies on the new surface, and instead of the room name being drawn against black, it gets drawn against transparency.

Before:
Flip Mode menu animation room name before

After:
Flip Mode menu animation room name after

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/GraphicsUtil.cpp
+++ b/desktop_version/src/GraphicsUtil.cpp
@@ -17,6 +17,7 @@ static SDL_Surface* RecreateSurfaceWithDimensions(
     const int height
 ) {
     SDL_Surface* retval;
+    SDL_BlendMode blend_mode;
 
     if (surface == NULL)
     {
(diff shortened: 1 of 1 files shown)


TerryCavanagh:
Looks good. Thanks!

PR #654, opened 2021-03-06
De-duplicate m_renderer and m_screenTexture creation in Screen.cpp

InfoTeddy:
The code to create `m_renderer` and `m_screenTexture` is now no longer copy-pasted for the VSync reset workaround.

## Legal Stuff:

  for these changes


flibitijibibo:
Technically this is correct, but the FIXMEs are more a reference to an SDL bug, which we should fix instead:

https://github.com/libsdl-org/SDL/issues/3673

Let's try to fix ^, then we can just delete the duplicated portion.

PR #660, opened 2021-03-17
Directly toggle fullscreen if keybind pressed in key.Poll()

InfoTeddy:
This moves the responsibility of toggling fullscreen when any of the three toggle fullscreen keybinds are pressed (F11, Alt+Enter, Alt+F) directly into `key.Poll()` itself, and not its caller (which is `main()` - more specifically, `fixedloop()`). Furthermore, the fullscreen toggle itself has been moved to a separate function that `key.Poll()` just calls, to prevent cluttering `key.Poll()` with more business logic (the function is already quite big enough as it is).

As part of my work in re-removing the 1-frame input delay in #535, I'm moving the callsite of `key.Poll()` around, and I don't want to have to lug this block of code around with it. I'd rather refactor it upfront than touch any more lines than necessary in that PR.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/KeyPoll.cpp
+++ b/desktop_version/src/KeyPoll.cpp
@@ -41,7 +41,6 @@ KeyPoll::KeyPoll(void)
 	leftbutton=0; rightbutton=0; middlebutton=0;
 	mx=0; my=0;
 	resetWindow = 0;
-	toggleFullscreen = false;
 	pressedbackspace=false;
 
 	useFullscreenSpaces = false;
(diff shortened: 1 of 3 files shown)


flibitijibibo:
I remember the timing of this being a little bit sensitive, particularly regarding keymap - what we can probably do to isolate this while preserving timing is have bool toggleFullscreen as a local variable, then do the toggle call at the bottom of KeyPoll. Roughly the same idea, just looks slightly different in Poll() itself.

PR #664, opened 2021-03-18
Add being able to press Esc to return to previous menu

InfoTeddy:
This is a small quality-of-life thing that makes it so you don't have to move your menu selection all the way over to the "return" button in order to return to the previous menu. You can just press Escape instead to return to the previous menu. The previous behavior of pressing Escape was to bring up the 'confirm quit' menu, or if you were in an options menu in-game, return to the pause menu.

If you're on the main menu (and thus don't have any previous menu) and press Escape, the game will instead bring up the 'confirm quit' menu. For consistency, the "quit game" option on the main menu will also bring up the 'confirm quit' menu as well, instead of immediately closing the game.

Pressing the controller button mapped to Escape will also work as well.

The only menus that don't have return buttons are the 'countdown' menus - so the game will not let you press Escape if there's a menu countdown happening.

Now that pressing Escape in the 'continue' menu will just bring you back to the 'play' menu, there's no need to specifically put `map.nexttowercolour()` first when canceling the 'confirm quit' menu.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -261,10 +261,9 @@ static void menuactionpress(void)
  #define MPOFFSET -2
 #endif
         case OFFSET+5:
-            //bye!
-            music.playef(2);
-            game.mainmenu = 100;
-            graphics.fademode = 2;
+            music.playef(11);
+            game.createmenu(Menu::youwannaquit);
+            map.nexttowercolour();
             break;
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Yeah, this feature is _really_ overdue... pretty sure I hit escape a double-digit number of times during 2.2.

PR #667, opened 2021-03-19
Add text outlines to textboxless textboxes and gravitron text

InfoTeddy:
This PR adds text outlines (if enabled) to textboxless textboxes - these are text boxes with RGB values of 0,0,0, which just get rendered as text without any text box behind it - and all gravitron text. I've refactored some of the text printing a bit, but also copy-pasted some of the text printing calculations, because of how the text printing functions work - and I'm beginning to think these text printing functions are a bit awful to work with... but that's a refactor for later.

These text outlines should be added in case they get placed against hard-to-contrast backgrounds, which could happen in custom levels (since you can start the Gravitron or Super Gravitron in custom levels), or if someone manages to glitch themselves out in the main game.

(Depends on #670).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -860,7 +860,7 @@ void Graphics::drawgui(void)
         {
             for (size_t j = 0; j < textbox[i].line.size(); j++)
             {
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Makes sense to me, but since this affects visuals I'll defer this one...

TerryCavanagh:
Yep, this seems sensible - I vaguely remember doing this on purpose, but it's a tiny detail ultimately and it's probably better this way.

PR #668, opened 2021-03-19
Fix oversight with unfocus pause screen in Flip Mode

InfoTeddy:
Flip Mode flips all the unfocus pause screen text upside-down, to make it read in reverse order. This looks kind of strange to me, and I don't think it was intended. So I'm flipping the text again so it's the right way up in Flip Mode.

## Legal Stuff:

  for these changes


flibitijibibo:
Will leave this one to Terry...

TerryCavanagh:
Yep, this was an oversight! Thanks for the fix.

PR #688, opened 2021-03-31
Fix memory leak when pasting text

InfoTeddy:
According to [SDL documentation](https://wiki.libsdl.org/SDL_GetClipboardText), the returned pointer needs to be freed. A glance at the source code confirms that the function allocates, and also Valgrind complains about it.

Also if it couldn't allocate, the game no longer segfaults (`std::string`s do not check if the pointer is non-`NULL` for `operator+=`).

## Legal Stuff:

  for these changes


flibitijibibo:
Right, forgot about that... I need to do this for GetBasePath too.

PR #690, opened 2021-03-31
Add WebAssembly support

leo60228:
## Changes:

Emscripten has a lot of weird quirks, which made porting 2.2 *very* annoying. Luckily, refactors in 2.3 lead to most of them not being hit anymore. This PR contains the boilerplate build system support, as well as workarounds for two Emscripten quirks (see comments and commit messages).

Note that due to an Emscripten bug you'll likely run into a few linker errors involving SDL. I have a [workaround](https://github.com/leo60228/SDL/commit/579384cbf236d610f2fa736d4746fcffb1c1931c) in my SDL fork, and I'm planning on PRing a proper fix to Emscripten (likely along with updating their SDL2 fork to 2.0.14).

## Legal Stuff:

  for these changes


desktop_version/src/main.cpp
```diff
@@ -356,6 +373,15 @@ static void inline deltaloop(void);
 
 static void cleanup(void);
 
+#ifdef __EMSCRIPTEN__
+void emscriptenloop()
```
InfoTeddy, line 377:
```suggestion
void emscriptenloop(void)
```

desktop_version/src/main.cpp
```diff
@@ -316,7 +324,16 @@ static enum LoopCode loop_run_active_funcs(void)
...
+            // Loop_continue if a Func_delta exists, otherwise Loop_stop
+            for (int i = 0; i < *num_active_funcs; i++)
+            {
+                if ((*active_funcs)[i].type == Func_delta)
+                {
+                    return Loop_continue;
+                }
+            }
+
+            break;
```
InfoTeddy, line 336: This is okay for now, but I have a better fix for the signal interrupt being ignored during unfocus pause.

InfoTeddy (changes requested):
Looks good to me, although I don't know much about Emscripten and haven't tested this myself.

flibitijibibo:
This makes sense to me - the filesystem changes might get simplified after we axe the old save migration. Let's get this in once 2.3 has calmed down and we've started ripping out the dead wood for 2.4.

desktop_version/src/main.cpp
InfoTeddy: Pretty sure that fix has made it in, so we don't need this anymore.

InfoTeddy (commented):
Looks good to me, just needs a rebase (and maybe remove that one thing).

leo60228:
> Pretty sure that fix has made it in, so we don't need this anymore.

The function doesn't seem to have been changed on master. Was it fixed elsewhere?

leo60228:
Rebased. Still need to test.

desktop_version/src/main.cpp
```diff
@@ -632,6 +652,8 @@ int main(int argc, char *argv[])
     }
 
     cleanup();
```
InfoTeddy, line 654: Any way to call `cleanup()` when the loop ends?

InfoTeddy (commented):
Got it - do you want me to merge this in now or wait until it's tested?

desktop_version/src/main.cpp
leo60228: I don't think so, but I could be wrong.

InfoTeddy: No big deal to me either way. It's only needed to minimize Valgrind memory leak messages.

leo60228:
For reference on building...
To generate `VVVVVV.js` and `VVVVVV.wasm`:
```
mkdir build
cd build
emcmake cmake .. -DCMAKE_BUILD_TYPE=Release # unoptimized wasm binaries are *huge*
export EMCC_LOCAL_PORTS=sdl2=/path/to/SDL # upstream is broken due to an emscripten bug that I should submit a patch for, workaround at https://github.com/leo60228/SDL/commit/579384cbf236d610f2fa736d4746fcffb1c1931c
emmake make # first build shouldn't be parallel (multiple instances try to build SDL2 at the same time, which doesn't work)
```
To convert `data.zip` to the correct Emscripten-specific format:
```
/path/to/emscripten/tools/file_packager.py VVVVVV.data --preload data.zip --js-output=VVVVVV.data.js --use-preload-cache
```
The entry point for the output is `VVVVVV.js`, which can't be executed directly. A minimal example webpage is at https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html, and will work for VVVVVV if you replace `{{{ SCRIPT }}}` with this:
```html
    <script src="VVVVVV.data.js"></script>
    <script src="VVVVVV.js"></script>
```
A more complete webpage would probably also configure somewhere to persist save data, and be less ugly than the default styling. I'm not sure if this is in scope for this PR, though.

leo60228:
I think this is ready to merge, unless we want to include a webpage to use with the Emscripten output.

PR #693, opened 2021-04-02
Add option to re-enable 1-frame input delay

InfoTeddy:
This is an option for speedrunners whose muscle memory is precisely trained and used to the 1-frame input delay that existed in 2.2 and below. It is located in Game Options -> Advanced Options, and is off by default.

To re-add the 1-frame input delay, we simply move the `key.Poll()` to the start of the frame, instead of before an input function gets ran - undoing what #535 did.

There is a frame ordering-sensitive issue here, where toggling `game.inputdelay` at the wrong time could cause double-polling. However, we only toggle it in an input function, which regardless is always guaranteed to be ran after `key.Poll()` (it either happened at the start of the frame or just before the input function got ran), so this is not an issue. But, in case we ever need to toggle this variable in the future, we can just use the defer callbacks system to defer the toggle to the end of the frame - also added by #535.

Added at the request of Habeechee on the VVVVVV speedrunning Discord server.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -385,6 +385,7 @@ void Game::init(void)
     kludge_ingametemp = Menu::mainmenu;
 
     disablepause = false;
+    inputdelay = false;
 }
 
 void Game::lifesequence(void)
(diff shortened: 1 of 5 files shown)


flibitijibibo:
I wonder if we need a speedrunner menu as part of #696... for now, we can get the functionality in.

PR #697, opened 2021-04-02
 Re-fix resumemusic/musicfadein once again

InfoTeddy:
So it looks like facb079b3597b380f876537523cd351a0e637b62 (PR #316) had a few issues.

The SDL performance counter doesn't really work that well. Testing reveals that unfocusing and focusing the game again results in the `resumemusic()` script command resuming the track at the wrong time. Even when not unfocusing the game at all, stopping a track and resuming it resumes it at the wrong time. (Only disabling the unfocus pause fixes this.)

Furthermore, there's also the fact that the SDL performance counter keeps incrementing when the game is paused under GDB. So... yeah.

Instead of dealing with the SDL performance counter, I'm just going to pause and resume the music directly (so the `stopmusic()` script command just pauses the music instead). As a result, we no longer can keep constantly calling `Mix_PauseMusic()` or `Mix_ResumeMusic()` when focused or unfocused, so I've moved those calls to happen directly when the relevant SDL events are received (the constant calls were originally in VCE, and whoever added them (I'm pretty sure it was Leo) was not the sharpest tool in the shed...).

And we are going to switch over to using our own fade system instead of the SDL mixer fade system. In fact, we were already using our own fade system for fadeins after collecting a trinket or a custom level crewmate, but we were still using the mixer system for the rest. This is an inconsistency that I am glad to correct, so we're also doing our own fadeouts now.

There is, however, an issue with the fade system where the length it goes for is inaccurate, because it's based on a volume-per-frame second calculation that gets truncated. But that's an issue to fix later - at least what I'm doing right now makes `resumemusic()` and `musicfadein()` work better than before.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -6742,3 +6742,33 @@ void Game::copyndmresults(void)
     ndmresulthardestroom = hardestroom;
     SDL_memcpy(ndmresultcrewstats, crewstats, sizeof(ndmresultcrewstats));
 }
+
+static inline int get_framerate(const int slowdown)
+{
+    switch (slowdown)
+    {
+    case 30:
+        return 34;
+    case 24:
+        return 41;
+    case 18:
+        return 55;
+    case 12:
+        return 83;
+    }
+
+    return 34;
+}
+
+int Game::get_timestep(void)
+{
+    switch (gamestate)
+    {
+    case EDITORMODE:
+        return 24;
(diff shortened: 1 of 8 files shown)


flibitijibibo:
GetPerformanceCounter is really a system timer, it's good for global time but not great if you need to pause or resume a lot, where the monotonic clock will work better.

PR #703, opened 2021-04-09
Clean up some things from eee98b0e071507a80e9ceec55eb29ca87a102319

InfoTeddy:
This cleans up a few code style things that eee98b0e071507a80e9ceec55eb29ca87a102319 introduced.

## Legal Stuff:

  for these changes


TerryCavanagh:
Hey, so, educate me on this one! (I think we all know at this point that I'm less a programmer and more three script kiddies in a trenchcoat)

In this commit: https://github.com/TerryCavanagh/VVVVVV/pull/703/commits/5a08af5615844db3dbf1b08e25076f0b6d0a26f6

...why is a switch statement better here?

flibitijibibo:
I dunno if it's _better_ as much as it allows for blocks without braces - the tradeoff is you _have_ to insert a `break;` or else it falls into itself, but I think most of the people who were contributing just slowly moved to switch because their tools supported it well (i.e. modern compilers will scream at you if you don't have `break;` at the end). Functionally, it's identical.

InfoTeddy:
So you don't have to repeat the left operand `game.currentmenuoption ==` multiple times. Furthermore, switch statements in C/C++ can be and most of the time are optimized to jump tables by the compiler. Normal if-trees can be optimized to jump tables too, but it's best to not leave it up to chance. Also, if you have a really giant if-tree you might run into issues with compilers (apparently MSVC has a limit of 128 branches in an if-tree, which resulted in the giant if-tree in `Scripts.cpp` having to be broken, which is one of the things that introduced internal scripting).

It's a small thing, but it's good to be consistent.

flibitijibibo:
>apparently MSVC has a limit of 128 branches in an if-tree, which resulted in the giant if-tree in Scripts.cpp having to be broken, which is one of the things that introduced internal scripting

Oh, didn't know that about MSVC... huh.

>Furthermore, switch statements in C/C++ can be and most of the time are optimized to jump tables by the compiler

Teeechnically this can happen for if/else as well (this was the case for our hardcoded number strings way back in 2.2), but yeah, switch is more explicit about this than if/else.

TerryCavanagh:
Ok, makes sense! Good to be consistent :)

PR #718, opened 2021-04-12
Re-fix handling filepaths with non-ASCII characters on Windows

Dav999-v:
In #52 I fixed VVVVVV not being able to handle filepaths with non-ASCII characters on Windows. 2f0a0bce4c663b7df2ed229ef897dcf3fd878ec6 and aa5c2d9dc260386c1b99fb9867b2a8ef7c82252f reintroduce this problem, however, by reverting the definition of mkdir to how it was before the fix and using the non-Unicode version of CreateDirectory. And I can confirm that VVVVVV indeed doesn't make its folder anymore with a Windows username of "тест". This PR fixes that issue.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -16,7 +16,12 @@
 #if defined(_WIN32)
 #include <windows.h>
 #include <shlobj.h>
-#define mkdir(a, b) CreateDirectory(a, NULL)
+int mkdir(char* path, int mode)
+{
+	WCHAR utf16_path[MAX_PATH];
+	MultiByteToWideChar(CP_UTF8, 0, path, -1, utf16_path, MAX_PATH);
+	return CreateDirectoryW(utf16_path, NULL);
+}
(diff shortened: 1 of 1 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -585,7 +590,7 @@ static void PLATFORM_getOSDirectory(char* output)
 	SHGetFolderPathW(NULL, CSIDL_PERSONAL, NULL, SHGFP_TYPE_CURRENT, utf16_path);
 	WideCharToMultiByte(CP_UTF8, 0, utf16_path, -1, output, MAX_PATH, NULL, NULL);
 	SDL_strlcat(output, "\\VVVVVV\\", MAX_PATH);
-	CreateDirectory(output, NULL);
+	mkdir(output, 0);
```
flibitijibibo, line 593: Nit:
```suggestion
	mkdir(output, 0777);
```

Otherwise lgtm.

PR #727, opened 2021-04-14
Fix music stopping when restarting a time trial

trelbutate:
## Changes:

This fixes the current bug where quick restarting a time trial will completely stop the music until you go back to the main menu.
I also cleaned up the code around it a bit by merging several nearly identical switch cases into one.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CONTRIBUTORS.txt
+++ b/desktop_version/CONTRIBUTORS.txt
@@ -11,6 +11,7 @@ Contributors
 * Brian Callahan (@ibara)
 * Dav999 (Dav999-v)
 * Allison Fleischer (AllisonFleischer)
+* Malte Grimm (@trelbutate)
 * Daniel Lee (@ddm999)
 * Fredrik Ljungdahl (@FredrIQ)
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Looks good! If you can add yourself here and here...

https://github.com/TerryCavanagh/VVVVVV/blob/master/desktop_version/CONTRIBUTORS.txt

https://github.com/TerryCavanagh/VVVVVV/blob/master/desktop_version/src/Credits.h#L86

... this is good to go.

trelbutate:
Right, done!

InfoTeddy:
I wouldn't have merged this.

This results in the music after restarting the time trial deviating from 2.2 behavior. In 2.2, the music would fade out, and then start back up again after the countdown. But this PR makes it so the music continues throughout the restart and never stops at all.

Also, there are plenty of other cases caused by the same root bug (they could happen in custom levels), so this only fixes one single case, and not every case.

I have addressed all of these issues in #728.

PR #730, opened 2021-04-17
Axe mouse cursor config option in favor of automatically toggling mouse

InfoTeddy:
Instead of having a one-size-fits-all config option, just make it so we hide the mouse when the user uses their keyboard or controller, and show it again when they move their mouse.

## Legal Stuff:

  for these changes


TerryCavanagh:
Makes sense to me!

PR #740, opened 2021-04-18
Fix undefined behavior if assets are missing and asserts if some of them are missing

InfoTeddy:
It's simple: The game committed undefined behavior if it had no assets. But now it doesn't.

To test this, I created a zip file with nothing inside (`touch nothing`, then `zip nothing.zip nothing`), then did `./VVVVVV -assets nothing.zip` and fixed every segfault until it worked properly.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -451,6 +451,11 @@ void FILESYSTEM_loadFileToMemory(
 	PHYSFS_sint64 length;
 	PHYSFS_sint64 success;
 
+	if (name == NULL || mem == NULL)
+	{
+		goto fail;
+	}
+
 	if (SDL_strcmp(name, "levels/special/stdin.vvvvvv") == 0)
 	{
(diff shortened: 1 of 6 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -447,11 +447,16 @@ void FILESYSTEM_loadFileToMemory(
...
+	PHYSFS_File *handle;
+	PHYSFS_sint64 length;
+	PHYSFS_sint64 success;
+
 	if (SDL_strcmp(name, "levels/special/stdin.vvvvvv") == 0)
 	{
 		// this isn't *technically* necessary when piping directly from a file, but checking for that is annoying
 		static std::vector<char> STDIN_BUFFER;
 		static bool STDIN_LOADED = false;
+		size_t length;
```
flibitijibibo, line 464: This now shadows the previous length variable, so it should be renamed.

InfoTeddy: Good catch! Fixed, rebased, force-pushed.

PR #741, opened 2021-04-18
Remove unnecessary Sint16 casts

InfoTeddy:
These casts are sprinkled all throughout the graphics code when creating and initializing an `SDL_Rect` on the same line. Unfortunately, most of these are unnecessary, and at worst are wasteful because they result in narrowing a 4-byte integer into a 2-byte one when they don't need to (`SDL_Rect`s are made up of 4-byte integers).

Now, removing them reveals why they were placed there in the first place - a warning is raised (`-Wnarrowing`) that implicit narrowing conversions are prohibited in initializer lists in C++11. (Notably, if the conversion wasn't narrowing, or implicit, or done in an initializer list, it would be fine. This is a really specific prohibition that doesn't apply if any of its sub-cases are true.)

We don't use C++11, but this warning can be easily vanquished by a simple explicit cast to int (similar to the error of implicitly converting `void*` to any other pointer in C++, which works just fine in C), and we only need to do it when the warning is raised (not every single time we make an `SDL_Rect`), so there we go.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -45,7 +45,7 @@ void Graphics::init(void)
     //Background inits
     for (int i = 0; i < numstars; i++)
     {
-        SDL_Rect s = {Sint16(fRandom() * 320), Sint16(fRandom() * 240), 2, 2};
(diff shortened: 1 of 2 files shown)


flibitijibibo:
This is a carryover from SDL 1.2, which used to use Sint16.

PR #742, opened 2021-04-18
Add error checking to base path setup 

InfoTeddy:
Previously, if the game couldn't set the write dir to the base directory, or couldn't make the base directory, or couldn't calculate the base directory, it would probably dereference `NULL` or read from uninitialized memory or murder your family or something. But now, I've eliminated the potential Undefined Behavior from the code dealing with the base path.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -39,7 +39,7 @@ static char levelDir[MAX_PATH] = {'\0'};
 static char assetDir[MAX_PATH] = {'\0'};
 static char virtualMountPath[MAX_PATH] = {'\0'};
 
-static void PLATFORM_getOSDirectory(char* output, const size_t output_size);
+static int PLATFORM_getOSDirectory(char* output, const size_t output_size);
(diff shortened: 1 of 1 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -707,17 +732,63 @@ void FILESYSTEM_enumerateLevelDirFileNames(
...
+	HRESULT retcode = SHGetFolderPathW(
+		NULL,
+		CSIDL_PERSONAL,
+		NULL,
+		SHGFP_TYPE_CURRENT,
+		utf16_path
+	);
+	int num_bytes;
+
+	if (retcode != S_OK)
```
flibitijibibo, line 749: This should use `SUCCEEDED()` or `FAILED()`, HRESULT is really weird and needs these macros to be reliable

InfoTeddy: Weird - I didn't find any mention of this totally very excellent API design when I looked up `HRESULT` for the first time. Regardless: Fixed, rebased, force-pushed.

PR #743, opened 2021-04-19
Separate pressing Enter to open map from pressing Enter to interact

### LINKED ISSUE #736: It's possible to open the map during a teleport
On 2.3-testing it seems to be possible to open the map while the teleport animation is playing. This threw me off the first time I noticed this, as I tend to spam Y on the controller to activate teleporters as fast as possible.

InfoTeddy:
This is a lot of copy-pasted code, but a little bit of copy-pasting never hurt anyone...

The keybind to interact with activity zones and teleporters is now separate from the keybind to open the map, or return to the editor from in-editor playtesting, or restart a time trial. The keybind is now E, and the default controller bind is X. No controller button prompts, but the game didn't have controller button prompts anyways, so whatever.

Doing this now because if people's muscle memory are going to be broken by not being able to spam the map keybind anymore, at least we can help a bit by changing the keybind so they can keep spamming it - their muscle memory is going to be broken anyways.

This option has to be enabled by going to the speedrunner menu options and selecting "interact button". It is disabled by default.

All prompt text needs to be string-interpolated every time they are drawn, because it is possible for people to change which interact button they use in the middle of gameplay, via the in-game options.

Closes #736.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -838,151 +838,151 @@ void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*=
         switch(trig)
         {
         case 0: //testing zone
-            block.prompt = "Press ENTER to explode";
+            block.prompt = "Press %s to explode";
             block.script = "intro";
             block.setblockcolour("orange");
             trig=1;
             break;
         case 1:
-            block.prompt = "Press ENTER to talk to Violet";
+            block.prompt = "Press %s to talk to Violet";
             block.script = "talkpurple";
             block.setblockcolour("purple");
             trig=0;
             break;
         case 2:
-            block.prompt = "Press ENTER to talk to Vitellary";
+            block.prompt = "Press %s to talk to Vitellary";
             block.script = "talkyellow";
             block.setblockcolour("yellow");
             trig=0;
             break;
         case 3:
-            block.prompt = "Press ENTER to talk to Vermilion";
+            block.prompt = "Press %s to talk to Vermilion";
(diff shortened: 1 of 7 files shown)


flibitijibibo:
This gets into 2.4 territory, but I can be overruled here if Terry thinks this is worth breaking the freeze.

TerryCavanagh:
Oh, uff, nope, sorry, I don't like this. I very deliberately kept the keybindings simple in VVVVVV and adding E (I guess because it's close to WASD?) as a necessary key really confuses things. Sorry.

A pull request I'd be happy to support would be for custom rebinding for each control in the game, where the default behaviour would be the classic "Enter" as interact key, and people could specify if they wanted to use another key for that instead. Or, less dramatically, I'd also be ok with a speedrunner's option to use E instead of Enter in the speedrunner menu. But I don't like the idea of changing the default here.

flibitijibibo:
Yeah, let's try and make the action set rewrite the very first thing we do for 2.4, so this can be made available ASAP.

InfoTeddy:
Interesting that you want to keep the keybindings simple; I recall one of the console versions separating Enter to open map from Enter to interact. Although, I don't play the console versions, so correct me if I'm wrong about that.

> I'd also be ok with a speedrunner's option to use E instead of Enter in the speedrunner menu.

Alright; I've added the toggle switch, rebased, and force-pushed. @TerryCavanagh can you take a look at this again?

TerryCavanagh:
Ok! Had a read through and that looks good to me, thanks!

PR #746, opened 2021-04-20
Fix only removing duplicate script names from script name list

InfoTeddy:
If there were two scripts with the same name, removing one of them would only remove the other script from the script name list, and not also remove the contents of said script - leading to a desync in state, which is probably bad.

Fixing this isn't as simple as removing the `break` statement - I either also have to decrement the loop variable when removing the script, or iterate backwards. I chose to iterate backwards here because it relocates less memory than iterating forwards.

Also I removed the `<algorithm>` include from `editor.cpp`.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -3,7 +3,6 @@
 #define ED_DEFINITION
 #include "editor.h"
 
-#include <algorithm>
 #include <stdio.h>
 #include <string>
 #include <tinyxml2.h>
@@ -469,8 +468,16 @@ void editorclass::removehookfromscript(std::string t)
 void editorclass::removehook(std::string t)
 {
     //Check the hooklist for the hook t. If it's there, remove it from here and the script
+    size_t i;
     removehookfromscript(t);
-    hooklist.erase(std::remove(hooklist.begin(), hooklist.end(), t), hooklist.end());
+    /* When this loop reaches the end, it wraps to SIZE_MAX; SIZE_MAX + 1 is 0 */
+    for (i = hooklist.size() - 1; i + 1 > 0; --i)
+    {
+        if (hooklist[i] == t)
+        {
+            hooklist.erase(hooklist.begin() + i);
+        }
+    }
 }
 
 void editorclass::addhook(std::string t)
@@ -4263,8 +4270,15 @@ void editorinput(void)
                 ed.keydelay=6;
             }
 
-            // Remove all pipes, they are the line separator in the XML
-            key.keybuffer.erase(std::remove(key.keybuffer.begin(), key.keybuffer.end(), '|'), key.keybuffer.end());
+            /* Remove all pipes, they are the line separator in the XML
+             * When this loop reaches the end, it wraps to SIZE_MAX; SIZE_MAX + 1 is 0 */
+            {size_t i; for (i = key.keybuffer.length() - 1; i + 1 > 0; --i)
+            {
+                if (key.keybuffer[i] == '|')
+                {
+                    key.keybuffer.erase(key.keybuffer.begin() + i);
+                }
+            }}
 
             ed.sb[ed.pagey+ed.sby]=key.keybuffer;
             ed.sbx = utf8::unchecked::distance(ed.sb[ed.pagey+ed.sby].begin(), ed.sb[ed.pagey+ed.sby].end());
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -452,15 +452,14 @@ void editorclass::addhooktoscript(std::string t)
 
 void editorclass::removehookfromscript(std::string t)
 {
-    //Find hook t in the scriptclass, then removes it (and any other code with it)
-    for (size_t i = 0; i < script.customscripts.size(); i++)
+    /* Find hook t in the scriptclass, then removes it (and any other code with it)
+     * When this loop reaches the end, it wraps to SIZE_MAX; SIZE_MAX + 1 is 0 */
+    size_t i;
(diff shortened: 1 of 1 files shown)


desktop_version/src/editor.cpp
```diff
@@ -4264,7 +4268,13 @@ void editorinput(void)
             }
 
             // Remove all pipes, they are the line separator in the XML
-            key.keybuffer.erase(std::remove(key.keybuffer.begin(), key.keybuffer.end(), '|'), key.keybuffer.end());
+            {size_t i; for (i = key.keybuffer.length(); i-- > 0; )
```
flibitijibibo, line 4271:
```suggestion
            {size_t i; for (i = key.keybuffer.length(); i > 0; i -= 1)
```

desktop_version/src/editor.cpp
```diff
@@ -454,23 +453,28 @@ void editorclass::addhooktoscript(std::string t)
...
     }
 }
 
 void editorclass::removehook(std::string t)
 {
     //Check the hooklist for the hook t. If it's there, remove it from here and the script
+    size_t i;
     removehookfromscript(t);
-    hooklist.erase(std::remove(hooklist.begin(), hooklist.end(), t), hooklist.end());
+    for (i = hooklist.size(); i-- > 0; )
```
flibitijibibo, line 471:
```suggestion
    for (i = hooklist.size(); i > 0; i -= 1)
```

desktop_version/src/editor.cpp
```diff
@@ -454,23 +453,28 @@ void editorclass::addhooktoscript(std::string t)
 void editorclass::removehookfromscript(std::string t)
 {
     //Find hook t in the scriptclass, then removes it (and any other code with it)
-    for (size_t i = 0; i < script.customscripts.size(); i++)
+    size_t i;
+    for (i = script.customscripts.size(); i-- > 0; )
```
flibitijibibo, line 457:
```suggestion
    for (i = script.customscripts.size(); i > 0; i -= 1)
```

flibitijibibo (commented):
Had a bit of difficulty reading these loops, otherwise seems fine.

desktop_version/src/editor.cpp
InfoTeddy: Your suggestion is wrong. This skips over `i` being 0.

flibitijibibo: This is why I _really_ don't like ++/--... let's find a way to make this loop work without these, I completely misread this loop because of them.

InfoTeddy: Took me a bit of thinking and using my search engine, but I found the solution that kept the exact semantics of the for-loop.

It's also worth noting your suggestion would result in the first array indexing being out of bounds, since my code relied on the post-decrement side-effect in the second statement.

For posterity, my reasoning went like this: Without changing the exact semantics, we can write out the implicit check-the-previous-variable explicitly. All other options (unless you want to start using C++20) change the exact semantics a bit, which I don't want.

Changing the `size_t` to an `int` (to avoid issues with wraparound) reduces the size of the variable. There's also `ssize_t`, but that's POSIX and non-portable, and it still reduces the size.

I thought at first that doing `for (i = length(); i != -1; --i)` or `for (i = length(); i + 1 > 0; --i)` means the loop never happens if we've reached the maximum possible size represented by the variable, but then I realized that actually, the loop variable can never be `length()` because all indices have to be between 0 and `length() - 1` inclusive. So if we reach the maximum possible size (i.e. `length()` is equal to `SIZE_MAX`), then `length()` can still be used as a 'sentinel value' of sorts to stop the loop.

I ended up going with `for (i = length(); i + 1 > 0; --i)`, because `for (i = length(); i != -1; --i)` relies on the reader not only knowing about implicit integer conversions, but also knowing the conversion rules (-1 here is going to be promoted to unsigned, which means it becomes `SIZE_MAX`), and knowing about unsigned wraparound, and knowing that unsigned wraparound is well-defined behavior.

flibitijibibo: Looking at it one more time, I wonder if it's just easier to do it like this...

```
size_t i;
for (i = 0; i < thing.length(); i += 1)
{
    if (whatever)
    {
        thing.erase(i);
        i -= i; /* Prevents skipping elements */
    }
}
```
It's not the prettiest thing but it's very clear and avoids having to fight type sizes and weird index conditions.

InfoTeddy: This works, but it ends up moving more memory than doing it backwards - specifically, it ends up having to move the rest of the duplicates plus the rest of the vector, whereas doing it backwards moves only the rest of the vector and not the duplicates.

flibitijibibo: The `i + 1 > 0` seems like it wouldn't be consistent if it decided to go to `~0` in the subtraction. I would even be okay with

```
size_t i;
for (i = thing.length() - 1 ; i > 0; i -= 1)
{
    if (whatever[i])
    {
        thing.erase(i);
    }
}
if (whatever[0]) thing.erase(0); /* Hope you like memmove! */
```

We can make this a static inline routine if needed.

InfoTeddy: > The `i + 1 > 0` seems like it wouldn't be consistent if it decided to go to `~0` in the subtraction.

So by `~0` you mean bits-all-1, and that `~0 + 1 > 0` wouldn't work? Well, `~0` would just be `SIZE_MAX`; `SIZE_MAX + 1` is 0, and 0 is not greater than 0.

flibitijibibo: If you want to document that above each loop, go for it...

InfoTeddy: Done, rebased, and force-pushed.

PR #749, opened 2021-04-23
Fix tower camera invincibility inconsistencies

InfoTeddy:
If you have invincibility enabled, the tower camera behavior is inconsistent.

In ascending towers, you can "push" the camera upwards; however you cannot push it downwards; at least it stays still when it comes up to you if you stay still. In descending towers, the camera moves quicker when you're at the bottom of the screen, but it's slower than your falling speed and quickly loses sight of you; the camera can be pushed upwards; unfortunately it also does a "bumping" motion if you're standing still when the camera reaches you, which gets real annoying and isn't particularly pleasant to look at.

There are two problems, so this PR does two fixes:

1. Pushing the camera now applies the appropriate counter-offset depending on the direction of the tower. You can now push the camera downwards in ascending towers.

2. To fix the "bumping" when the camera reaches you if you stand still, there are now a 8-pixel-high "gray areas" at the top and bottom of the screen where the camera simply won't move if you're in them.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Logic.cpp
+++ b/desktop_version/src/Logic.cpp
@@ -892,11 +892,53 @@ void gamelogic(void)
 
                     if (above_screen)
                     {
-                        map.ypos-=10;
+                        if (obj.entities[player].yp - map.ypos <= 0)
+                        {
+                            if (graphics.towerbg.scrolldir == 1)
+                            {
+                                /* Descending tower:
+                                 * Counteract 10 pixels of terminal velocity
+                                 * + 2 pixels of camera movement */
+                                map.ypos -= 12;
+                            }
+                            else
+                            {
+                                /* Ascending tower:
+                                 * Move 8 out of 10 pixels of terminal velocity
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Will let Terry check this out since it involves camera control and design... seems right to me as far as code is concerned.

InfoTeddy:
I've split off the background smear fix to #753.

TerryCavanagh:
Ah, this one I think I'm going to have to check out and see. I don't think it's worth holding up v2.3 over, though?

InfoTeddy:
@TerryCavanagh If it helps, I've [made a video](https://www.youtube.com/watch?v=2Xj4mRIi2Gs) showcasing the differences that this PR makes, so you don't have to build and run it yourself.

TerryCavanagh:
Ok, yep, thank you! That answers my concerns, happy to merge :)

PR #750, opened 2021-04-23
Custom activity zone text and colours

AllyTally:
This PR adds two new internal commands to the game, being `setactivitycolour` and `setactivitytext`.
`setactivitycolour(colour)` sets the colour of the next activity zone spawned to the colour you pass as an argument, and `setactivitytext` changes the text of the net activity zone spawned to the text on the next line.
These changes were made this way to keep forwards compatibility; levels that use these commands will not break on earlier versions, only have the default orange `Press ENTER to activate terminal` text which would be shown anyway.

Additionally, activity zones spawned with ID 35 (custom level terminal) which are spawned through `createactivityzone` have had their default text changed to `Press ENTER to interact` (from `Press ENTER to activate terminal`). This change was done because using `createactivityzone` with invalid arguments would spawn an activity zone with the value stored in the `i` variable, which we can manipulate through other commands to set to `35`, which is normally the activity zone that activates terminals, meaning it's the only custom activity zone we can actually control. Since we can use this without it being attached to terminals, I think it's better for the default to be `Press ENTER to interact` instead of `Press ENTER to activate terminal`.

Terry has said [that these changes are fine,](https://owo.whats-th.is/7EwJMoC.png) so I'm making the pull request now.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -759,7 +759,7 @@ void entityclass::generateswnwave( int t )
     }
 }
 
-void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*= 0*/, const std::string& script /*= ""*/ )
+void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*= 0*/, const std::string& script /*= ""*/, bool custom /*= false*/)
 {
     k = blocks.size();
 
@@ -1051,7 +1051,12 @@ void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*=
             trig=0;
             break;
         case 35:
-            block.prompt = "Press %s to activate terminal";
+            if (custom)
+            {
+                block.prompt = "Press %s to interact";
+            } else {
+                block.prompt = "Press %s to activate terminal";
+            }
             block.script = "custom_"+customscript;
             block.setblockcolour("orange");
             trig=0;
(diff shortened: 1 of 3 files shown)


TerryCavanagh:
Yep, this looks good to me! If anyone else has any concerns about version compatibility here, feel free to weigh in, but otherwise, happy to merge this for 2.4! Thanks @AllyTally.

desktop_version/src/Entity.cpp
```diff
@@ -1048,7 +1050,14 @@ void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*=
...
         case 35:
-            block.prompt = "Press ENTER to activate terminal";
+            if (custom)
+            {
+                block.prompt = "Press ENTER to interact";
+            }
+            else
+            {
+                block.prompt = "Press ENTER to activate terminal";
+            }
```
InfoTeddy, line 1060: These need to be changed to be `"Press %s to interact"` and `"Press %s to activate terminal"`, due to the new interact button option.

InfoTeddy (approved):
I will merge this after a rebase (and after you fix the default text).

desktop_version/src/Entity.h
```diff
@@ -193,6 +193,8 @@ class entityclass
     bool customwarpmode, customwarpmodevon, customwarpmodehon;
     std::string customscript;
     bool customcrewmoods[Game::numcrew];
+    std::string customactivitycolour;
+    std::string customactivitytext;
```
InfoTeddy, line 197: These should also be reset in `scriptclass::hardreset()`.

PR #755, opened 2021-05-04
Enable returning to Secret Lab immediately from Super Gravitron

InfoTeddy:
When you enter the Super Gravitron, you have to wait until the Super Gravitron actually starts before being able to press Enter to return to the Secret Lab. This is annoying if you just want to get back to the Secret Lab. So, I've made it so the press-Enter-to-return functionality is enabled from the moment that the Super Gravitron starts.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2064,7 +2064,8 @@ void gameinput(void)
     {
         // Do nothing
     }
-    else if (game.swnmode == 1 && game.swngame == 1)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This seems like something that would have a weird side effect... any idea why it may have been this way originally?

InfoTeddy:
It was probably to prevent being able to freeze the game by starting a music track while one was already fading out. But that bug is fixed in 2.3.

PR #758, opened 2021-05-06
Allow using Esc in Super Gravitron quit menu

InfoTeddy:
This is a small quality-of-life fix in the same vein as allowing the player to press Esc in the teleporter menu (which they weren't able to do in 2.2, either).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2215,13 +2215,16 @@ void mapinput(void)
         {
             game.press_action = true;
         }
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Assigning to Terry... this makes sense to me but it's one of those "wait, this is _too_ easy" sort of things where I expect all hell to break loose if some weird condition is met when this key is pressed.

TerryCavanagh:
This looks ok to me. If on the offchance it does introduce a new bug, it won't be a very serious one I think. Merging!

PR #768, opened 2021-05-21
Add zip structure checks for user friendliness

InfoTeddy:
If a level zip is named `LEVELNAME.zip`, the level file inside it must also be named `LEVELNAME.vvvvvv`, else custom assets won't work.

This is because when we mount the zip file, we simply add `LEVELNAME.vvvvvv` to the levels directory. Then whenever we load `LEVELNAME.vvvvvv`, we look at the filename, remove the extension, and look for the assets inside the zip of the same name, `LEVELNAME.zip`.

As a result, if someone were to make a level zip with assets but mismatch the filename, the assets wouldn't load. Furthermore, if someone were to add extra levels in the same zip, they wouldn't have any assets load for them as well, which could be confusing.

To make things crystal-clear to the user, we now filter out any zips that have incorrect structures like that, and print a message to the terminal. Unfortunately nothing gets shown for non-terminal users, but at least doing this and filtering out the zips is less confusing than letting them through but with the issues mentioned above.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -11,6 +11,7 @@
 #include "Exit.h"
 #include "Graphics.h"
 #include "Maths.h"
+#include "Unused.h"
 #include "UtilityClass.h"
 
 /* These are needed for PLATFORM_* crap */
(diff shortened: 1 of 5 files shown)


InfoTeddy:
I did not notice the title was like that, sorry about that...

PR #774, opened 2021-06-12
Prevent losing level data with rollcredits

InfoTeddy:
When `rollcredits` is ran during in-editor playtesting, all unsaved data is lost. To prevent this, just return to the editor if `rollcredits` is ran, with a note saying "Rolled credits".

## Legal Stuff:

  for these changes


TerryCavanagh:
Good catch, this one!

PR #777, opened 2021-06-12
Outline time trial countdown

InfoTeddy:
This makes the text much more readable against certain backgrounds (if you have text outline enabled), especially against the Warp Zone background (when you start in "This is how it is").

Without outline:
Warp Zone time trial without outline
Bad contrast, especially when the text glow turns blue.

With outline:
Warp Zone time trial with outline
Excellent contrast.

## Legal Stuff:

  for these changes


TerryCavanagh:
Ah, yep, good call here.

PR #778, opened 2021-06-12
Use hiddenname for hardestroom if roomname is empty

InfoTeddy:
In earlier 2.3, if the roomname was empty, Dimension VVVVVV was used instead. However, instead of doing that, it's better to just use the hiddenname instead. Both because it's less hardcoded, and some rooms have hidden names that aren't Dimension VVVVVV.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -3962,7 +3962,7 @@ void Game::gethardestroom(void)
         }
         else if (map.roomname == "")
         {
-            hardestroom = "Dimension VVVVVV";
+            hardestroom = map.hiddenname;
         }
     }
 }


TerryCavanagh:
Hmm - could you give me an example of when this might have a different value?

InfoTeddy:
The Ship has a different hiddenname.

TerryCavanagh:
So, the hypothetical benefit would be something like: if the player somehow died on the ship more than in the open dimension, the hardest room would read "The Ship" instead?

I guess this is harmless, but I don't get the point...

InfoTeddy:
The player can press R on the ship. And for what it's worth, I'm planning on adding hidden names to the Secret Lab rooms.

TerryCavanagh:
Ok, sure! Let's do it.

PR #782, opened 2021-06-13
Support for OpenBSD

strikersh:
## Changes:

- use fseeko and ftello like FreeBSD in tinyxml2
- use current directory as basePath if NULL (OpenBSD doesn't actually support this feature it is disabled via a patch in their ports)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -148,8 +148,8 @@ int FILESYSTEM_init(char *argvZero, char* baseDir, char *assetsPath)
 
 	if (basePath == NULL)
 	{
-		puts("Unable to get base path!");
-		return 0;
+		puts("Unable to determine base path, falling back to current directory");
+		basePath = SDL_strdup("./");
 	}
 
 	/* Mount the stock content last */
--- a/third_party/tinyxml2/tinyxml2.cpp
+++ b/third_party/tinyxml2/tinyxml2.cpp
@@ -103,7 +103,7 @@ distribution.
 #if defined(_WIN64)
 	#define TIXML_FSEEK _fseeki64
 	#define TIXML_FTELL _ftelli64
-#elif defined(__APPLE__) || (__FreeBSD__)
+#elif defined(__APPLE__) || (__FreeBSD__) || (__OpenBSD__)
 	#define TIXML_FSEEK fseeko
 	#define TIXML_FTELL ftello
 #elif defined(__unix__) && defined(__x86_64__)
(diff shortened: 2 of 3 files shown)


third_party/tinyxml2/tinyxml2.cpp
```diff
@@ -103,7 +103,7 @@ distribution.
 #if defined(_WIN64)
 	#define TIXML_FSEEK _fseeki64
 	#define TIXML_FTELL _ftelli64
-#elif defined(__APPLE__) || (__FreeBSD__)
+  #elif defined(__APPLE__) || (__FreeBSD__) || (__OpenBSD__)
```
flibitijibibo, line 106:
```suggestion
#elif defined(__APPLE__) || (__FreeBSD__) || (__OpenBSD__)
```

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -148,8 +148,7 @@ int FILESYSTEM_init(char *argvZero, char* baseDir, char *assetsPath)
 
 	if (basePath == NULL)
 	{
-		puts("Unable to get base path!");
-		return 0;
+		basePath = SDL_strdup("./");
```
flibitijibibo, line 151: Having this as a fallback is okay but we need to have a warning message noting that we had to fall back to ".", though it's probably worth fixing this in PhysFS and SDL.

flibitijibibo (changes requested):
You can also add yourself to `CONTRIBUTORS` if you want, other than the notes this seems okay.

desktop_version/src/FileSystemUtils.cpp
strikersh: The issue is that OpenBSD kernel doesn't support this feature see : https://openports.se/devel/sdl2

flibitijibibo:
Looks good, thanks!

third_party/tinyxml2/tinyxml2.cpp
```diff
@@ -103,7 +103,7 @@ distribution.
 #if defined(_WIN64)
 	#define TIXML_FSEEK _fseeki64
 	#define TIXML_FTELL _ftelli64
-#elif defined(__APPLE__) || (__FreeBSD__)
+#elif defined(__APPLE__) || (__FreeBSD__) || (__OpenBSD__)
```
InfoTeddy, line 106: Shouldn't this be a PR to upstream? Now if people compile with `-DBUNDLE_DEPENDENCIES=OFF` (as package maintainers might want), their TinyXML-2 will use the wrong fseek/ftell.

flibitijibibo: Oh, missed that this was thirdparty - @strikersh, are you able to send this to upstream tinyxml2?

strikersh: submitted it there, and added Net and DragonFly bsd too (fseeko64 isn't available on bsds)

PR #787, opened 2021-06-19
Custom levels: only count inbounds trinkets and crewmates

InfoTeddy:
In the past, people have reported having glitched levels where they can't get the trinket star or can't complete the level because the number of trinkets or crewmates is one higher than what can be obtained in the level.

How did this happen? Well, it turns out that if you place an entity, and then resize the level to be smaller, that entity still exists. This is inconsequential for most entities, but if the entity is a trinket or crewmate, that entity is still counted towards the number of trinkets or crewmates in the level.

One fix would be to just remove entities whenever the level is downsized, but then if someone accidentally downsizes the level and wants to go back, that entity will be gone. Plus, it would be inconsistent with tiles, because tiles don't get removed when you downsize the level. Also, it wouldn't fix existing levels where people have managed to place trinkets or crewmates out of bounds.

So instead, `ed.numtrinkets()` and `ed.numcrewmates()` should simply ignore trinkets and crewmates that are outside the playable area. That way, levels with glitched trinkets and crewmates can still be completed, and can still be completed with the trinket star.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/editor.cpp
+++ b/desktop_version/src/editor.cpp
@@ -5895,12 +5895,21 @@ Uint32 editorclass::getonewaycol(void)
     return graphics.getRGB(255, 255, 255);
 }
 
+static bool inbounds(const edentities* entity)
+{
+    extern editorclass ed;
+    return entity->x >= 0
+    && entity->y >= 0
+    && entity->x < ed.mapwidth * 40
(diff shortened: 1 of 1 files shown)


flibitijibibo:
RIP any levels that were paying tribute to basically every level in Sonic Adventure 2

PR #790, opened 2021-07-30
Add in-game timer option

lsaa:
## Changes:
Since the game's speedruns are timed based on the IGT it would be useful to display it during gameplay.

How it looks:
Hello

It displays behind textboxes and the cutscene screen border effect, doesn't display on super gravitron and the game complete screen
Render Order

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -380,6 +380,7 @@ void Game::init(void)
     fadetolabdelay = 0;
 
     over30mode = true;
+    showingametimer = false;
 
     ingame_titlemode = false;
 #if !defined(NO_CUSTOM_LEVELS) && !defined(NO_EDITOR)
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -6103,7 +6103,7 @@ void Game::createmenu( enum Menu::MenuName t, bool samemenu/*= false*/ )
         option("input delay");
         option("interact button");
         option("fake load screen");
-        option("show in game timer");
+        option("toggle in game timer");
         option("return");
         menuyoff = 0;
         maxspacing = 15;
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -125,7 +125,6 @@ void Game::init(void)
     edteleportent = 0; //Added in the port!
     companion = 0;
     roomchange = false;
-    isingamecompletescreen = false;
 
 
     quickrestartkludge = false;
@@ -2591,7 +2590,6 @@ void Game::updatestate(void)
             break;
         case 3501:
             //Game complete!
-            isingamecompletescreen = true;
             unlockAchievement("vvvvvvgamecomplete");
             unlocknum(5);
             crewstats[0] = true;
@@ -2807,7 +2805,6 @@ void Game::updatestate(void)
             map.finalstretch = false;
 
             graphics.setbars(320);
-            isingamecompletescreen = false;
 
             teleport_to_new_area = true;
             teleportscript = "gamecomplete";
@@ -2815,7 +2812,6 @@ void Game::updatestate(void)
 
         case 3520:
             //NO DEATH MODE COMPLETE JESUS
-            isingamecompletescreen = true;
             hascontrol = false;
             crewstats[0] = true;
 
@@ -2826,7 +2822,6 @@ void Game::updatestate(void)
             if(graphics.fademode == 1)	state++;
             break;
         case 3522:
-            isingamecompletescreen = false;
             copyndmresults();
             quittomenu();
             createmenu(Menu::nodeathmodecomplete);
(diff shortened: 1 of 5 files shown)


desktop_version/src/Render.cpp
```diff
@@ -623,6 +623,13 @@ static void menurender(void)
             else
                 graphics.Print(-1, 65, "Fake loading screen is ON", tr, tg, tb, true);
             break;
+        case 4:
+            graphics.bigprint(-1, 30, "In Game Timer", tr, tg, tb, true);
+            if (game.showingametimer)
```
InfoTeddy, line 628: Should probably have braces around these single-line statements.

desktop_version/src/Render.cpp
```diff
@@ -1644,6 +1651,11 @@ void gamerender(void)
         }
     }
 
+    if (graphics.fademode==0 && !game.intimetrial && !game.isingamecompletescreen && game.swngame != 1 && game.showingametimer) {
```
InfoTeddy, line 1654: Opening braces should have their own line.

desktop_version/src/Render.cpp
```diff
@@ -1644,6 +1651,11 @@ void gamerender(void)
...
+        graphics.bprint(6, 6, "TIME :",  255,255,255);
```
InfoTeddy, line 1655: Why the space?

InfoTeddy (changes requested):
Looks good, just some minor style things. I don't think this is going to be merged in 2.3, since the scheduled release date is going to be in a bout a week (@flibitijibibo can confirm),

desktop_version/src/Render.cpp
lsaa: gotcha, the one above it didn't have braces too so I just did it like that

desktop_version/src/Render.cpp
lsaa: Removed the space

desktop_version/src/Render.cpp
lsaa: new line added

desktop_version/src/Game.cpp
```diff
@@ -124,6 +124,7 @@ void Game::init(void)
     edteleportent = 0; //Added in the port!
     companion = 0;
     roomchange = false;
+    isingamecompletescreen = false;
```
InfoTeddy, line 127: Instead of introducing another variable, you can probably just check for `game.state` instead. That way, if the Game Complete gamestate sequence gets interrupted, the timer won't go missing.

InfoTeddy (changes requested):

desktop_version/src/Game.cpp
lsaa: Alright, gonna try to figure out all the game complete states so it doesn't end up flashing on and off

InfoTeddy: Ok, well, I'll help you out. It should just be sufficient to do `(game.state >= 3501 && game.state <= 3518) || (game.state >= 3520 && game.state <= 3522)`.

lsaa: that should work yeah

```
bool Game::isingamecompletescreen()
{
    return (game.state >= 3501 && game.state <= 3518) || (game.state >= 3520 && game.state <= 3522);
}
```

do you think having it as a class method is overkill?

InfoTeddy: Not overkill at all - I do the same elsewhere, e.g. `incompetitive()`, `inspecial()`.

lsaa: Done :+1:

flibitijibibo:
This looks better than I was expecting - I was thinking it would be more invasive but it's actually pretty isolated! I think we can sneak this in with a week to test...

Only one thing: The new bool needs to be initialized to false, in case the XML doesn't have an entry yet.

flibitijibibo (approved):
Alright, this makes sense to me - if someone wants to test first, I probably won't be around to merge this until Monday while I wrap up SDL 2.0.16 and FNA 21.08.

lsaa:
There was a huge gap between "TIME:" and the current time
Looks like this now
image

flibitijibibo:
Any last review here before it goes in?

We'll need a rebase of this first, but aside from that I don't see anything wrong in here.

InfoTeddy:
I just quickly compiled and tested it, it seems fine enough to me.

lsaa:
I played the game through to the end a few times with it and didn't find any cases where it blocks text or shouldn't be on. Also took a look in the level editor and it seemed ok but I don't have much experience with it so I could've forgotten to check some screens.

InfoTeddy:
The level editor uses a different state (and different render function), and since the timer is only rendered in `gamerender`, you don't need to worry about it displaying on a level editor screen. However it will show up during playtesting in the level editor (since that uses the actual "game" state).

flibitijibibo:
Looks good, thanks for checking this out!

PR #793, opened 2021-08-05
Split glitchrunner mode into multiple versions

InfoTeddy:
Previously, turning glitchrunner mode on essentially locked you to emulating 2.0, and turning it off just meant normal 2.3 behavior. But what if you wanted 2.2 behavior instead? Well, that's what I had to ask when a TAS of mine would desync in 2.3 because of the two-frame delay fix (glitchrunner off), but would also desync because of 2.0 warp lines (glitchrunner on).

What I've done is made it so there are three states to glitchrunner mode now: 2.0 (previously just the "on" state), 2.2 (previously a state you couldn't use), and "off". Furthermore, I made it an enum, so in case future versions of the game patch out more glitches, we can add them to the enum (and the only other thing we have to update is a lookup table in `GlitchrunnerMode.c`). Also, 2.2 glitches exist in 2.0, so you'll want to use `GlitchrunnerMode_less_than_or_equal()` to check glitchrunner version.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/GlitchrunnerMode.c
+++ b/desktop_version/src/GlitchrunnerMode.c
@@ -1,0 +1,69 @@
+#include "GlitchrunnerMode.h"
+
+#include <SDL_assert.h>
+#include <SDL_stdinc.h>
+
+#define LOOKUP_TABLE \
+    FOREACH_ENUM(GlitchrunnerNone, "") \
+    FOREACH_ENUM(Glitchrunner2_0, "2.0") \
+    FOREACH_ENUM(Glitchrunner2_2, "2.2") \
+
+const char* GlitchrunnerMode_enum_to_string(const enum GlitchrunnerMode mode)
+{
+    switch (mode)
+    {
+#define FOREACH_ENUM(MODE, STRING) \
+    case MODE: \
+        return STRING;
+
+    LOOKUP_TABLE
+
+#undef FOREACH_ENUM
+
+    /* Compiler raises warning about this enum not being handled. */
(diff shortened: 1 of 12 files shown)


desktop_version/src/GlitchrunnerMode.c
```diff
@@ -0,0 +1,71 @@
...
+#undef FOREACH_ENUM
+
+    /* Compiler raises warning about this enum not being handled. */
+    case GlitchrunnerNumVersions:
+        break;
+    }
+
+    /* It's impossible to get here, but a compiler warning still gets raised
+     * if we don't return. */
+    SDL_assert(0 && "Passed non-existent GlitchrunnerMode!");
```
flibitijibibo, line 30: This might be a problem for existing config files... do we want an emergency fallback for "1" and "0" (or true/false, whatever it was) to assume that it's either going all for 2.0 or none?

InfoTeddy: It's not a problem. When we parse the config file, we pass it through `GlitchrunnerMode_string_to_enum()`, and it defaults to returning `GlitchrunnerNone` if it's not a valid string. So by the time we reach `GlitchrunnerMode_enum_to_string()` it's definitely going to be a valid enum. Probably the only way to trigger that assert would be do to something stupid like `GlitchrunnerMode_enum_to_string((enum GlitchrunnerMode) 69);`. That said, I'll remove the comment, since I don't think it's "impossible" in the sense that I meant.

PR #799, opened 2021-08-05
Add unfocus audio pause option

InfoTeddy:
Some people prefer the 2.2 behavior where unfocusing pauses the game, but the music still plays. One such person is Trinket9 on the VVVVVV Discord server, who wanted it that way.

The reason audio pausing was added in the first place was to prevent desyncing music in levels with cutscenes that synced to music. Rather than reverting it, let's add this option instead.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -390,6 +390,7 @@ void Game::init(void)
     slidermode = SLIDER_NONE;
 
     disablepause = false;
+    disableaudiopause = false;
     inputdelay = false;
 }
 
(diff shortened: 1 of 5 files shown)


AllyTally:
I wonder if this could be made into a setting levels can force on or off with a command or something...?

flibitijibibo:
> I wonder if this could be made into a setting levels can force on or off with a command or something...?

This does seem reasonable - in other games this is a common issue with cutscenes where pausing isn't possible, so it has to either let it go or pause the entire program including sound. Actually _implementing_ it, on the other hand... might be an annoying mishmash of bools.

flibitijibibo:
This patch is pretty clear, so will get it in now... @AllyTally, definitely file a new issue for the level setting so that we don't forget to do this. Not sure if 2.3 will have it but 2.4 really should.

PR #800, opened 2021-08-05
Always reset player when starting gamemode

InfoTeddy:
This fixes a bug where the player would always be facing right if they were loading in for the first time. This essentially made them always ignore the facing direction set in the save file if the facing direction was leftwards.

The problem is facing direction only gets set in `map.resetplayer()`, but if loading in for the first time, that path is never taken (unless you are loading a main game quicksave that's inside a tower). The solution is to always reset the player, even after creating them for the first time.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -2654,10 +2654,7 @@ void scriptclass::startgamemode( int t )
 		{
 			obj.createentity(game.savex, game.savey, 0, 0); //In this game, constant, never destroyed
 		}
-		else
-		{
-			map.resetplayer();
-		}
+		map.resetplayer();
 		map.gotoroom(game.saverx, game.savery);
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This seems like a bigger change than it sounds... is this 2.2 behavior?

InfoTeddy:
No, in fact this bug also exists in 2.2.

However since 2.2, `map.resetplayer()` also sets `game.lifeseq` to 0 (unless it's after a death), which fixes another inconsistency. And, anyways, `map.resetplayer()` *already* always gets called for each subsequent load-in after the first one, so if there were any bugs from always calling `map.resetplayer()` we'd know about them already.

flibitijibibo:
Very well, wasn't sure if we'd run into an issue where too much stuff got overridden on reset.

PR #808, opened 2021-08-07
Be more user-friendly if levels have asset/zip issues

InfoTeddy:
This PR does the following:

1. Gracefully handles a custom level having an asset issue after its assets are loaded, by returning to the title screen and printing the error for the player.
2. Displays a message warning about improperly-structured zips, if there are any, when the levels list is loaded.

They both make it so users don't need a console in order to be notified of what's wrong with their levels. This is good, because not everyone uses a console.

Also, it's a bit rude to close the game and bail completely if a level has an issue with its assets, so I've fixed that too.

Bad level asset

Bad zip structure

Both of these depend on some new text wrapping functions, which is why they're both in this PR.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -151,6 +151,9 @@ void Graphics::init(void)
     tiles2_mounted = false;
     minimap_mounted = false;
 #endif
+
+    SDL_zeroa(error);
+    SDL_zeroa(error_title);
 }
 
 void Graphics::destroy(void)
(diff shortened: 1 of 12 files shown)


desktop_version/src/Graphics.cpp
```diff
@@ -515,6 +515,118 @@ void Graphics::PrintAlpha( int _x, int _y, std::string _s, int r, int g, int b,
...
+    *len = 0;
+
+    if (str[idx] == '\0')
+    {
+        return false;
+    }
+
+    while (true)
+    {
+        if ((str[idx] & 0xC0) == 0x80)
```
Dav999-v, line 540: This is not really future-proof and overall a bit of a "dirty" hack, will all non-ASCII characters always have the same width? My idea for future CJK localizations _would_ probably work with this since all characters (including ASCII) would just be 10x10 or 12x12, but that's not a good assumption to make since we _do_ have a dynamic `Graphics::bfontlen` and it takes a `uint32_t` for a reason.

InfoTeddy: This would have to be changed later to actually concatenate the bytes together into a `uint32_t`, yes. The problem is I can't use UTF-CPP on C strings, so I would have to reimplement checking the bytes and concatenating them together (or find another library), and I don't need to do that now.

flibitijibibo: This should be okay for 2.3, but will definitely need a FIXME so it can be found and addressed when necessary.

InfoTeddy: Added a FIXME with a force-push.

InfoTeddy:
Well I'm stupid again.

PR #811, opened 2021-08-07
 Be more user-friendly if levels have asset/zip issues

InfoTeddy:
#808 

This is the second time I've accidentally deleted an open PR branch, let's try this again...

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/UtilityClass.h
+++ b/desktop_version/src/UtilityClass.h
@@ -68,6 +68,16 @@ void _VVV_between(
         sizeof(middle) \
     )
 
+#ifndef __has_attribute
+#   define __has_attribute(x) 0
+#endif
+
+#if __has_attribute(__fallthrough__)
+#   define VVV_fallthrough __attribute__((__fallthrough__))
+#else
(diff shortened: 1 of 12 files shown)


desktop_version/src/UtilityClass.h
```diff
@@ -68,6 +68,25 @@ void _VVV_between(
         sizeof(middle) \
     )
 
+#ifndef __has_attribute
+#   define __has_attribute(x) 0
+#endif
+
+#if __has_attribute(__fallthrough__)
```
flibitijibibo, line 75: This might need to be named something less generic like `VVV_case_fallthrough` so that it doesn't accidentally step on a commonly-used word.

flibitijibibo (changes requested):
Other than the fallthrough macro this makes sense - we would need this level of robustness for 10-foot and mobile builds at minimum.

desktop_version/src/UtilityClass.h
InfoTeddy: Done.

PR #813, opened 2021-08-12
Add audiopause(on/off) command

AllyTally:
## Changes:
If you have unfocus pause off, and unfocus audio pause off, then this command will go into effect. When it's set to on, the audio will pause when you unfocus the game. When it's set to off, the audio will not. This is different from the setting, and gets saved to the save file.
This was created in response to #799 being merged.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -391,6 +391,7 @@ void Game::init(void)
 
     disablepause = false;
     disableaudiopause = false;
+    disabletemporaryaudiopause = true;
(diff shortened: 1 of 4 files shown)


InfoTeddy (approved):
Looks good to me!

flibitijibibo:
Pleasantly surprised, usually when audio stuff gets involved it gets messy...

Thanks for writing this so quickly! Odds are I'll do a rebuild tomorrow since I need to squeeze in a Wayland fix or two.

PR #814, opened 2021-08-12
Set Secret Lab/Super Gravitron hidden names

InfoTeddy:
It seems appropriate that these rooms should have their names shown on the map screen.

## Legal Stuff:

  for these changes


flibitijibibo:
Code makes sense, just need a quick art/writing check with @TerryCavanagh...

PR #823, opened 2021-08-12
Refactor colors in internal commands

AllyTally:
## Changes:

Originally this started as a "deduplicate a bunch of duplicated code in script commands" PR, but as I was working on that, I discovered there's a lot more that needs to be done than just deduplication. Anything which needs a crewmate entity now calls `getcrewmanfromname(name)`, and anything which just needs the crewmate's color calls `getcolorfromname(name)`. This was done to make sure that everything works consistently and no copy/pasting is required. Next is the fallback; instead of giving up and doing various things when it can't find a specific color, it now attempts to treat the color name as an ID, and if it can't then it returns -1, where each individual command handles that return value. This means we can keep around AEM -- a bug used in custom levels -- by not doing anything with the return value if it's -1.

Also, for some reason, there were two `crewcolour` functions, so I stripped out the one in entityclass and left (and modified) the one in the graphics class, since the graphics class also has the `crewcolourreal` function.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -1799,7 +1799,7 @@ void scriptclass::run(void)
 			else if (words[0] == "createlastrescued")
 			{
 				r = graphics.crewcolour(game.lastsaved);
-				if (r <= 1)
+				if (r == 1 || r == PURPLE)
 				{
 					r = GRAY; // Default to gray if invalid color.
 				}
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -89,6 +89,33 @@ void scriptclass::tokenize( const std::string& t )
 	}
 }
 
+static int getcolorfromname(std::string name)
+{
+	if      (name == "player")     return CYAN;
+	else if (name == "cyan")       return CYAN;
+	else if (name == "red")        return RED;
+	else if (name == "green")      return GREEN;
+	else if (name == "yellow")     return YELLOW;
+	else if (name == "blue")       return BLUE;
+	else if (name == "purple")     return PURPLE;
+	else if (name == "customcyan") return CYAN;
+	else if (name == "gray")       return GRAY;
+	else if (name == "teleporter") return TELEPORTER;
+
+	int color = help.Int(name.c_str(), -1);
+	if (color < 0) return -1; // Not a number (or it's negative), so we give up
+	return color; // Last effort to give a valid color, maybe they just input the color?
+}
+
+static int getcrewmanfromname(std::string name)
+{
+	if (name == "player") return obj.getplayer(); //  Return the player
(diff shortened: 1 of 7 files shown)


desktop_version/src/Script.cpp
```diff
@@ -2114,25 +1798,28 @@ void scriptclass::run(void)
...
-				else if (game.lastsaved==4)
-				{
-					r=13;
-				}
-				else if (game.lastsaved==5)
-				{
-					r=16;
-				}
-				else
+				int crew_color = 1;
```
InfoTeddy, line 1801: Why does this default to 1? Also, it doesn't seem like `crew_color` is used here. And I'm wondering why the `r` assignments don't just use the color enums.

InfoTeddy (changes requested):
Looks good to me, except for a few things.

desktop_version/src/Script.cpp
AllyTally: I've replaced that entire section with `graphics.crewcolour()`, which I seem to have forgotten was a thing. Whoops.

desktop_version/src/Script.cpp
```diff
@@ -2114,25 +1798,10 @@ void scriptclass::run(void)
...
-				{
-					r=13;
-				}
-				else if (game.lastsaved==5)
-				{
-					r=16;
-				}
-				else
+				r = graphics.crewcolour(game.lastsaved);
+				if (r <= 1)
```
InfoTeddy, line 1802: This `r <= 1` is a bit weird. I presume you were trying to blacklist Violet's color (20) along with Viridian's? But it doesn't do that, because 20 is greater than 1.

InfoTeddy (changes requested):

desktop_version/src/Script.cpp
AllyTally: Whoops.

desktop_version/src/Script.cpp
```diff
@@ -2114,25 +1798,10 @@ void scriptclass::run(void)
...
-				{
-					r=13;
-				}
-				else if (game.lastsaved==5)
-				{
-					r=16;
-				}
-				else
+				r = graphics.crewcolour(game.lastsaved);
+				if (r == 1 || r == PURPLE)
```
InfoTeddy, line 1802: Not sure what this `r == 1` here. `graphics.crewcolour()` will return either 0, 20, 14, 15, 13, or 16. So this doesn't blacklist Viridian's color.

InfoTeddy (changes requested):

desktop_version/src/Script.cpp
AllyTally: oh my god i meant to write 0

PR #835, opened 2021-09-01
Call fsync on Emscripten

leo60228:
## Changes:

This is necessary to allow pages consuming VVVVVV.js to persist save data. See commit messages for further details.

## Legal Stuff:

  for these changes


flibitijibibo:
Do we want to treat this as a PhysFS issue? The code gets pretty specific, but at the same time this looks _exactly_ like another platform's commit system, so maybe send this PR to the PhysFS issue tracker and see what Ryan thinks.

leo60228:
That could make sense, but in my opinion the optimization to prevent all code using PhysFS from being affected by asyncify is important, and PhysFS can't really do that without knowledge of the host program that it doesn't have.

flibitijibibo:
That makes sense to me - on the other platform we added an extension where you just call `CommitChanges()` and it effectively does what's done here; you call it after making all the filesystem updates you expect to make and the async work is done internally.

leo60228:
That's fair. For what it's worth, if my wild guess as to the platform you're referring to is correct, I'm pretty sure that the OS isn't smart enough to process multiple filesystem calls in parallel (but I could be misremembering).

leo60228:
Ugh, I misread the PhysFS code. It automatically flushes when calling `PHYSFS_close`. That makes things a *bit* trickier, but I think I can still work around that.

leo60228:
The more I look at this, the less I think that it can be easily handled. Asyncify doesn't have any way to handle functions that may or may not be async (like `PHYSFS_close`), so there's no real way to restrict what asyncify applies to. Because of this, I'll close this PR and let consumers add `-s ASYNCIFY` to CFLAGS if they need persistence.

PR #838, opened 2021-09-01
Call FS.syncfs on Emscripten

leo60228:
## Changes:

While thinking about #835, I realized that I can sidestep asyncify entirely by directly calling the relevant functions via inline JS.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -879,7 +879,14 @@ bool FILESYSTEM_saveTiXml2Document(const char *name, tinyxml2::XMLDocument& doc)
 	PHYSFS_close(handle);
 
 #ifdef __EMSCRIPTEN__
-	EM_ASM(FS.syncfs(false, function(err) { if (err) abort(err); }));
+	EM_ASM(FS.syncfs(false, function(err)
+	{
+		if (err)
+		{
+			console.warn("Error saving:", err);
+			alert("Error saving. Check console for more information.");
+		}
+	}));
 #endif
 
 	return true;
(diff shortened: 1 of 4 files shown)


desktop_version/src/FileSystemUtils.cpp
```diff
@@ -872,6 +877,11 @@ bool FILESYSTEM_saveTiXml2Document(const char *name, tinyxml2::XMLDocument& doc)
 	}
 	PHYSFS_writeBytes(handle, printer.CStr(), printer.CStrSize() - 1); // subtract one because CStrSize includes terminating null
 	PHYSFS_close(handle);
+
+#ifdef __EMSCRIPTEN__
+	EM_ASM(FS.syncfs(false, function(err) { if (err) abort(err); }));
```
flibitijibibo, line 882: Is there something better than abort()? It'd be cool to be able to return false, but since this is async...

leo60228: Emscripten's `abort` throws a JS exception, meaning that it's quite a bit more debugging-friendly than on native (you get a stack trace, though admittedly not an especially useful one if you don't have debug symbols).

flibitijibibo: Dug into SDL a bit, since it's pretty much the only window into emscripten that I have, and it seems like we can at least do something where we assign a bool and return that... an example of assigning from inline js:

https://github.com/libsdl-org/SDL/blob/main/src/video/emscripten/SDL_emscriptenvideo.c#L175-L177

If there's another wasm expert reading this I'd be very interested in their opinion here; I'm just reviewing this from a generic platform engineer's perspective.

leo60228: The fundamental issue here is that WASM can't block on the FS.syncfs call. Asyncify automatically transforms code to use callbacks, but it causes a sizeable performance and code size penalty. We could maybe set a flag that there was an error and check it at the beginning of each frame, but I'm not even really sure what we would do to handle the error after the function causing it already returned.
If a non-fatal error is desired, we could just do `console.warn` instead of `abort`.

leo60228: Actually, I think something like `console.warn('Error saving:', err); alert('Error saving. Check console for more information.');` would be best.

PR #847, opened 2021-10-13
Add `setactivityposition(x,y)`, add new textbox color `transparent`

AllyTally:
## Changes:

This commit adds a new internal command as a part of the visual activity zone changes I've been making.
This one allows the user to reposition the activity zone prompt to anywhere on the screen.
In addition, this commit adds the textbox color `transparent`, which just sets r, g and b to 0.
rgb(0, 0, 0) normally creates the color black, however in VVVVVV textboxes, it makes the background
of them invisible, and makes the text the off-white color which the game uses elsewhere.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1093,8 +1093,8 @@ void entityclass::createblock( int t, int xp, int yp, int w, int h, int trig /*=
     }
     else
     {
-        block.activity_x = 16;
-        block.activity_y = 4;
+        block.activity_x = 0;
+        block.activity_y = 0;
     }
 
     if (!reuse)
--- a/desktop_version/src/Render.cpp
+++ b/desktop_version/src/Render.cpp
@@ -1985,14 +1985,16 @@ void gamerender(void)
             game.activity_lastprompt.c_str()
         );
 
+        int centered_x = ((160 ) - ((graphics.len(final_string)) / 2));
+
         if (game.activity_r == 0 && game.activity_g == 0 && game.activity_b == 0)
         {
-            graphics.bprint(game.activity_x - 11, game.activity_y + 8, final_string, 196*act_alpha, 196*act_alpha, (255 - help.glow)*act_alpha, true);
+            graphics.bprint(centered_x + game.activity_x, game.activity_y + 12, final_string, 196*act_alpha, 196*act_alpha, (255 - help.glow)*act_alpha);
         }
         else
         {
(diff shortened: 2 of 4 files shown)


desktop_version/src/Entity.cpp
```diff
@@ -71,6 +71,8 @@ void entityclass::init(void)
     customenemy = 0;
     customwarpmode = false; customwarpmodevon = false; customwarpmodehon = false;
     customactivitycolour = "";
+    customactivitypositionx = -1;
+    customactivitypositiony = -1;
```
InfoTeddy, line 75: These need to be reset in `scriptclass::hardreset()`.

desktop_version/src/Render.cpp
```diff
@@ -1985,8 +1985,15 @@ void gamerender(void)
...
-        graphics.Print(5, 12, final_string, game.activity_r*act_alpha, game.activity_g*act_alpha, game.activity_b*act_alpha, true);
+        if (game.activity_r == 0 && game.activity_g == 0 && game.activity_b == 0)
+        {
+            graphics.bprint(game.activity_x - 11, game.activity_y + 8, final_string, 196*act_alpha, 196*act_alpha, (255 - help.glow)*act_alpha, true);
+        }
+        else
+        {
+            graphics.drawtextbox(game.activity_x, game.activity_y, 36, 3, game.activity_r*act_alpha, game.activity_g*act_alpha, game.activity_b*act_alpha);
+            graphics.Print(game.activity_x - 11, game.activity_y + 8, final_string, game.activity_r*act_alpha, game.activity_g*act_alpha, game.activity_b*act_alpha, true);
+        }
```
InfoTeddy, line 1996: It might be hard to spot (I've been strongly considering refactoring all the print callsites), but what's happening here is that the text in the prompt is always centered and thus the x-position doesn't matter (including transparent prompts), whereas the surrounding text box is not centered at all and thus the x-position affects it there.

I'm not sure what your intentions here are with regards to horizontal positioning, but this is at the very least an inconsistency.

Dav999-v: > I'm not sure it makes much difference. Either you have to rebase and fix conflicts on all of my work, or I have to rebase and fix conflicts on all of your work.

I meant that you could hold off on refactoring print callsites until after the localization branch is merged, so you can directly base the changes on that - as you already suggest is a possibility in the second paragraph of your comment.

> Besides, it's virtually guaranteed you'll run into conflicts when maintaining an unmerged fork for extremely long periods of time.

I know, I have already encountered and resolved a lot of conflicts, I'm just saying that the print lines would cause exceedingly high amounts of them so it'd be better to avoid those extra hours of concentrated staring at little details.

PR #858, opened 2021-12-25
Simplify time formatting functions

Dav999-v:
## Changes:

(The commit message says it all:)

Here's my notes on all the existing functions and what kind of time
formats they output:

- Game::giventimestring(int hrs, int min, int sec)
	H:MM:SS
	MM:SS

- Game::timestring()
// uses game.hours/minutes/seconds
	H:MM:SS
	MM:SS

- Game::partimestring()
// uses game.timetrialpar (seconds)
	MM:SS

- Game::resulttimestring()
// uses game.timetrialresulttime (sec) + timetrialresultframes (1/30s)
	MM:SS.CC

- Game::timetstring(int t)
// t = seconds
	MM:SS

- Game::timestringcenti(char* buffer, const size_t buffer_size)
// uses game.hours/minutes/seconds/frames
	H:MM:SS.CC
	MM:SS.CC

- UtilityClass::timestring(int t)
// t = frames, 30 frames = 1 second
	S:CC
	M:SS:CC

This is kind of a mess, and there's a lot of functions that do the same
thing except using different variables. For localization, I also want
translators to be able to localize all these time formats - many
languages use the decimal comma instead of the decimal point (12:34,56)
maybe some languages really prefer something like 1時02分11秒44瞬...
Which I don't know to be correct, but it's good to be prepared for it
and not restrict translators arbitrarily to only changing ":" and "."
when we can start making the system better in the first place.

I added a new function, UtilityClass::format_time. This is the place
where all time formats come together, given the number of seconds and
optionally frames. I have simplified the above-mentioned functions
somewhat, but I haven't given them a complete refactor or renaming -
I mainly made sure that they all use the same backend so I can make the
formats consistent and properly localizable.

(And before we start shoving more temporary char buffers everywhere
just to get rid of the std::string's, maybe we need to think of a
globally used working buffer of size SCREEN_WIDTH_CHARS+1, as a
register of sorts, for when any line of text needs to be made or
processed, then printed, and then goes unused. Maybe help.textrow,
or something like that.)

As for this commit, the available time formats are now more consistent
and changed a little in some places. Leading zeroes for the first unit
are now no longer included, time trial results and the Super Gravitron
can now display hours when they went to 60 minutes before, and we now
always use .CC instead of :CC. These are the formats:
- H:MM:SS
- H:MM:SS.CC
- M:SS
- M:SS.CC
- S.CC  (only used when always_minutes=false, for the Gravitrons)

Here's what changes to the current functions:
- Game::partimestring() is removed - it was used in two places, and
  could be replaced by game.timetstring(game.timetrialpar)
- Game::giventimestring(h,m,s) and Game::timestring() are now wrappers
  for the other functions
- The four remaining functions (Game::resulttimestring(),
  Game::timetstring(t), Game::timestringcenti(buffer, buffer_size)
  and UtilityClass::timestring(t)) are now wrappers for the "central
  function", UtilityClass::format_time.
- UtilityClass::twodigits(int t) is now unused so it's also removed.
- I also added int UtilityClass::hms_to_seconds(int h, int m, int s)

### Spot the differences:

In a time trial Time trial menu In-game stats Super Gravitron highscore screen In the Super Gravitron In the normal Gravitron

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -1412,7 +1412,7 @@ void Game::updatestate(void)
             obj.removetrigger(82);
             hascontrol = false;
 
-            timetrialresulttime = seconds + (minutes * 60) + (hours * 60 * 60);
+            timetrialresulttime = help.hms_to_seconds(hours, minutes, seconds);
             timetrialresultframes = frames;
             timetrialresulttrinkets = trinkets();
             timetrialresultshinytarget = timetrialshinytarget;
@@ -2605,13 +2605,16 @@ void Game::updatestate(void)
             graphics.textboxcenterx();
             break;
         case 3502:
+        {
             state++;
             statedelay = 45+15;
 
             graphics.createtextboxflipme("  All Crew Members Rescued!  ", -1, 64, 0, 0, 0);
-            savetime = timestring();
-            savetime += "." + help.twodigits(frames*100 / 30);
+            char buffer[SCREEN_WIDTH_CHARS + 1];
+            timestringcenti(buffer, sizeof(buffer));
+            savetime = buffer;
             break;
+        }
         case 3503:
         {
             state++;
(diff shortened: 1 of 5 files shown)


flibitijibibo (approved):
Code changes make sense to me.

InfoTeddy:
They look fine to me too, I'll merge if testing doesn't reveal anything bad...

flibitijibibo:
Oh also, the scratch buffer idea makes sense too. It would definitely help avoid stack overflows which is probably less scary than the 2.2 days but is still worth being careful about.

InfoTeddy:
Yep, I can definitely agree on that too.

PR #869, opened 2022-03-11
Port to FAudio

N00byKing:
## Changes:

I was browsing the repo and saw #829, with the note that porting to FAudio should be "really easy" for someone else to do.
Answer is no it wasn't (tbf I never touched any audio stuff ever), but I tried nonetheless.

This should be functional, but buggy. I'm opening the PR to see if I should continue or just stop.
~~Thanks for looking at my awful code~~

CI is failing because I used a submodule for FAudio. I can either make CI clone recursive or just copy folder contents, will wait for feedback on that.
Had to increase C Standard Version due to stb_vorbis (90 -> 99). If someone has suggestions on how to avoid that please share.

Most Bugs revolve around tabbing in and out of the game. Sometimes there is no `Resume()` generated for some reason I think. For example, tabbing out during the intro sequence causes the game to stay mute until you do it again during gameplay. I think it might be because of `FAudio_StopEngine()` in `Pause()`, but as that was written in the comment I let it be.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV
- [x] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes

PR #871, opened 2022-03-14
Migrate third-party dependencies to submodules

InfoTeddy:
As directed by Ethan in #869, this PR migrates every third-party dependency (PhysFS, TinyXML-2, LodePNG, and UTF-CPP) to be submodules. They are a bit hairy to work with, but are decidedly better than just copy-pasting the source files directly.

## Legal Stuff:

  for these changes


InfoTeddy:
Ah, okay, so the first roadblock is already here, and it has to do with LodePNG.

Basically, we compile LodePNG as C, not C++. The standard supported way to do this (as specified in its `README`) is to rename `lodepng.cpp` to `lodepng.c`. Otherwise, keeping it named as `.cpp` will make it get compiled with a C++ compiler, which results in C++ linkage, which breaks the build with a link error because we are linking with C.

So, to fix this, I decided to make a commit in the LodePNG submodule renaming `lodepng.cpp` to `lodepng.c`. And this didn't really seem like a good idea, but it worked fine for me locally. It however does not work on other people's computers, because that commit isn't in upstream, they don't have that commit, and there's no way for them to get that commit.

So I'm not really sure what to do to fix this issue.

flibitijibibo:
Can we cheat and have an internal build unit called `VVV_lodepng.c` that just has `#include "lodepng.cpp`?

InfoTeddy:
That sounds like it'd work. Give me a second though, rebasing this is going to be hairy...

flibitijibibo (approved):
Changes lgtm. (Also good grief, nearly 32k lines in dependencies... and that's not even factoring in SDL_mixer.)

As soon as CI is happy this can be merged in so N00by can rebase.

InfoTeddy:
Well that was easy... maybe we don't have an issue with checking out because we're using `v1` of `actions/checkout`.

InfoTeddy:
Also:

> (Also good grief, nearly 32k lines in dependencies... and that's not even factoring in SDL_mixer.)

Just some rough statistics of line count at the moment (using `find ./ | wc -l`):

- **VVVVVV**: If naïvely counting every file in `desktop_version/src/`, 72,000 lines. Although, arguably, `Labclass.cpp`, `Finalclass.cpp`, `Otherlevel.cpp`, `Spacestation2.cpp`, `Tower.cpp`, `WarpClass.cpp`, `Scripts.cpp`, `TerminalScripts.cpp`, and `Credits.h` only contain static data, and shouldn't be counted (in particular, the level files ought to be moved to external files somehow, because embedding them in the source is, to put it bluntly, terrible). When deleting those files, the count drops to 44,000 lines.
- **SDL2**: Just counting `src/` and `include/`, we get 477,000 lines. Although most of that is just support for other platforms and will be compiled out on an actual build. Figuring out which files get ignored or even which lines inside each file get ignored is not a trivial task though, so I won't bother here.
- **SDL2_mixer**: Again, only counting `src/` and `include/`, we get 22,000 lines. Do you think that's a lot? Wait until you see...
- **FAudio**: Only counting `src/` and `include`/, it's 33,000 lines, which I did not expect - I thought it'd be less than SDL2_mixer's. Although, it's possible I'm counting files that I shouldn't be counting here.
- **LodePNG**: We are only using `lodepng.cpp` (compiled as C) and `lodepng.h`, so only counting those. This has 9,000 lines, which isn't too bad (especially compared to the above).
- **TinyXML-2**: Only counting `tinyxml2.cpp` and `tinyxml2.h`, it is 5,000 lines. Actually decent considering it's parsing, you know, XML.
- **PhysFS**: Only counting the `src/` directory, we get 25,000 lines. However, that's a bit unfair, because we don't actually *use* all of those files in VVVVVV (at least if you're bundling dependencies). Counting only the files we use in our `CMakeLists.txt`, along with `physfs.h`, it's only 12,000 lines.
- **UTF-CPP**: We only use `unchecked.h`, and that file is 300 lines. That's actually really good and also pleasantly surprised me too!

flibitijibibo:
The FAudio count sounds about right... XAudio has a bunch of large chunks that we cut out in the static link, like XACT and XAPO builtins. The mixer isn't too long but the SIMD stuff adds a lot of lines in exchange for dramatically improved performance.

PR #874, opened 2022-04-10
Fix music changes between areas

N00byKing:
## Changes:
Missed a Bug in #869.
When changing Areas, [this](https://github.com/TerryCavanagh/VVVVVV/blob/master/desktop_version/src/Music.cpp#L1010) line checks if Music is halted before playing the next track. However, it was never being halted, just paused.

Changing `haltdasmusik` to actually halt `MusicTrack` fixes it, but given the comment placed I wonder if this is the right approach. Could also check if Music is paused at the original location.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -357,6 +357,7 @@ end:
             FAudioSourceVoice_FlushSourceBuffers(musicVoice);
             FAudioVoice_DestroyVoice(musicVoice);
             musicVoice = NULL;
+            paused = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Yeah, so the comment is there because we rely on the music to simply pause so custom levels can resume it later with `resumemusic` and `musicfadein`.

It looks like halting here will destroy the music, so halting here won't work. Instead it's probably better to change `halted` to return being paused.

N00byKing:
So turns out there isn't a way to check if a voice is playing or not. Added a var to check if its paused

InfoTeddy (approved):
Tested and it seems to work fine. Music area changes now work again and `resumemusic`/`musicfadein` work as intended.

PR #882, opened 2022-06-09
Fix semi random volume jumps

### LINKED ISSUE #880: Mute/low volume doesn't always affect first fragments of sounds
This seems to be an FAudio regression, since it doesn't happen on a8feba029f157a2f009fe712b8cf988cd6571bf7 but it does on f877eb3b562ca10b433c6cee94e84a9116da7c95, as well as `master`.

When all sound is muted with M, or when the volume is set low, you may notice small fragments of sound effects still being played at regular volume. The effect is inconsistent (for example, it happened only once with the menu squeak in the first video).

(Make sure to unmute both videos)

https://user-images.githubusercontent.com/44736680/168695640-35dc96a5-4d3e-4bd7-bfdf-e1bc64fac325.mp4

https://user-images.githubusercontent.com/44736680/168696274-b2e3d733-906c-47ce-aef6-b1b6a2782563.mp4

N00byKing:
## Changes:

Fix #880, hopefully.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -165,6 +165,7 @@ end:
                     vlog_error("Unable to queue sound buffer");
                     return;
                 }
+                FAudioVoice_SetVolume(voices[i], volume, FAUDIO_COMMIT_NOW);
                 if (FAudioSourceVoice_Start(voices[i], 0, FAUDIO_COMMIT_NOW))
                 {
                     vlog_error("Unable to start voice processing");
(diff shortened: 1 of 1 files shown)


InfoTeddy:
If this seems good to Ethan and Dav then it can be merged.

desktop_version/src/Music.cpp
```diff
@@ -165,10 +165,12 @@ class SoundTrack
                     vlog_error("Unable to queue sound buffer");
                     return;
                 }
-                if (FAudioSourceVoice_Start(voices[i], 0, FAUDIO_COMMIT_NOW))
+                if (FAudioSourceVoice_Start(voices[i], 0, 144))
                 {
                     vlog_error("Unable to start voice processing");
                 }
+                FAudioVoice_SetVolume(voices[i], volume, 144);
```
flibitijibibo, line 172: Anything blocking this from being above the Start()?

N00byKing: not that i know of. Want me to move it?

flibitijibibo: Definitely - if that doesn't break anything it eliminates the need for operation sets entirely.

N00byKing: Alright, will move it up and see if Dav gets the issue

Dav999-v: No problems so far!

PR #896, opened 2022-10-31
Region system

mothbeanie:
## Changes:

This PR adds a map region system. Regions allow the minimap to display only part of a level even if the level has larger bounds (for example, a 5x5 region in a 20x16 level). Regions can also have custom images. If region`id`.png is present in /graphics, then the minimap will display that image when the current region is `id`. The default region is `0`, and in region `0` the map will display minimap.png (instead of region0.png) if it is present, in order to maintain forwards compatibility. There are a maximum of 401 regions, from `0` to `400`.
It adds three commands to manage them:
`setregion(id,x,y,x2,y2)` - sets region `id` from `x,y` to `x2,y2`.
`changeregion(id)` - sets the current region to `id`. If region `id` doesn't exist, then it will act as if you aren't in any region at all.
`removeregion(id)` - removes region `id` from the array.

Here is a demonstration video:

https://github.com/TerryCavanagh/VVVVVV/assets/19874772/64f2c254-51bd-4b19-9cf0-7ba828d46a55

This PR will not break forwards compatibility. Playing a level which uses this system with an older version of the game probably won't even affect their ability to navigate, although they might end up seeing rooms on the minimap that the level creator intended to be outside of the bounds of the displayed map.

Regions are saved to and loaded from a level's save file, as well as what the current region is. The map cursor and icons for trinkets and teleporters will display properly - if the player/a trinket/a teleporter is outside the bounds of the current region, it won't render on the map screen.

This also cleans up a lot of map code, moving more things into `MapRenderData`. Thanks to Ally and Dav999 for help with this!!!

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -5921,7 +5921,6 @@ void Game::customloadquick(const std::string& savfile)
         {
             map.currentregion = help.Int(pText);
         }
-#if !defined(NO_CUSTOM_LEVELS)
         else if (SDL_strcmp(pKey, "regions") == 0)
         {
             tinyxml2::XMLElement* pElem2;
@@ -5960,7 +5959,6 @@ void Game::customloadquick(const std::string& savfile)
                 map.setregion(thisid, thisrx, thisry, thisrx2, thisry2);
             }
         }
-#endif
     }
 }
 
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -228,16 +228,6 @@ void Game::init(void)
 
     customcol=0;
 
-    map.currentregion = 0;
-    for (size_t i = 0; i < SDL_arraysize(map.region); i++)
-    {
-        map.region[i].isvalid = false;
-        map.region[i].rx = 0;
-        map.region[i].ry = 0;
-        map.region[i].rx2 = 0;
-        map.region[i].ry2 = 0;
-    }
-
     SDL_memset(crewstats, false, sizeof(crewstats));
     SDL_memset(ndmresultcrewstats, false, sizeof(ndmresultcrewstats));
     SDL_memset(besttimes, -1, sizeof(besttimes));
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -228,6 +228,16 @@ void Game::init(void)
 
     customcol=0;
 
+    map.currentregion = 0;
+    for (size_t i = 0; i < SDL_arraysize(map.region); i++)
+    {
+        map.region[i].isvalid = false;
+        map.region[i].rx = 0;
+        map.region[i].ry = 0;
+        map.region[i].rx2 = 0;
+        map.region[i].ry2 = 0;
+    }
+
     SDL_memset(crewstats, false, sizeof(crewstats));
     SDL_memset(ndmresultcrewstats, false, sizeof(ndmresultcrewstats));
     SDL_memset(besttimes, -1, sizeof(besttimes));
@@ -5907,6 +5917,50 @@ void Game::customloadquick(const std::string& savfile)
             map.roomnameset = true;
             map.roomname_special = true;
         }
+        else if (SDL_strcmp(pKey, "currentregion") == 0)
+        {
+            map.currentregion = help.Int(pText);
+        }
+#if !defined(NO_CUSTOM_LEVELS)
+        else if (SDL_strcmp(pKey, "regions") == 0)
+        {
+            tinyxml2::XMLElement* pElem2;
+            for (pElem2 = pElem->FirstChildElement(); pElem2 != NULL; pElem2 = pElem2->NextSiblingElement())
+            {
+                int thisid = 0;
+                int thisrx = 0;
+                int thisry = 0;
+                int thisrx2 = (cl.mapwidth - 1);
+                int thisry2 = (cl.mapheight - 1);
+                if (pElem2->Attribute("id"))
+                {
+                    thisid = help.Int(pElem2->Attribute("id"));
+                }
+
+                for (tinyxml2::XMLElement* pElem3 = pElem2->FirstChildElement(); pElem3 != NULL; pElem3 = pElem3->NextSiblingElement())
+                {
+                    if (SDL_strcmp(pElem3->Value(), "rx") == 0 && pElem3->GetText() != NULL)
+                    {
+                        thisrx = help.Int(pElem3->GetText());
+                    }
+                    if (SDL_strcmp(pElem3->Value(), "ry") == 0 && pElem3->GetText() != NULL)
+                    {
+                        thisry = help.Int(pElem3->GetText());
+                    }
+                    if (SDL_strcmp(pElem3->Value(), "rx2") == 0 && pElem3->GetText() != NULL)
+                    {
+                        thisrx2 = help.Int(pElem3->GetText());
+                    }
+                    if (SDL_strcmp(pElem3->Value(), "ry2") == 0 && pElem3->GetText() != NULL)
+                    {
+                        thisry2 = help.Int(pElem3->GetText());
+                    }
+                }
+
+                map.setregion(thisid, thisrx, thisry, thisrx2, thisry2);
+            }
+        }
+#endif
     }
 }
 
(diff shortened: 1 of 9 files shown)


mothbeanie:
This has been updated to work with the current map code!

mothbeanie:
This has been updated to work with the current map code. Again.

desktop_version/src/Game.cpp
```diff
@@ -5907,6 +5917,50 @@ void Game::customloadquick(const std::string& savfile)
             map.roomnameset = true;
             map.roomname_special = true;
         }
+        else if (SDL_strcmp(pKey, "currentregion") == 0)
+        {
+            map.currentregion = help.Int(pText);
+        }
+#if !defined(NO_CUSTOM_LEVELS)
```
InfoTeddy, line 5924: You do not need a check for `NO_CUSTOM_LEVELS`:
1. This is already in the quicksave load function that is specifically for custom levels (`customloadquick`).
2. We actually removed `NO_CUSTOM_LEVELS` and `NO_EDITOR` builds a while ago. (To be fair, when you originally wrote this we still had them.)
```suggestion

```

desktop_version/src/Game.cpp
```diff
@@ -5907,6 +5917,50 @@ void Game::customloadquick(const std::string& savfile)
...
+                    if (SDL_strcmp(pElem3->Value(), "ry2") == 0 && pElem3->GetText() != NULL)
+                    {
+                        thisry2 = help.Int(pElem3->GetText());
+                    }
+                }
+
+                map.setregion(thisid, thisrx, thisry, thisrx2, thisry2);
+            }
+        }
+#endif
```
InfoTeddy, line 5963:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+        data.legendyoff += 1;
+        break;
+    }
+
+    return data;
+}
+
+void mapclass::setregion(int id, int rx, int ry, int rx2, int ry2)
+{
+#if !defined(NO_CUSTOM_LEVELS)
```
InfoTeddy, line 2294:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+    if (INBOUNDS_ARR(id, region))
+    {
+        region[id].isvalid = true;
+        region[id].rx = SDL_clamp(rx, 0, cl.mapwidth - 1);
+        region[id].ry = SDL_clamp(ry, 0, cl.mapheight - 1);
+        region[id].rx2 = SDL_clamp(rx2, 0, cl.mapwidth - 1);
+        region[id].ry2 = SDL_clamp(ry2, 0, cl.mapheight - 1);
+    }
+#endif
```
InfoTeddy, line 2303:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+        region[id].ry = 0;
+        region[id].rx2 = 0;
+        region[id].ry2 = 0;
+    }
+#endif
+}
+
+void mapclass::changeregion(int id)
+{
+#if !defined(NO_CUSTOM_LEVELS)
```
InfoTeddy, line 2322:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+    if (INBOUNDS_ARR(id, region))
+    {
+        currentregion = id;
+        cl.generatecustomminimap();
+    }
+#endif
```
InfoTeddy, line 2328:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+}
```
InfoTeddy, line 2329: Please end the file with a newline.

desktop_version/src/Game.cpp
```diff
@@ -228,6 +228,16 @@ void Game::init(void)
...
+    map.currentregion = 0;
+    for (size_t i = 0; i < SDL_arraysize(map.region); i++)
+    {
+        map.region[i].isvalid = false;
+        map.region[i].rx = 0;
+        map.region[i].ry = 0;
+        map.region[i].rx2 = 0;
+        map.region[i].ry2 = 0;
+    }
+
```
InfoTeddy, line 240: 1. Why are you initializing `mapclass` attributes inside `Game`?
2. Instead of using a `for`-loop to initialize `region`, I would much rather you write `SDL_zeroa(region);`. It's concise and covers cases in the future when we may add more attributes to `regionstruct`.
3. You need to also zero `map.currentregion` and `map.region` in `scriptclass::hardreset`, otherwise they will persist after loading a new level.

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+        region[id].isvalid = false;
+        region[id].rx = 0;
+        region[id].ry = 0;
+        region[id].rx2 = 0;
+        region[id].ry2 = 0;
+    }
+#endif
+}
+
+void mapclass::changeregion(int id)
```
InfoTeddy, line 2320: When the region being removed is the current region, `cl.generatecustomminimap();` is _not_ called again, so the minimap image ends up being wrong.

InfoTeddy (changes requested):
I was expecting this to be a quick review. Unfortunately, it wasn't. :stuck_out_tongue:

I haven't even fully tested everything yet.

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2213,117 @@ void mapclass::twoframedelayfix(void)
...
+    default:
+        data.legendxoff += 3;
+        data.legendyoff += 1;
+        break;
+    }
+
+    return data;
+}
+
+void mapclass::setregion(int id, int rx, int ry, int rx2, int ry2)
```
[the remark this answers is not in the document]
mothbeanie: Back when I first wrote the pr I had this idea that region 0 could be editable - I don't think that's a great idea now, and while it was intentional and not an oversight at the time, I've now made it so you can no longer use `setregion` or `removeregion` on region 0. Additionally, the game will no longer load `region0.png` if it exists and will always use `minimap.png` for region 0 (as it is the default), to prevent there being multiple files which can potentially be used for the same purpose

mothbeanie:
Applied changes from the review!

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2216,117 @@ void mapclass::twoframedelayfix(void)
...
```
InfoTeddy, line 2295: It matters which order you input the coordinates, so if you `setregion` coordinates backwards (e.g. 5,5,1,1, which is (5,5) to (1,1)), then you can end up with funky-looking minimaps like this.

A blank minimap with missing lines and inverted corners

You should either disallow this or make it work by reversing the order.

InfoTeddy (changes requested):

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2216,131 @@ void mapclass::twoframedelayfix(void)
...
+{
+    if (INBOUNDS_ARR(id, region) && id > 0)
+    {
+        // swap the variables if they're entered in the wrong order
+        if (rx2 > rx)
```
InfoTeddy, line 2300: This is the wrong way around and not only does it not fix wrongly-ordered inputs, it actually breaks all correctly-ordered inputs by incorrectly swapping them around.

```suggestion
        if (rx2 < rx)
```

desktop_version/src/Map.cpp
```diff
@@ -2215,3 +2216,131 @@ void mapclass::twoframedelayfix(void)
...
+        {
+            int temp = rx;
+            rx = rx2;
+            rx2 = temp;
+        }
+        if (ry2 > ry)
```
InfoTeddy, line 2306:
```suggestion
        if (ry2 < ry)
```

InfoTeddy (changes requested):

mothbeanie:
HAHAHAHA OMG OKAY I'LL FIX THAT

PR #897, opened 2022-11-03
Teleporter additions

AllyTally:
## Changes:

In an effort to reduce the split between custom levels and the main game, I've made a system where when exiting a teleporter, the player follows a three point Bézier curve. Point A (the starting point) is the teleporter's center, and point B and C are saved in the teleporter's data, along with facing direction, gravity state and the amount of frames that following the curve should take.

Along with adding a teleporter edentity to house this addition, this PR also runs `setteleporter()` dynamically when loading edentities, much like trinkets.

However, this PR does NOT replace the current system, and is instead locked to editor-only at the moment, for the following reasons:

- The main game does not use the edentity system (this is possible to change, but...)
- There's certain glitches that speedrunners/TASes use, and I don't know much about glitchrunner mode, so I think someone else should do that part.

## Legal Stuff:

  for these changes


AllyTally:
I've pushed a new commit which adds p5 and p6 to the `createentity` scripting command, since they were added to the internal function.

desktop_version/src/Entity.cpp
```diff
@@ -1,6 +1,7 @@
 #define OBJ_DEFINITION
 #include "Entity.h"
 
+#include <cmath>
```
InfoTeddy, line 4: Why `#include` this? I don't see any functions from it being used here. (And this is a moot point, but if you were to `#include` it, please do `#include <math.h>` instead of `#include <cmath>`.)

desktop_version/src/Map.cpp
```diff
@@ -1885,6 +1912,11 @@ void mapclass::loadlevel(int rx, int ry)
             case 13: // Warp Tokens
                 obj.createentity(ex, ey, 13, ent.p1, ent.p2);
                 break;
+            case 14: // Teleporter
+            {
```
InfoTeddy, line 1916:
```suggestion

```

desktop_version/src/Map.cpp
```diff
@@ -1885,6 +1912,11 @@ void mapclass::loadlevel(int rx, int ry)
...
+                obj.createentity(ex, ey, 14, 0, ((rx + (ry * 100)) * 20) + tempcheckpoints, ent.p1, ent.p2, ent.p3, ent.p4, ent.p5, ent.p6);
+                break;
+            }
```
InfoTeddy, line 1919:
```suggestion

```

third_party/c-hashmap
InfoTeddy, line 1: Huh?

InfoTeddy (changes requested):
Style pass.

desktop_version/src/Map.cpp
```diff
@@ -412,19 +414,44 @@ void mapclass::roomnamechange(const int x, const int y, const char** lines, cons
...
+    }
+    return a.x < b.x;
+}
+
 void mapclass::initcustommapdata(void)
 {
     shinytrinkets.clear();
+    teleporters.clear();
+
+    std::vector<SDL_Point> teleporters;
```
InfoTeddy, line 431: This code works but you should not be shadowing the class member attribute, it impacts readability. Please rename this to `customteleporters`.

InfoTeddy (changes requested):
Another style pass.

PR #898, opened 2022-11-07
Deduplicate map rendering code

AllyTally:
## Changes:

This PR deduplicates a lot of map rendering code, as it's cloned a few times for the following reasons:
- Main game map
- Teleporter screen map
- Custom level map

This splits a lot of map rendering features into smaller functions, instead of a bunch of slightly different copy-pasted huge functions. This also allows the teleporter screen to work properly in custom levels with smaller map sizes, and with showing the legend (trinkets, teleporters, targets) in those as well.

This also cleans up some bad code such as fog rendering, so it uses a small for loop instead of 30+ lines of code...

**NOTE:** One of these changes means the teleporter screen in custom levels no longer always shows `minimap.png` -- it shows the map that custom levels are supposed to show. This is a _fix_, but I thought I'd mention it.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Render.cpp
+++ b/desktop_version/src/Render.cpp
@@ -24,6 +24,16 @@ static int tr;
 static int tg;
 static int tb;
 
+struct MapRenderData {
+    int width;
+    int height;
+    int zoom;
+    int xoff;
+    int yoff;
+    int legendxoff;
+    int legendyoff;
+};
+
 static inline void drawslowdowntext(void)
 {
     switch (game.slowdown)
@@ -2040,6 +2050,35 @@ static void draw_roomname_menu(void)
  * the same in Flip Mode. */
 #define FLIP(y, h) (graphics.flipmode ? 220 - (y) - (h) : (y))
 
+static MapRenderData getmaprenderdata()
+{
+    MapRenderData data;
+
+    data.width = map.custommode ? map.customwidth : 20;
+    data.height = map.custommode ? map.customheight : 20;
+    data.zoom = map.custommode ? map.customzoom : 1;
+    data.xoff = map.custommode ? map.custommmxoff : 0;
+    data.yoff = map.custommode ? map.custommmyoff : 0;
+
+    switch (data.zoom)
+    {
+    case 4:
+        data.legendxoff = 60;
+        data.legendyoff = 35;
+        break;
+    case 2:
+        data.legendxoff = 48;
+        data.legendyoff = 26;
+        break;
+    default:
+        data.legendxoff = 43;
+        data.legendyoff = 22;
+        break;
+    }
+
+    return data;
+}
+
 static void rendermap(void)
 {
 #ifndef NO_CUSTOM_LEVELS
@@ -2057,24 +2096,20 @@ static void rendermap(void)
 
 static void rendermapfog(void)
 {
-    int mapwidth = map.custommode ? map.customwidth : 20;
-    int mapheight = map.custommode ? map.customheight : 20;
-    int mapzoom = map.custommode ? map.customzoom : 1;
-    int mapxoff = map.custommode ? map.custommmxoff : 0;
-    int mapyoff = map.custommode ? map.custommmyoff : 0;
+    MapRenderData data = getmaprenderdata();
 
-    for (int j = 0; j < mapheight; j++)
+    for (int j = 0; j < data.height; j++)
     {
-        for (int i = 0; i < mapwidth; i++)
+        for (int i = 0; i < data.width; i++)
         {
             if (!map.isexplored(i, j))
             {
                 // Draw the fog, depending on the custom zoom size
-                for (int x = 0; x < mapzoom; x++)
+                for (int x = 0; x < data.zoom; x++)
                 {
-                    for (int y = 0; y < mapzoom; y++)
+                    for (int y = 0; y < data.zoom; y++)
                     {
-                        graphics.drawimage(2, mapxoff + 40 + (x * 12) + (i * (12 * mapzoom)), mapyoff + 21 + (y * 9) + (j * (9 * mapzoom)), false);
+                        graphics.drawimage(2, data.xoff + 40 + (x * 12) + (i * (12 * data.zoom)), data.yoff + 21 + (y * 9) + (j * (9 * data.zoom)), false);
                     }
                 }
             }
@@ -2086,25 +2121,7 @@ static void rendermaplegend(void)
 {
     // Draw the map legend, aka teleports/targets/trinkets
 
-    int zoom_offset_x;
-    int zoom_offset_y;
-    int zoom_mult = map.custommode ? map.customzoom : 1;
-
-    switch (zoom_mult)
-    {
-    case 4:
-        zoom_offset_x = 60;
-        zoom_offset_y = 35;
-        break;
-    case 2:
-        zoom_offset_x = 48;
-        zoom_offset_y = 26;
-        break;
-    default:
-        zoom_offset_x = 43;
-        zoom_offset_y = 22;
-        break;
-    }
+    MapRenderData data = getmaprenderdata();
 
     const int tile_offset = graphics.flipmode ? 3 : 0;
 
@@ -2112,11 +2129,11 @@ static void rendermaplegend(void)
     {
         if (map.showteleporters && map.isexplored(map.teleporters[i].x, map.teleporters[i].y))
         {
-            graphics.drawtile(zoom_offset_x + (map.teleporters[i].x * 12 * zoom_mult), zoom_offset_y + (map.teleporters[i].y * 9 * zoom_mult), 1127 + tile_offset);
+            graphics.drawtile(data.legendxoff + (map.teleporters[i].x * 12 * data.zoom), data.legendyoff + (map.teleporters[i].y * 9 * data.zoom), 1127 + tile_offset);
         }
         else if (map.showtargets && !map.isexplored(map.teleporters[i].x, map.teleporters[i].y))
         {
-            graphics.drawtile(zoom_offset_x + (map.teleporters[i].x * 12 * zoom_mult), zoom_offset_y + (map.teleporters[i].y * 9 * zoom_mult), 1126 + tile_offset);
+            graphics.drawtile(data.legendxoff + (map.teleporters[i].x * 12 * data.zoom), data.legendyoff + (map.teleporters[i].y * 9 * data.zoom), 1126 + tile_offset);
         }
     }
 
@@ -2126,7 +2143,7 @@ static void rendermaplegend(void)
         {
             if (!obj.collect[i])
             {
-                graphics.drawtile(zoom_offset_x + (map.shinytrinkets[i].x * 12 * zoom_mult), zoom_offset_y + (map.shinytrinkets[i].y * 9 * zoom_mult), 1086 + tile_offset);
+                graphics.drawtile(data.legendxoff + (map.shinytrinkets[i].x * 12 * data.zoom), data.legendyoff + (map.shinytrinkets[i].y * 9 * data.zoom), 1086 + tile_offset);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+void rendermapfog(void)
+{
+    int mapwidth = map.custommode ? map.customwidth : 20;
+    int mapheight = map.custommode ? map.customheight : 20;
+    int mapzoom = map.custommode ? map.customzoom : 1;
+    int mapxoff = map.custommode ? map.custommmxoff : 0;
+    int mapyoff = map.custommode ? map.custommmyoff : 0;
+
+    // Draw the fog of war
+    for (int j = 0; j < mapheight; j++)
```
InfoTeddy, line 2069:
```suggestion
    for (int j = 0; j < mapheight; ++j)
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+    {
+        for (int i = 0; i < mapwidth; i++)
```
InfoTeddy, line 2071:
```suggestion
        for (int i = 0; i < mapwidth; ++i)
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+            zoom_offset_y = 35;
+            break;
+        case 2:
+            zoom_offset_x = 48;
+            zoom_offset_y = 26;
+            break;
+        default:
+            zoom_offset_x = 43;
+            zoom_offset_y = 22;
+            break;
```
InfoTeddy, line 2109:
```suggestion
    case 4:
        zoom_offset_x = 60;
        zoom_offset_y = 35;
        break;
    case 2:
        zoom_offset_x = 48;
        zoom_offset_y = 26;
        break;
    default:
        zoom_offset_x = 43;
        zoom_offset_y = 22;
        break;
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+    }
+
+    int tile_offset = graphics.flipmode ? 3 : 0;
```
InfoTeddy, line 2112:
```suggestion
    const int tile_offset = graphics.flipmode ? 3 : 0;
```

desktop_version/src/Render.cpp
```diff
@@ -2769,48 +2688,40 @@ void teleporterrender(void)
...
+    default:
+        zoom_offset_x = 43;
+        zoom_offset_y = 22;
+        break;
     }
 
     tempx = map.teleporters[game.teleport_to_teleporter].x;
     tempy = map.teleporters[game.teleport_to_teleporter].y;
-    if (game.useteleporter && ((help.slowsine%16)>8))
+    if (game.useteleporter && ((help.slowsine % 16) > 8))
```
InfoTeddy, line 2722:
```suggestion
    if (game.useteleporter && help.slowsine % 16 > 8)
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
  * the same in Flip Mode. */
 #define FLIP(y, h) (graphics.flipmode ? 220 - (y) - (h) : (y))
 
+void rendermap(void)
+{
+#ifndef NO_CUSTOM_LEVELS
+    if (map.custommode)
+    {
+        //draw the map image
```
InfoTeddy, line 2048:
```suggestion
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+        graphics.drawpixeltextbox(35 + map.custommmxoff, 16 + map.custommmyoff, map.custommmxsize + 10, map.custommmysize + 10, 65, 185, 207);
+        graphics.drawpartimage(graphics.minimap_mounted ? 1 : 12, 40 + map.custommmxoff, 21 + map.custommmyoff, map.custommmxsize, map.custommmysize);
+        return;
+     }
+#endif /* NO_CUSTOM_LEVELS */
+
+    //draw the map image
```
InfoTeddy, line 2055:
```suggestion
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+
+void rendermapfog(void)
+{
+    int mapwidth = map.custommode ? map.customwidth : 20;
+    int mapheight = map.custommode ? map.customheight : 20;
+    int mapzoom = map.custommode ? map.customzoom : 1;
+    int mapxoff = map.custommode ? map.custommmxoff : 0;
+    int mapyoff = map.custommode ? map.custommmyoff : 0;
+
+    // Draw the fog of war
```
InfoTeddy, line 2068: The comment is not needed when this routine is in a separate function and it's in the name of the function.
```suggestion
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2040,143 @@ static void draw_roomname_menu(void)
...
+            graphics.drawrect(40 + ((game.roomx - 100) * 12) + 2, 21 + 2, 12 - 4, 180 - 4, 16, 245 - (help.glow * 2), 245 - (help.glow * 2));
+        }
+        else if (map.cursorstate == 1)
+        {
+            if (int(map.cursordelay / 4) % 2 == 0)
+            {
+                graphics.drawrect(40 + ((game.roomx - 100) * 12), 21, 12, 180, 255, 255, 255);
+                graphics.drawrect(40 + ((game.roomx - 100) * 12) + 2, 21 + 2, 12 - 4, 180 - 4, 255, 255, 255);
+            }
+            if (map.cursordelay > 30) map.cursorstate = 2;
```
InfoTeddy, line 2160: I didn't realize the Tower code still mutated `map.cursorstate` in a delta function, this should be removed.
```suggestion
```

desktop_version/src/Script.cpp
```diff
@@ -2857,6 +2857,10 @@ void scriptclass::startgamemode( int t )
         map.resetplayer();
         map.gotoroom(game.saverx, game.savery);
         map.initmapdata();
+
+        cl.generatecustomminimap();
+        map.customshowmm = true;
+
```
InfoTeddy, line 2863: `map.customshowmm` is reset to `true` in `scriptclass::hardreset` anyways, so this instance is unnecessary (and all the other instances of `map.customshowmm = true;` should probably have been removed in the first place).

The bigger problem with this, though, is that you're doing this in case 20, which is only when the level editor is entered. At that point, a level has not yet been loaded, so the generated minimap will always be completely blank. The `cl.generatecustomminimap();` should be moved to case 21, which is when in-editor playtesting is entered.

InfoTeddy (changes requested):
Apologies for the delay. Here's a style pass, along with one issue I found from doing a little bit of testing.

Also, I'm not a huge fan of the several copy-pasted variables of `mapwidth`, `mapheight`, `mapzoom` (which is named `zoom_mult` at one point), `mapxoff`, `mapyoff`, `zoom_offset_x` and `zoom_offset_y`. The best solution for this would probably be to write a helper function that computes these variables and returns a `struct` containing them.

desktop_version/src/Render.cpp
Dav999-v: What's wrong with using `i++`, and right with using `++i`, in a for loop? I'm much more used to seeing `i++` in general, and it's not good for consistency either because `i++` is also much more common in VVVVVV itself:

```
desktop_version/src$ egrep -r '; *[A-Za-z0-9_]+ *\+\+ *)' | wc -l
318
desktop_version/src$ egrep -r '; *\+\+ *[A-Za-z0-9_]+ *)' | wc -l
41
```

InfoTeddy: I prefer the prefix operator because it definitively guarantees there's no temporary generated to hold the previous value as opposed to the postfix operator. Although, it probably doesn't matter much because compilers _should_ figure out that the previous value isn't used in these cases. I didn't consider that the postfix was in dominant usage.

AllyTally:
Everything raised should be resolved now.

desktop_version/src/Render.cpp
```diff
@@ -24,6 +24,16 @@ static int tr;
 static int tg;
 static int tb;
 
+struct MapRenderData {
```
InfoTeddy, line 27:
```suggestion
struct MapRenderData
{
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2050,142 @@ static void draw_roomname_menu(void)
...
+     }
+#endif /* NO_CUSTOM_LEVELS */
+
+    graphics.drawpixeltextbox(35, 16, 250, 190, 65, 185, 207);
+    graphics.drawimage(1, 40, 21, false);
+}
+
+static void rendermapfog(void)
+{
+    MapRenderData data = getmaprenderdata();
```
InfoTeddy, line 2099:
```suggestion
    const MapRenderData data = getmaprenderdata();
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2050,142 @@ static void draw_roomname_menu(void)
...
+            }
+        }
+    }
+}
+
+static void rendermaplegend(void)
+{
+    // Draw the map legend, aka teleports/targets/trinkets
+
+    MapRenderData data = getmaprenderdata();
```
InfoTeddy, line 2124:
```suggestion
    const MapRenderData data = getmaprenderdata();
```

desktop_version/src/Render.cpp
```diff
@@ -2040,6 +2050,142 @@ static void draw_roomname_menu(void)
...
+            {
+                graphics.drawtile(data.legendxoff + (map.shinytrinkets[i].x * 12 * data.zoom), data.legendyoff + (map.shinytrinkets[i].y * 9 * data.zoom), 1086 + tile_offset);
+            }
+        }
+    }
+}
+
+static void rendermapcursor(const bool flashing)
+{
+    MapRenderData data = getmaprenderdata();
```
InfoTeddy, line 2154:
```suggestion
    const MapRenderData data = getmaprenderdata();
```

desktop_version/src/Render.cpp
```diff
@@ -2736,31 +2680,13 @@ void teleporterrender(void)
...
-        //tower!instead of room y, scale map.ypos
-        graphics.drawrect(40 + ((game.roomx - 100) * 12) + 2, 21  + 2, 12 - 4, 180 - 4, 16, 245 - (help.glow * 2), 245 - (help.glow * 2));
-    }
-    else
-    {
-        graphics.drawrect(40 + ((game.roomx - 100) * 12) + 2, 21 + ((game.roomy - 100) * 9) + 2, 12 - 4, 9 - 4, 16, 245 - (help.glow * 2), 245 - (help.glow * 2));
-    }
+    // Draw a box around the currently selected teleporter
+
+    MapRenderData data = getmaprenderdata();
```
InfoTeddy, line 2689:
```suggestion
    const MapRenderData data = getmaprenderdata();
```

desktop_version/src/Script.cpp
```diff
@@ -2888,10 +2888,18 @@ void scriptclass::startgamemode( int t )
         map.resetplayer();
         map.gotoroom(game.saverx, game.savery);
         map.initmapdata();
-        if(cl.levmusic>0){
+
```
InfoTeddy, line 2891:
```suggestion
```

InfoTeddy (changes requested):
Another style pass.

Also, I found a major issue. In maps that aren't square (e.g. 5x10), the legend is drawn with the wrong offset.

A minimap with shiny trinket legends drawn in the wrong place

This is the test level I used for the above screenshot:
* [5x10_minimap_test.zip](https://github.com/TerryCavanagh/VVVVVV/files/10125840/5x10_minimap_test.zip)

InfoTeddy (approved):
Thanks!

PR #899, opened 2022-11-07
Gravity Token additions

AllyTally:
## Changes:

Gravity tokens are interesting. They were added in the mechanics test, but were never used after. You can assume they became gravity lines, but they're still pretty interesting.

This PR freshens them up a little -- a better default sprite, a sound when you touch them, and making them respawn on death.

Additionally, a few createentity arguments have been added:

- `meta1` is a sprite ID offset
- `meta2` is the colour
- `p1` is if the entity animates or not (0 is doesn't animate)
- `p2` is a sound ID offset

Of course, these additions shouldn't affect gameplay, other than respawning on death which is just convenience (you can just leave and reenter the room to respawn them otherwise), so these changes are solely visual and auditory. Hopefully this breathes some life into the unused entity.

~~(Who knows, maybe we could even get an edentity later when more of those are added due to the main-game-as-custom-level goal...?)~~ I have now added an edentity for consistency, however this means this PR now affects gameplay.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -671,6 +671,10 @@ static void draw_entities(void)
                 font::print(PR_FONT_8X8, x, y, "////", 255 - help.glow, 255 - help.glow, 255 - help.glow);
                 graphics.draw_rect(x, y, 32, 8, graphics.getRGB(255, 255, 255));
                 break;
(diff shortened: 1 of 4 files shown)


AllyTally:
I've like to mention that I decided to add a flip token edentity for consistency with other open PRs.

Issue #900, opened 2022-11-07
Non-WAV sound effects no longer play


AllyTally:
With SDL_mixer, if you replaced a sound effect, it'd play as long as the audio format was supported, even though the filename clearly ended in `.wav`, like `cry.wav`. Level creators used this quirk to have OGG sound effects instead to save space. However, with the move to FAudio, this is no longer the case, with non-WAV sounds simply not playing instead.

PR #902, opened 2022-11-07
Add custom file playback to playef()

Fussmatte:
## Changes:

This PR allows users to call custom filenames via the `playef()` command. For example, `playef(mysfx)` will play back either `sounds/mysfx.ogg` or `sounds/mysfx.wav`. Custom sounds are loaded with the level, and (for now) only the level's own `sounds/` folder is read from.

This uses a new function called `FILESYSTEM_enumerateAssets()` which gets all the files from a specific folder of a custom level's assets folder. This could allow other possibilities for loading custom files (such as graphics) in the future.

Big thanks to @Dav999-v for helping me write this!

## Legal Stuff:

  for these changes


Fussmatte:
Worth noting: if you run this version right now, `.ogg` files cannot be loaded this way. This is due to #900, so once that's fixed it should work with no issues.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -865,6 +867,33 @@ void FILESYSTEM_enumerateLevelDirFileNames(
     }
 }
 
+std::vector<std::string> FILESYSTEM_enumerateAssets(const char* folder)
+{
```
Dav999-v, line 871: Maybe this could use a comment at the top of the function saying what it does, because it _only_ enumerates level-specific assets and that might be otherwise misleading.

desktop_version/src/Music.cpp
```diff
@@ -1130,6 +1154,21 @@ void musicclass::playef(int t)
...
+    {
+        size_t lastindex = soundTracks[i].name.find_last_of('.');
+        std::string rawname = soundTracks[i].name.substr(0, lastindex);
+        if (t == rawname)
+        {
+            soundTracks[i].Play();
+            return;
+        }
+    }
+    return;
```
Dav999-v, line 1169: This return doesn't technically hurt but it's redundant, the function ends here either way.

desktop_version/src/Script.cpp
```diff
@@ -328,7 +328,12 @@ void scriptclass::run(void)
             }
             if (words[0] == "playef")
             {
-                music.playef(ss_toi(words[1]));
+                if (words[1] != "0" && ss_toi(words[1]) == 0)
+                {
+                    music.playef_name(raw_words[1]);
+                } else {
```
Dav999-v, line 334:
```suggestion
                }
                else
                {
```

Dav999-v (changes requested):

desktop_version/src/Music.cpp
```diff
@@ -100,6 +99,7 @@ class SoundTrack
         SDL_RWops *fileIn;
         SDL_zerop(this);
         FILESYSTEM_loadAssetToMemory(fileName, &mem, &length, false);
+        name = std::string(SDL_strrchr(fileName, '/') + 1);
```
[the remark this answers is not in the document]
Fussmatte: So it's good if I put it after `valid = true`?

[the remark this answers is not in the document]
Dav999-v: Yeah, should be!

Dav999-v (approved):

Dav999-v:
I just needed `FILESYSTEM_enumerateAssets()` for #922, so it's now in my PR as well - meaning if #922 gets merged before this one, all changes this makes to FileSystemUtils.[cpp/h] are no longer necessary and need to be undone. Also, I decided to change how it works, so this will also need to be replaced:
```cpp
    std::vector<std::string> customSounds = FILESYSTEM_enumerateAssets("sounds/");
    for (int i = 0; i < customSounds.size(); i++)
    {
        // use customSounds[i]
    }
```
by something like this:
```cpp
    void* handle = NULL;
    const char* item;
    while ((item = FILESYSTEM_enumerateAssets("sounds/", &handle)) != NULL)
    {
        // use item
    }
    FILESYSTEM_freeEnumerate(&handle);
```

InfoTeddy:
Yes.

Fussmatte:
So apparently OGG files don't work here after all, even though OGG playback was fixed. Will have to look into this before it can be merged...

Issue #832, opened 2021-09-01
Port the mobile version enchancements back to the native version


TerryCavanagh:
So, the mobile version has a number of enhancements that I'm very fond of, which aren't in the native version of the game. A list:

![image](https://user-images.githubusercontent.com/2874150/131651450-04910673-4937-486e-adcb-670669e382e4.png)
 - Nice, clear touch buttons in menus for big fingers, including a super big button to start the game

![image](https://user-images.githubusercontent.com/2874150/131651698-93dc78a8-4296-4281-a9cf-147239cbbc9f.png)
 - Little icons for player levels and a significantly nicer player level selection menu (these are hardcoded screenshots for the included levels, I'm not sure what the solution is for player levels in general but I'm open to ideas)

![image](https://user-images.githubusercontent.com/2874150/131651404-3f83d988-ea28-4624-9110-64bc2708291d.png)
 - A tutorial that explains how to use the touch screen controls (which should be skippable ideally)

![image](https://user-images.githubusercontent.com/2874150/131651674-ea9ab57e-41db-40fe-bd91-917f53cc72d2.png)
 - Selectable touch screen control schemes, (swipe and tap as default)

![image](https://user-images.githubusercontent.com/2874150/131651638-e6033774-77dd-4a10-8a14-51dfbfa72cb1.png)
 - This is a big one: a standalone build for Super Gravitron. I just forked the source to make this originally, but it could be a #define...

Eventually, it would be great to have new mobile versions, so re-implementing and updating these features is something I'd love to do! I have no estimate for when I might get around to that, but I figured I could at least write them up for now so that other people could see what I'm thinking.

AllyTally:
I actually recreated the super gravitron using the source (https://github.com/AllyTally/VVVVVV/tree/super-grav), however it hasn't been updated in a while... I might start from scratch next time I decide to work on it. It contains the new "about VVVVVV" menu option with images and all, plus two custom modes other than the normal super gravitron. It also uses a separate save file.

I was going to release it eventually, but started thinking of an editor I could make for it... stopping development rather quickly. Once #834 is made, I might try to port my changes to the latest commit.

leo60228:
For what it's worth, VVVVVV-CE had an Android port, as well as an iOS branch. iOS was only on a branch because SDL_mixer was very broken. iOS build systems are *terrible*, and audio only worked on the headphones (I found multiple reports of this over the years from searching). Hopefully that'll automatically be fixed with FAudio, though?

[...]

PR #907, opened 2022-11-13
Super Gravitron #define

AllyTally:
## Changes:

As requested in #832, this adds a define for Super Gravitron. Super Gravitron is a version of the mobile VVVVVV port, which only contains the Super Gravitron. This PR ports it to the native version. This doesn't do any changes to the options menu however, meaning you can still activate flip mode and stuff, however I don't think it's that big of a deal... (that is to say, I'm tired of working on this PR)

## Legal Stuff:

  for these changes


AllyTally:
Here's the graphics which would need to be added to the data.zip -- [graphics.zip](https://github.com/TerryCavanagh/VVVVVV/files/9998787/graphics.zip)

PR #909, opened 2022-11-21
Editor entity blocks

AllyTally:
## Changes:

Activity zones are pretty hardcoded, so this makes them data-driven instead, using the `edentity` system. This is, once again, laying foundation for the goal of reducing the split between custom levels and the main game.

Activity zones are ID 20, and rendered the same as script boxes. This adds `activitytext` and `activitycolour` properties to editor entities.

EDIT: This PR now has edentities for every block type. 21 are collision boxes, 22 are damage boxes, 23 are directional boxes and 24 are safe boxes.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/CustomLevels.cpp
+++ b/desktop_version/src/CustomLevels.cpp
@@ -1214,6 +1214,15 @@ bool customlevelclass::load(std::string _path)
                 edEntityEl->QueryIntAttribute("p5", &entity.p5);
                 edEntityEl->QueryIntAttribute("p6", &entity.p6);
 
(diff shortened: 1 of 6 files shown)


AllyTally:
I've now added a `createblock` command, since I realized that activity zones should have the ability to be conditional -- and activity zones should be able to spawn themselves. The main game does both of these!

NyakoFox:
This PR has turned into a draft because in its current state, it is pretty flimsy (and `createblock` is far too technical). Will take another look at this some other time.

PR #910, opened 2022-11-21
Custom textbox colors

AllyTally:
## Changes:

Textbox colors are also hardcoded, this loads them from XML, this is a visual change.

```xml
  <TextboxColours>
      <colour r="170" g="47" b="88" name="test"></colour>
  </TextboxColours>
  ```

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/BlockV.cpp
+++ b/desktop_version/src/BlockV.cpp
@@ -2,6 +2,7 @@
 
 #include <SDL_stdinc.h>
 
+#include "CustomLevels.h"
 #include "Font.h"
 
 blockclass::blockclass(void)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: desktop_version/src/BlockV.cpp [hunk 1]
```diff
@@ -44,6 +46,15 @@ void blockclass::rectset(const int xi, const int yi, const int wi, const int hi)
 
 void blockclass::setblockcolour(const char* col)
 {
+#if !defined NO_CUSTOM_LEVELS
```
InfoTeddy, line 49:
```suggestion
#ifndef NO_CUSTOM_LEVELS
```

desktop_version/src/Script.cpp
```diff
@@ -415,6 +415,15 @@ void scriptclass::run(void)
             {
                 //oh boy
                 //first word is the colour.
+#if !defined NO_CUSTOM_LEVELS
```
InfoTeddy, line 418:
```suggestion
#ifndef NO_CUSTOM_LEVELS
```

InfoTeddy (changes requested):
This works in internal but not in simplified. It should probably work in simplified, too.

Also, style pass.

InfoTeddy (approved):
After discussing it on Discord, Ally and I agreed that it doesn't need to work in simplified because `orange`, `white`, and `transparent` are internal-only, and simplified scripting has other considerations for color (like squeak and position). Looks good to me!

PR #916, opened 2022-11-30
Allow 26 textbox lines

AllyTally:
## Changes:

For some reason, textboxes can only have 11 lines added before the line list gets cleared. This changes that behavior, allowing a total of 26 lines before *ignoring* extra lines added. 26 is the maximum amount of lines that the screen can display vertically without the textbox going off-screen. As far as I can tell, this is only visual, and doesn't affect any used glitches.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -520,9 +520,8 @@ void scriptclass::run(void)
                 textx = ss_toi(words[2]);
                 texty = ss_toi(words[3]);
 
-                // If words[4] ends with L, set large to true.
-                textlarge = (words[4].size() > 0 && words[4][words[4].size() - 1] == 'l');
-                int lines = ss_toi(words[4].substr(0, words[4].size() - (textlarge ? 1 : 0)));
+                textlarge = endsWith(words[4].c_str(), "l") || endsWith(words[4].c_str(), "L");
+                int lines = ss_toi(words[4]);
 
                 //Number of lines for the textbox!
                 txt.clear();
@@ -3523,7 +3522,7 @@ bool scriptclass::loadcustom(const std::string& t)
                     add("text(blue,0,0,"+words[1]+")");
                 break;
             }
-            int ti = ss_toi(words[1].c_str());
+            int ti = ss_toi(words[1]);
             int nti = ti >= 0 ? ti : 1;
             for (int ti2 = 0; ti2 < nti; ti2++)
             {
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -48,6 +48,7 @@ scriptclass::scriptclass(void)
     textpad_right = 0;
     textpadtowidth = 0;
     textcase = 1;
+    textlarge = false;
 }
 
 void scriptclass::clearcustom(void)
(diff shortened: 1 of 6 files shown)


InfoTeddy (changes requested):
As discussed in the VVVVVV Discord, some levels depend on the 11-line limit to create a 0-line text box by creating a text box with 12 lines, so I want a way to not break them and force them to update their scripts. The least intrusive way to do this would be to add a postfix `L` to the argument (similar to how in C/C++, postfixing `L` to an `int` literal makes it a `long`) that signals if the text box should be bumped up to this new limit or not.

InfoTeddy (commented):
Once you squash and rebase I'll merge this.

desktop_version/src/Script.cpp
```diff
@@ -519,9 +520,13 @@ void scriptclass::run(void)
                 textx = ss_toi(words[2]);
                 texty = ss_toi(words[3]);
 
+                // If words[4] ends with L, set large to true.
+                textlarge = (words[4].size() > 0 && words[4][words[4].size() - 1] == 'l');
+                int lines = ss_toi(words[4].substr(0, words[4].size() - (textlarge ? 1 : 0)));
```
InfoTeddy, line 525: You don't need to do any complicated `substr` stuff because `ss_toi` will take in every numeric character until it encounters one that isn't a numeral. Though I agree with keeping it out of the for-loop conditional so it doesn't get evaluated on every iteration.
```suggestion
                int lines = ss_toi(words[4]);
```

desktop_version/src/Script.cpp
```diff
@@ -519,9 +520,13 @@ void scriptclass::run(void)
                 textx = ss_toi(words[2]);
                 texty = ss_toi(words[3]);
 
+                // If words[4] ends with L, set large to true.
+                textlarge = (words[4].size() > 0 && words[4][words[4].size() - 1] == 'l');
```
InfoTeddy, line 524: Use `endsWith`. Then the code becomes self-documenting and the comment is unnecessary.

Also, check for `L` too, just in case someone does `text,,,,26L`, which wouldn't lowercase the last argument due to not ending with an argument separator (this doesn't happen with `say,26L` because the simplified parser passes the argument through with an ending argument separator).
```suggestion
                textlarge = endsWith(words[4].c_str(), "l") || endsWith(words[4].c_str(), "L");
```

desktop_version/src/Script.cpp
```diff
@@ -3514,11 +3523,13 @@ bool scriptclass::loadcustom(const std::string& t)
                     add("text(blue,0,0,"+words[1]+")");
                 break;
             }
-            int ti=help.Int(words[1].c_str());
-            int nti = ti>=0 ? ti : 1;
-            for(int ti2=0; ti2<nti; ti2++){
+            int ti = ss_toi(words[1].c_str());
```
InfoTeddy, line 3526: `ss_toi` takes in a `const std::string&`, not a `const char*`.
```suggestion
            int ti = ss_toi(words[1]);
```

InfoTeddy (changes requested):
Will merge after these are fixed.

InfoTeddy (approved):
Thanks!

PR #917, opened 2022-12-06
Add state locking

### LINKED ISSUE #448: The animation to teleport back to the Ship using the post-game map menu SHIP option can be interrupted
Since the animation to teleport back to the Ship uses the state machine, anything that sets the gamestate can interrupt it. This includes falling onto a teleporter, as that sets the gamestate to 2000, or even falling into another room, as in non-finalmode and non-custommode the gamestate is set to 0 whenever you do a screen transition (if you've never done the Comms Relay cutscene).

Potentially, there are also script boxes that the player could move into after activating the animation and interrupt the animation that way, too, but this seems hard to pull off without also triggering a screen transition that would've interrupted the animation anyways.

You can also temporarily interrupt the animation by starting it right before falling into or coming into contact with a hazard. It'll only be a temporary interruption if you respawn in the same room, at which point the animation will resume right after the death animation is done. But it'll be a complete interruption if you respawn in a different room, because the gotoroom will set the gamestate to 0, as explained above.

AllyTally:
## Changes:

This PR adds a few new functions to manage changing the gamestate (`game.state`). While you're teleporting to the ship, it prevents the game from setting a new state until the initial teleport animation is finished. This is done using `lockstate()`, `unlockstate()`, and the generic teleport animation setting the state _manually_.

Fixes #448.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -370,7 +370,6 @@ void Game::init(void)
     disableaudiopause = false;
     disabletemporaryaudiopause = true;
     inputdelay = false;
-
 }
 
 void Game::lifesequence(void)
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -717,7 +717,7 @@ void Game::savetele_textbox(void)
     }
 }
 
-void Game::setstate(int gamestate)
+void Game::setstate(const int gamestate)
 {
     if (!statelocked)
     {
@@ -725,7 +725,7 @@ void Game::setstate(int gamestate)
     }
 }
 
-void Game::setstate(int gamestate, int delay)
+void Game::setstate(const int gamestate, const int delay)
 {
     if (!statelocked)
     {
@@ -742,7 +742,7 @@ void Game::incstate()
     }
 }
 
-void Game::setstatedelay(int delay)
+void Game::setstatedelay(const int delay)
 {
     if (!statelocked)
     {
@@ -763,7 +763,8 @@ void Game::unlockstate()
 void Game::updatestate(void)
 {
     statedelay--;
-    if(statedelay <= 0){
+    if (statedelay <= 0)
+    {
         statedelay = 0;
         glitchrunkludge=false;
     }
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -2677,7 +2677,7 @@ void Game::updatestate(void)
         case 3502:
         {
             incstate();
-            statedelay = 45+15;
+            setstatedelay(45+15);
 
             graphics.createtextboxflipme("  All Crew Members Rescued!  ", -1, 64, 0, 0, 0);
             char buffer[SCREEN_WIDTH_CHARS + 1];
@@ -2698,7 +2698,7 @@ void Game::updatestate(void)
         case 3504:
         {
             incstate();
-            statedelay = 45+15;
+            setstatedelay(45+15);
 
             std::string tempstring = savetime;
             graphics.createtextboxflipme("   Game Time:", 64, 96, 0,0,0);
@@ -2714,7 +2714,7 @@ void Game::updatestate(void)
             break;
         case 3506:
             incstate();
-            statedelay = 45+15;
+            setstatedelay(45+15);
 
             graphics.createtextboxflipme("Total Deaths:", 64, 135, 0,0,0);
             graphics.createtextboxflipme(help.String(deathcounts), 180, 135, 0, 0, 0);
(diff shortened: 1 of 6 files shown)


desktop_version/src/Game.cpp
```diff
@@ -369,6 +370,7 @@ void Game::init(void)
     disableaudiopause = false;
     disabletemporaryaudiopause = true;
     inputdelay = false;
+
```
InfoTeddy, line 373: Not sure why this blank line was added.

desktop_version/src/Game.cpp
```diff
@@ -715,11 +717,54 @@ void Game::savetele_textbox(void)
     }
 }
 
+void Game::setstate(int gamestate)
```
InfoTeddy, line 720:
```suggestion
void Game::setstate(const int gamestate)
```

desktop_version/src/Game.cpp
```diff
@@ -715,11 +717,54 @@ void Game::savetele_textbox(void)
...
+{
+    if (!statelocked)
+    {
+        state = gamestate;
+    }
+}
+
+void Game::setstate(int gamestate, int delay)
```
InfoTeddy, line 728:
```suggestion
void Game::setstate(const int gamestate, const int delay)
```

desktop_version/src/Game.cpp
```diff
@@ -715,11 +717,54 @@ void Game::savetele_textbox(void)
...
+
+void Game::incstate()
+{
+    if (!statelocked)
+    {
+        state++;
+    }
+}
+
+void Game::setstatedelay(int delay)
```
InfoTeddy, line 745:
```suggestion
void Game::setstatedelay(const int delay)
```

desktop_version/src/Game.cpp
```diff
@@ -715,11 +717,54 @@ void Game::savetele_textbox(void)
...
+{
+    statelocked = false;
+}
+
 void Game::updatestate(void)
 {
     statedelay--;
-    if(statedelay<=0){
-        statedelay=0;
+    if(statedelay <= 0){
```
InfoTeddy, line 766:
```suggestion
    if (statedelay <= 0)
    {
```

desktop_version/src/Script.cpp
```diff
@@ -1117,6 +1117,7 @@ void scriptclass::run(void)
             }
             else if (words[0] == "gamestate")
             {
+                // Allow the gamestate command bypass statelock, at least for now
```
InfoTeddy, line 1120:
```suggestion
                // Allow the gamestate command to bypass statelock, at least for now
```

InfoTeddy (changes requested):
Style pass.

I also noticed that `game.state = activetrigger;` wasn't changed to use the `setstate` function, it should probably do that or otherwise the teleport animation can be interrupted by the player moving through a script box.

InfoTeddy (changes requested):
I forgot to check for `statedelay` assignments last time (my bad), so here are the gamestates of `statedelay` assignments that haven't been changed to use `setstatedelay` and should be:

* 3502
* 3504
* 3506
* 3507

desktop_version/src/Game.cpp
```diff
@@ -2905,32 +2943,33 @@ void Game::updatestate(void)
             break;
         }
         case 4003:
-            state = 0;
+            state++;
```
InfoTeddy, line 2946: I don't think this ends up causing any issues, but why was this changed to `state++;`? It should be left the same.

InfoTeddy (changes requested):
Another style pass.

I also noticed that the overload of `setstate` that sets both `state` and `statedelay` isn't used outside of `Game.cpp`. I presume this was done for clarity so it'd be easy to tell that the state delay is being set alongside the state as well. But if that's the case, then why not just axe the overload and go back to calling both functions separately?

AllyTally:
Whoops, the `state++;` was an accident

desktop_version/src/Game.cpp
```diff
@@ -336,7 +336,8 @@ void Game::init(void)
     stat_trinkets = 0;
 
     state = 1;
-    statedelay = 0;
+    setstatedelay(0);
```
InfoTeddy, line 339: Overzealous find-and-replace? This should stay as `statedelay = 0;` because it's an init function.

InfoTeddy (changes requested):
One more style nitpick before I merge this.

AllyTally:
Should be good now

desktop_version/src/Game.cpp
```diff
@@ -336,7 +336,8 @@ void Game::init(void)
     stat_trinkets = 0;
 
     state = 1;
-    statedelay = 0;
+    setstatedelay = 0;
```
InfoTeddy, line 339: No, this is incorrect. `setstatedelay` is a function, not a variable.
```suggestion
    statedelay = 0;
```

InfoTeddy (changes requested):

desktop_version/src/Game.cpp
AllyTally: agh, sorry, was distracted while committing

InfoTeddy: lmao

InfoTeddy (approved):
Will merge once CI passes.

PR #919, opened 2022-12-12
Unhardcoded glitchy and changing roomnames

AllyTally:
## Changes:

Roomname animations ("glitchy" and "changing" ones, used in the final level) are very hardcoded. Roomnames are solely visual, and some fun could be had with making them look cool.

Sample:

```xml
        <SpecialRoomnames>
            <static x="0" y="0" flag="2">This shows if flag 2 is on!</static>
            <glitch x="0" y="0" flag="1">
                <text>On the Waterfront</text>
                <text>On the Vaterfront</text>
            </glitch>
            <transform x="1" y="0" loop="0" flag="1">
                <text>Television Newsveel</text>
                <text>Television Newsvel</text>
                <text>TelevisvonvNewsvel</text>
                <text>TvlvvvsvonvNevsvel</text>
                <text>vvvvvvsvovvNe svel</text>
                <text>vhv vvv'vvovv vevl</text>
                <text>vhv V v'Cvovv vewv</text>
                <text>vhe 9 v'Cvovv vewv</text>
                <text>vhe 9 v'Cvovv Newv</text>
                <text>The 9 O'Cvovk Newv</text>
                <text>The 9 O'Clock News</text>
            </transform>
        </SpecialRoomnames>
```

Additionally, a `setroomname` command has been added, which sets the roomname temporarily to the next script line.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/CustomLevels.cpp
+++ b/desktop_version/src/CustomLevels.cpp
@@ -391,6 +391,7 @@ void customlevelclass::reset(void)
     onewaycol_override = false;
 
     customcolours.clear();
+    map.specialroomnames.clear();
 }
 
 const int* customlevelclass::loadlevel( int rxi, int ryi )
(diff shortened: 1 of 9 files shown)


desktop_version/src/CustomLevels.cpp
```diff
@@ -1243,6 +1245,59 @@ bool customlevelclass::load(std::string& _path)
...
+                    {
+                        if (SDL_strcmp(textElement->Value(), "text") == 0)
+                        {
+                            name.text.push_back(std::string(textElement->GetText()));
+                        }
+                    }
+                }
+
+                // Push to the start, since the for loop which processes these later `break`s
+                map.specialroomnames.insert(map.specialroomnames.begin(), name);
```
InfoTeddy, line 1298: Not sure I understand this comment. Why exactly does it matter that this should be pushed to the start of the vector? The for-loop in `updateroomnames` will still have *O*(*n*) runtime on average, this doesn't seem to do anything.

InfoTeddy (commented):
I still have yet to do a style pass, however this is something I noticed.

desktop_version/src/CustomLevels.cpp
InfoTeddy: I guess this sort of implicit priority is par for the course for this game (e.g. which script box executes first when you touch two script boxes at a time depends on which script box was placed first), but I do find myself thinking there could be something more explicit instead of the order of the tags in the XML. (This is not me saying that it'd be worth it, because I am not sure of that.)

Anyways, regardless of that, it'd be better if you `push_back`ed into the vector as normal, and simply iterated backwards in `updateroomnames`. That way it makes it clearer that it's being handled in reverse in `updateroomnames`.

InfoTeddy (commented):
Since this is intended to un-hardcode room names from the main game, is it possible that the Final Level room names in the main game can be made to use this system as well? (Doesn't need to load from an XML file, just needs to load the room names along with other main game stuff in `mapclass::initmapdata`.)

AllyTally:
This has been done in the latest commit.

desktop_version/src/Map.cpp
```diff
@@ -396,6 +398,59 @@ void mapclass::transformname(int t)
...
+                setroomname(roomname.text[roomname.progress].c_str());
+            }
+            if (roomname.type == TRANSFORM)
+            {
+                roomname.delay--;
+                if (roomname.delay <= 0)
+                {
+                    roomname.progress++;
+                    roomname.delay = 2;
+                    if (roomname.progress >= roomname.text.size())
```
InfoTeddy, line 246: This line produces this warning:
```c++
/A/p/r/VVVVVV/desktop_version/src/Map.cpp: In member function ‘void mapclass::updateroomnames()’:
/A/p/r/VVVVVV/desktop_version/src/Map.cpp:442:43: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  442 |                     if (roomname.progress >= roomname.text.size())
      |                         ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
```
```suggestion
                    if ((size_t) roomname.progress >= roomname.text.size())
```

desktop_version/src/Map.cpp
```diff
@@ -396,6 +398,59 @@ void mapclass::transformname(int t)
...
+    {
+        return;
+    }
+
+    const int rx = game.roomx - 100;
+    const int ry = game.roomy - 100;
+
+    for (int i = specialroomnames.size() - 1; i >= 0; i--)
+    {
+        Roomname &roomname = specialroomnames[i];
```
InfoTeddy, line 217: Use a regular pointer instead of a C++ reference. Trying to limit the amount of C++isms used in the codebase...

InfoTeddy (changes requested):
Style pass.

Also, there is a bug where the room name "Gvnsmoke" is not translated when using a non-English language. I tested it with Dutch (Nederlands), which should translate to "Zuster Klivia".

AllyTally:
Fixed and force pushed.

InfoTeddy (approved):
Thanks!

Issue #859, opened 2022-01-15
Controller button icons within text


Fussmatte:
As we've been talking on the Discord server, I'd like to propose here a suggestion for using controller button icons based on the player's input method. @InfoTeddy tells me that the input layer can tell the game which controller is in use, allowing the possibility to display e.g. Xbox buttons for an Xbox controller, Nintendo buttons for a Switch controller, etc. Usage of generic button icons could be possible as well.

These are some sprites I've drawn to demonstrate. The icons' base size are 10×10 pixels. The generic Start/Select icons are double-wide (20×10 pixels) as per @Dav999-v's suggestion.
![VVVVVVbutton](https://user-images.githubusercontent.com/44736084/149636555-0eb23b87-cea5-426f-947e-9543f2ff11d7.png)

Here's how they could look in action, along with guides to show how they would be aligned within Latin, Japanese and Chinese text:
![VVVVVVbutton_demo](https://user-images.githubusercontent.com/44736084/149637068-01ed0922-1a07-4d44-b098-002137e9edc9.png)
(This is assuming 8x8, 8x12 and 12x12 font sizes respectively — I don't know if these sizes will be used in the end for the latter two, though it does seem likely.)

One issue that the Japanese and Chinese fonts don't have is that the 10x10 icons may be too tall to comfortably display on text with multiple lines (eg dialogue boxes).

In terms of implementation, I think it could be efficient to encode each button value as an internal code (e.g. `Please press $B_USE now`) and replacing the code with either text (e.g. `ENTER`, `ACTION`) or for buttons, the appropriate number of spaces for each font (1/2 for Latin and Chinese, 2/3 for Japanese), then rendering the appropriate button in that space. Even nicer if it can do it on the fly based on the last-used input method (like how Deltarune does it).

Please forgive me if this is too much to dump here, but I thought it was worth sharing.

flibitijibibo:
The action set system can indeed identify specific hardware - however, the layer system actually makes it so you just get glyphs and not necessarily the product/vendor ID. This could be a good excuse to provide both bitmaps and descriptions though, so we can use the customized images here.

The text code is exactly the right thing to do, so definitely move forward with this idea - we can still support it even before action sets are integrated; we can just check which device is in use and hardcode the device/glyph dictionary for now.

Dav999-v:
[...]

PR #922, opened 2023-01-07
Rewrite font system

Dav999-v:
## Changes:

This PR completely overhauls the font system and text printing.

Design goals of the new font system include supporting colored button glyphs, different fonts for different languages, and larger fonts than 8x8 for Chinese, Japanese and Korean, while being able to support their 30000+ characters without hiccups, slowdowns or high memory usage. And to have more flexibility with fonts in general.

Some screenshots of what's possible with the new font system (this is still work in progress, and sorry if I butchered the Chinese - this is just a machine translated demo):

Demo Chinese time trial
Press (A) to teleport

Another goal is to improve the print functions. There were about 10 different ones that were specifically for printing larger text, printing text with a border around it, printing text with a custom alpha, printing right-aligned text... and combinations of those. These are now all replaced by `font::print` and `font::print_wrap`, that take flags as the first argument, which can be combined. CJK and button glyphs introduce the need for even more parameters to know what font should be selected, and there are already a few shortcomings in existing code caused by print functions not being flexible enough, like local code to right-align text or align text around a certain X value, PrintWrap not also supporting borders around the text, and possibly some other things. It also fixes code duplication and overall messiness of the print functions.

Since this PR changes all print functions and `graphics.len`, here's somewhat of a migration guide:
Print function replacements
PrintWrap and len

## Legal Stuff:

  for these changes


Dav999-v:
I think this PR is ready for review now!

This PR covers everything for CJK languages to support larger fonts than 8x8 (up to 12 pixels high), different languages using different fonts, custom levels showing up in the font the level was made for, colored glyphs, and more flexible print functions.

What it does not yet cover is actually using button glyphs in the places where they're needed, by automatically converting an action like "flip" or "interact" to the correct button label or glyph. The button glyphs themselves (see #859) are also not yet included. But the text renderer now fully supports displaying colored glyphs, and everything is planned out for it, so we're pretty close (I'll probably make another PR that prepares everything for that - it'd mostly just need to declare an extendable enum for all actions, an initial function that takes one of these actions and can convert it into "ENTER" or "E" or "<glyph for Steam Deck Y>", and some string changes)

desktop_version/src/Script.cpp
```diff
@@ -2424,6 +2424,19 @@ void scriptclass::run(void)
                     position--;
                 }
             }
+            else if (words[0] == "setfont")
+            {
+#if !defined(NO_CUSTOM_LEVELS)
```
InfoTeddy, line 2429: I know `if !defined` is littered throughout the codebase but I'd like to fix this style sooner rather than later.
```suggestion
#ifndef NO_CUSTOM_LEVELS
```

InfoTeddy (changes requested):
Style pass. About halfway through the commits I started skimming through them since I already had a couple comments anyway.

desktop_version/src/Font.cpp
```diff
@@ -310,36 +310,132 @@ int get_advance(const Font* f, const uint32_t codepoint)
...
-        graphics.draw_grid_tile(f->image, glyph->image_idx, x, y, f->glyph_w, f->glyph_h, 255, 255, 255, 255, scale, scale * (graphics.flipmode ? -1 : 1));
+        r = g = b = colorglyph_bri;
+    }
+
+    graphics.draw_grid_tile(f->image, glyph->image_idx, x, y, f->glyph_w, f->glyph_h, r, g, b, a, scale, scale * (graphics.flipmode ? -1 : 1));
+
+    return glyph->advance * scale;
+}
+
+#define FLAG_PART(start, count) (flags >> start) % (1 << count)
```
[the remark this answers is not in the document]
Dav999-v: Sorry about that - I _have_ actually been amending commits as much as I could throughout this PR's development, which worked many times, but sometimes there was just no way to do so without squashing more and more commits together into an unreasonably big one, or spending a lot of time repeating my fixes across commits and understanding/fixing annoying conflicts (which I have also needed to do). This is a big PR, so things won't be immediately perfect the first time and code moves around. And this is also the first PR where I'm, from the beginning, trying to rewrite history to remove obsolete design decisions and mistakes as if they never happened, so maybe I just need to do it a lot to get better at reshaping history! 😄 

I've also kept in the back of my mind while working on this that we'd probably do a review stream again and I could warn you there whenever something changed in a later commit, but we ended up not doing that.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -1042,26 +1042,120 @@ void FILESYSTEM_enumerateLevelDirFileNames(
...
+        {
+            PHYSFS_freeList(handle->physfs_list);
+            VVV_free(handle->mounted_path);
+            VVV_free(handle);
+        }
+    }
+}
 
-        if (FILESYSTEM_isDirectory(fullName) && *item[0] != '.')
+const char* FILESYSTEM_enumerate(const char* folder, void** p_handle)
```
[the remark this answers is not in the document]
Dav999-v: Originally I did have the handle type as a public struct, but I wasn't really satisfied with the code until I made it private to `FileSystemUtils.cpp` like that - it resulted in many simplifications, off the top of my head:
- The name could be simply `enum_handle` instead of something like `FILESYSTEM_enum_handle`
- `MAX_PATH` was a big problem when I stored the mounted path in the struct, and one of the reasons for the change - originally I had a malloc for the path just because only `FileSystemUtils.cpp` knew about `MAX_PATH`, until I could change it into a fixed-size array by declaring it only in `FileSystemUtils.cpp`.
- I aimed to make the callsite code as simple as possible, to make it easier to use and to have less code duplication like always having to declare the same couple of variables. This is at the cost of a little more complexity in the implementation and a heap allocation, but it should be well-enough documented. I don't really like the idea of copying that `SDL_zero(handle);` to every callsite that just wants to enumerate files.
- If the callsite accesses `enum_handle.item` directly, it gives a very misleading result since it doesn't point to the current item. Though I guess some good naming could've helped, it felt like an implementation detail better not exposed.

desktop_version/src/Font.cpp
```diff
@@ -16,7 +17,14 @@ SDL_Texture* LoadImage(const char *filename, const TextureLoadType loadtype);
 namespace font
 {
 
-Font temp_bfont; // replace with like, a vector of all loaded fonts
+static FontContainer fonts_main = {0};
+static FontContainer fonts_custom = {0};
```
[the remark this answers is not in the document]
Dav999-v: Hm, I don't remember ever seeing that warning, and I'm not seeing it now on my Linux GCC. I do think I wanted to be very explicit about the zero-initialization here, hence the extra 0. I'll try to see which CIs give the warning and whether `{}` resolves it (otherwise I guess not being explicit about it is not the end of the world, might just cause someone to think it's uninitialized data that'll wreak havoc).

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
InfoTeddy: - The name can just be `EnumHandle`; it doesn't have to be something like `FILESYSTEM_enum_handle`.
- It can just be initialized like `enum_handle handle = {};`, then.
- Name it `_item` then. The leading underscore means it's an internal implementation detail that shouldn't be used by the caller (even though leading underscores are randomly used elsewhere in the codebase when they shouldn't be...).

desktop_version/src/Font.cpp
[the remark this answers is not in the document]
InfoTeddy: It's not on CI. I compile with `-Wall -Wextra`.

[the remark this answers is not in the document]
Dav999-v: Oh, I always thought that was the default in our CMakeLists.

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
Dav999-v: Hm, sounds like an acceptable compromise.

desktop_version/src/Font.cpp
```diff
@@ -310,36 +310,132 @@ int get_advance(const Font* f, const uint32_t codepoint)
...
+
+    if (pf.align_cen || pf.align_right)
+    {
+        const int textlen = graphics.len(text) * pf.scale;
+
+        if (pf.align_cen)
+        {
+            if (x == -1)
+            {
+                x = 160;
```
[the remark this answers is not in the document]
Dav999-v: Just marking resolved in advance of force-pushing all fixes...

desktop_version/src/Font.cpp
```diff
@@ -0,0 +1,345 @@
...
+                    continue;
+                }
+
+                int advance = subElem->IntAttribute("advance", -1);
+                int color = subElem->IntAttribute("color", -1);
+
+                for (uint32_t codepoint = start; codepoint <= end; codepoint++)
+                {
+                    GlyphInfo* glyph = get_glyphinfo(f, codepoint);
+                    if (glyph != NULL)
```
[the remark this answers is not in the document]
Dav999-v: Is it okay if I only fix this in the latest commit which inserted that code? This is a case where the code moved around and evolved a whole bunch, and I'd have to dig deep to find every commit which individually places/moves this and repeat myself while either fixing conflicts or carefully peeling commits apart. (Or should I squash everything?)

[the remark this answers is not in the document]
InfoTeddy: Okay sure, it can be fixed only in the latest commit.

Dav999-v:
I've applied part of the fixes, but it's late so I'll continue with the rest and force-push tomorrow.

desktop_version/src/FileSystemUtils.cpp
```diff
@@ -1042,26 +1042,120 @@ void FILESYSTEM_enumerateLevelDirFileNames(
     }
 }
 
-std::vector<std::string> FILESYSTEM_getLanguageCodes(void)
+typedef struct _enum_handle
```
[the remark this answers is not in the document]
Dav999-v: I had a bit too much of a C mindset here :P

In C, I'm also used to prefixing the non-typedef name of a typedef'd struct with an underscore; if a struct is typedef'd, it's unwanted to still also use the `struct`-qualified name, so the underscore makes it clear which of the two is the "good" name, and that the `struct`-qualified name is an implementation detail in a way.

desktop_version/src/Font.cpp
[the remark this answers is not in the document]
Dav999-v: Ok, actually, I found a way to make repeating fixes across commits more bearable: if I `git rebase -i -Xtheirs upstream/master`, the only part I have to worry about is repeating myself across every commit - not also about fixing all the conflicts it causes (since it'll just ignore the fix made to the earlier commit when applying the more recent commit). I'd still recommend if you find a problem during review to also check the latest commit to see if it's still there, but this should at least make the process less of a time-waster for both of us.

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
InfoTeddy: At that point you don't need to name the struct then. You can just do `typedef struct { ... } name;`.

desktop_version/src/FileSystemUtils.cpp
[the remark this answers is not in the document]
Dav999-v: Did all of this, and also made the enumeration stuff a single commit.

Dav999-v:
All requested changes have now been made.

desktop_version/src/Editor.cpp
```diff
@@ -2003,6 +2011,39 @@ static void editormenuactionpress(void)
...
+            {
+                if (option_cur == game.currentmenuoption)
+                {
+                    selected = loc::languagelist[i].font_idx;
+                    break;
+                }
+                option_cur++;
+            }
+        }
+        cl.level_font_name = font::get_main_font_name(selected);
```
InfoTeddy, line 2034: Between this and `case Menu::ed_font` in `Game::createmenu`, I'm having a hard time wrapping my head around the logic here. It seems like the point of `option_cur` is to exclude the 8x8 font from selection? (But then why is there an "other" option? Is that not the 8x8 font? What does "other" even mean?) If so, then wouldn't it be clearer to make an explicit list of fonts somewhere else, and then have this menu simply reference that?

desktop_version/src/Game.cpp
```diff
@@ -4837,6 +4842,7 @@ void Game::serializesettings(tinyxml2::XMLElement* dataNode, const struct Screen
 
     xml::update_tag(dataNode, "lang", loc::lang.c_str());
     xml::update_tag(dataNode, "lang_set", (int) loc::lang_set);
+    xml::update_tag(dataNode, "new_level_font", loc::new_level_font.c_str());
```
InfoTeddy, line 4845: I'm not sure if this should be saved to game settings as the new default level font. We don't do things like, I don't know, save the default author name to game settings for convenience. I don't see the problem with just letting people set the font to the font they want manually whenever they create new levels (which I can't imagine would be too often, and if you're doing something like creating many test levels, there's still a bunch of other stuff you have to set up each time anyways so I don't see the point). And in fact, this might not make a difference anyways if people are creating levels with custom fonts or a bunch of different fonts (they'd still have to go through the menu and change the font).

desktop_version/src/Editor.cpp
```diff
@@ -1459,15 +1461,18 @@ void editorrender(void)
...
-            font::print(PR_BOR | PR_CJK_HIGH | PR_RIGHT, 316, 199, "("+help.String(ed.levx+1)+","+help.String(ed.levy+1)+")", 196, 196, 255 - help.glow);
-
+            int bgheight = 2 + font::height(0);
+            int toolnamelen = font::len(0, toolname);
+            graphics.fill_rect(0,207-bgheight,toolnamelen+8,bgheight+1, graphics.getRGB(32,32,32));
+            graphics.fill_rect(0,208-bgheight,toolnamelen+7,bgheight, graphics.getRGB(0,0,0));
+            font::print(PR_BOR | PR_CJK_HIGH, 2,199, toolname, 196, 196, 255 - help.glow);
+
+            char coords[8];
+            vformat_buf(coords, sizeof(coords), "({x},{y})", "x:int, y:int", ed.levx+1, ed.levy+1);
```
InfoTeddy: Unless the style of coordinates is going to be localizable, this should just be using `SDL_snprintf`.

InfoTeddy (changes requested):
Another style pass. I might merge once these get addressed.

desktop_version/src/Game.cpp
Dav999-v: If someone is for example Japanese, they expect to be able to make a Japanese level. So in a Japanese VVVVVV, the font _has_ to be the Japanese font by default, because they won't immediately know there is an option to change it, or understand why they can't type properly in the editor.

But if you want to always make English levels (which is understandable - I'm Dutch myself, we have people from France, Germany, etc, all not restricting themselves to just communities in their own language, and making English levels instead), having to manually switch to the 8x8 font every time would get annoying, and I do not want to annoy people into setting their entire VVVVVV to English just because they make levels for an international audience.

So, I'd normally be saying the same thing - "just change the font manually whenever you create a new level" - but I really want to avoid making using your preferred language have any annoying downsides that make people switch away from using their preferred language. Same principle will apply to localized LIES/YES/etc enemies: casual gamers might prefer them, but there _has_ to be an option to use the English graphics, otherwise all speedrunners will feel forced to set their _entire game_ to English just so they're not at a disadvantage with the pixel-perfect enemy collision.

desktop_version/src/Editor.cpp
Dav999-v: Hm, I've been thinking it'd actually be good to move to consistently using one style of string formatting instead of two ones mixed together?

desktop_version/src/Editor.cpp
Dav999-v: The font screen looks like this:

Font screen

So, you get an option for each language that needs a special font, and then the "other" option makes the most sense at the bottom - it's a "none of the above", because all other languages use the 8x8 font.

desktop_version/src/Game.cpp
InfoTeddy: > So, I'd normally be saying the same thing - "just change the font manually whenever you create a new level" - but I really want to avoid making using your preferred language have any annoying downsides that make people switch away from using their preferred language.

I guess I'm not fully on board with this option being implicitly made the new default, when there's nowhere that explicitly indicates that, and the user might not expect it to be the case. Maybe add some text stating that the selected font will be the new default?

There's also the issue that I'm not sure how to clear the default option and reset it back to the _default_ default (automatically using the font of the currently selected language) without having to manually edit `settings.vvv`. There doesn't appear to be an "automatic" option in the font selection menu.

> Same principle will apply to localized LIES/YES/etc enemies: casual gamers might prefer them, but there _has_ to be an option to use the English graphics, otherwise all speedrunners will feel forced to set their _entire game_ to English just so they're not at a disadvantage with the pixel-perfect enemy collision.

I thought that this wouldn't be an issue in the first place, because no matter what graphics are drawn, the collision would always use the English graphics. Unless the issue is with the mismatched visuals and collision?

desktop_version/src/Editor.cpp
InfoTeddy: Okay, so "other" is in fact the 8x8 font. Why does the code seemingly exclude it only to add it again later?

desktop_version/src/Editor.cpp
InfoTeddy: I'm not too sure. Ideally, if the string is known at compile-time, then we should use `SDL_snprintf`, because it can be easily checked for correctness by compilers and linters (it's annotated with `SDL_PRINTF_FORMAT_STRING` and `SDL_PRINTF_VARARG_FUNC`) without needing to do run-time testing. VFormat makes sense where it's used currently because it was designed for strings not known at compile-time, that may change at run-time, and those strings are untrusted.

desktop_version/src/Game.cpp
Dav999-v: > I guess I'm not fully on board with this option being implicitly made the new default, when there's nowhere that explicitly indicates that, and the user might not expect it to be the case. Maybe add some text stating that the selected font will be the new default?

I personally don't think it's a huge issue if people make a new level and didn't expect the font to be carried over from the last time they changed it - if anything, they'll go "ah, so I just have to change it back now and _that_ will be remembered". In any case, they already know the menu exists because they used it before, and by seeing text being in the wrong font, it's obvious the game apparently remembers what font they set for their next level.

> There's also the issue that I'm not sure how to clear the default option and reset it back to the _default_ default (automatically using the font of the currently selected language) without having to manually edit `settings.vvv`. There doesn't appear to be an "automatic" option in the font selection menu.

I don't think people need to be aware of when the font is attached to the language and when it isn't (how often do people change a level's font and then change the game's language and then look confused at the font in new levels they make), but if you switch the level font to the current language's font (in other words, you're switching it _back_ to what it was), it reattaches the default font to the current language.

I agree it's not perfect mind-reading, but I think it's the only system that's user-friendly in all cases and does the right thing most of the time (for both inexperienced and impatient users who can't be bothered to read UI text, as well as experienced and attentive users), and it's better than adding specific menu clutter to control what the default should be and being verbose about it, or expecting users to understand text that could easily be misunderstood as "changing the font for this level will change it for all existing levels too, and they will break" or be a hard to understand crammed line that raises about as many questions as "the old level's new font will instead of the old default font now be the new _default_ for any _more_ recent level!". I think it's not too important for people to have to think too hard about it when they need to change the font.

The other option is to always ask what the font should be when creating a new level, but I don't really like that idea, and even then I feel like making a choice should also initialize the selection to that choice the next time you make a new level.

> I thought that this wouldn't be an issue in the first place, because no matter what graphics are drawn, the collision would always use the English graphics. Unless the issue is with the mismatched visuals and collision?

Yeah, I'm talking about the visuals - if you can't "see exactly what you're doing" in your preferred language and you can in English, that'd be an unreasonable downside of using your preferred language for speedrunners.

desktop_version/src/Editor.cpp
Dav999-v: So that it's at the bottom of the list ("other language than all of the above"), and not unrightfully tied to "English" (or any of the other ~15 languages specifically).

desktop_version/src/Editor.cpp
Dav999-v: Fair enough.

It might actually be that we'll end up being asked to make this translatable (because `(1;2)` happened to develop as a strong convention in a specific language or something) but I guess we'll see, and we can always change it back in that case.

desktop_version/src/Editor.cpp
InfoTeddy: So it's a de-duplication measure. Why are only languages with the 8x8 font de-duplicated and not any other fonts that might be used by multiple languages?

desktop_version/src/Game.cpp
InfoTeddy: I think the text can just be as simple as "This will also be the default font for new levels you create." It's a whole different menu, there's enough space for it.

desktop_version/src/Editor.cpp
Dav999-v: Well, I don't really expect any other font than 8x8 to be used by more than one language (the han unification makes CJK characters render mutually incorrect across languages, so you have to have language-specific fonts, and CJK fonts do often target only one of the languages). And even if a font is reused, then you just get two options to choose from that result in the same font (without prioritizing one language and hiding the other, or implementing a system that can display multiple languages in a single option like `繁體中文/한국어`, but that system may also go unused).

In the meantime, there are definitely a lot of languages that use the 8x8 font, so the menu would be filled with language options that all have the same effect.

For what it's worth: Dicey Dungeons also has separate fonts for Simplified Chinese, Traditional Chinese, Korean and Japanese in its files, and there is only one font that is used by more than one language: the Latin+Cyrillic one.

desktop_version/src/Game.cpp
Dav999-v: But then again: why do people need to know upfront? Can't they see it when they make a new level, and put any text in it? If anyone is going to be that confused because they don't remember changing the font for another level, then they won't remember that text either. I don't think it's something people will need to think about that much; if the font doesn't look right, now you know it's being remembered, and you can just change it.

desktop_version/src/Editor.cpp
InfoTeddy: I was thinking that if there were multiple languages with the same font, it would confuse the user if they saw a different language being selected that wasn't the one they had selected, even though both use the same font. This is more of a problem if the user adds some fonts of their own. And this is why I don't think a language ⟷ font mapping is perfect and the list should just be based off of the actual fonts themselves instead of the languages, even if most of the names are going to be something like "the 한국어 font".

desktop_version/src/Game.cpp
InfoTeddy: I've dealt with way too many users to know just how much they can miss what should be obvious. I'd rather put the text in just to be safe rather than sorry, and I don't think there's any downsides to putting the text in.

Dav999-v: That is actually exactly my concern: many users will not read and understand the text anyway, or will just miss it, and otherwise they're made to pause and think about the implications of being warned that they're making a change that will affect them in the future. It may read like a warning, something you need to pay attention to, when in fact, I don't think anyone will find it important enough? And it has a kind of "now's the moment to decide, you cannot change it later" air around it, making it feel important.

I don't see what the issue is if someone changes the font, and then later makes a new level, and the editor is still on that font _without_ the user knowing. (Even if the user knows because they read the text, it might not be on their mind.) As soon as they're surprised by the font in their level, they can change it (they've shown they at least know you can), and then it doesn't bother them again for newer levels. Unless they really expect the game to be able to read their mind.

desktop_version/src/Editor.cpp
Dav999-v: To be clear, the fonts in this menu are only global/main ones, not level-specific custom ones, because the font also applies to the title and description in the levels list. If you want to use a completely custom font in a level, you'd have to use `setfont`. Or just override the assets for the global font in your level by giving them the same name, so that it's the standard global font in the levels list and automatically turns into the custom font in the level.

I have actually considered a "font name" system ("the 한국어 font"), but thought it'd be simpler to just take the names from the language list, since we already had them, it was probably going to be a 1:1 mapping, and I didn't want font names to become anything other than language names. But I guess it wouldn't matter too much if fonts need to have a "name" attribute that also specifies the language name again.

InfoTeddy: Yeah, I'd rather have an explicit list of font names rather than fetching it through the language list (and then asking the user to select the language when in reality they're selecting the font). It would make the code clearer.

While you're at it, please don't call the 8x8 font "other". That gives me vibes of MuseScore 3's "New All" button. Name it anything other than "other", maybe something like "default", "standard", or even "latin/cyrillic alphabet" if you have to.

desktop_version/src/Game.cpp
InfoTeddy: Okay. You can keep it this way then.

I do think it can be improved by having text that says which font is used on the previous screen (and in the list too, but especially the previous screen) without needing the user to go in to the list to check. Because some may not realize you can press Esc to go back, and there's no button in the list to go back to the previous menu without selecting a font.

Dav999-v: Hm, I think that makes sense - it'd be like the slowdown menu, which shows the current setting as well as letting you change it.

I guess on the previous screen it would be something on the bottom left, just like the editor ghosts hint? (Except maybe permanently visible instead of just when you highlight the option, because it _is_ a kind of metadata...)

Dav999-v:
I addressed all the issues, and also some other things.

One thing that's not yet fixed, which was raised on Discord, is that `setfont()` in a script can change the font of a text box while it is fading out (or not being removed). Lemme know if I should fix that in this PR, or if that'd be fine as a separate fix.

InfoTeddy:
> One thing that's not yet fixed, which was raised on Discord, is that `setfont()` in a script can change the font of a text box while it is fading out (or not being removed). Lemme know if I should fix that in this PR, or if that'd be fine as a separate fix.

It's fine as a separate fix.

InfoTeddy (approved):
Looks good to me.

PR #923, opened 2023-01-08
Move the game off of surfaces, and use the render system instead

AllyTally:
## Changes:

This is quite a big PR.

Ever since VVVVVV was initially ported to C++ in 2.0, it has used surfaces from SDL. The downside is, that's all software rendering. This PR moves most things off of surfaces, and all into GPU, by using textures and SDL_Renderer.

Pixel-perfect collision has been kept by keeping a copy of sprites as surfaces. There's plans for pixel-perfect collision to use masks instead of reading pixel data directly, but that's out of scope for this PR.

Everything has been ported over, and from a bit of playtesting, I see no visual differences between the old system and the new one. This PR should help make the code more understandable and easier to mess with, while also providing a fair framerate boost, cutting down on CPU, and it maybe even cuts down on memory a bit.

Also as a quick note, `graphics.reloadresources()` in `main` now gets called _after_ the renderer is created, because textures cannot be created without a renderer.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV
- [x] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes

PR #926, opened 2023-02-14
Fix screen shake during flip mode not flipping the screen

AllyTally:
## Changes:

This fixes a regression introduced by #923 where flip mode no longer flips the screen while the screen is shaking.

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Will merge after CI passes.

PR #930, opened 2023-02-21
Allow multiple lines and wordwrapping in activity zone prompts

Dav999-v:
## Changes:

Activity zone prompts have always been limited to a single line, because the text box had a hardcoded size. A translator requested for the possibility to add a subtitle under music names for the jukebox, and the easiest solution is to make it possible to translate a prompt with multiple lines. This is possible now, and the textbox even wordwraps automatically (using the normal `font::print_wrap`).

music subtitle

(I wouldn't really like to see translations using multiple lines for stuff like "Press ENTER to talk to Vitellary", especially if it wraps with one name and not with another, but if a string is too long, wordwrapping will look better than text running out of the box.)

## Legal Stuff:

  for these changes


InfoTeddy (approved):
thanks lol

PR #931, opened 2023-02-21
Add magic string to the start of the `game` global

### LINKED ISSUE #928: suggestion: build without PIE on macOS, if possible?
The Linux and Windows builds of VVVVVV are not [position-independent executables](https://en.wikipedia.org/wiki/Position-independent_code) because the compilers used for these releases do not enable PIE by default. This allows autosplitters on Linux and Windows to hardcode relevant memory addresses for different versions (or since 64be99d496186fa28ccff4b260c87b733ecb1ac8, run the program with `-addresses` to prevent needing to hardcode future versions).

macOS's compiler enables PIE by default, thus enabling address space layout randomization, which resulted in me making some [wild implementation-specific memory searching code](https://github.com/iliana/vitellary/blob/0db7e43bacacd7f1ffc7cf7e4a55b886646debad/src/game/macos.rs#L17). `-addresses` won't address the issue for macOS because the addresses will be different every time the program starts.

A flag to disable PIE for Linux builds may help in the future as well; if the build environment ends up being updated from CentOS 7, generally any newer Linux distro will enable PIE by default.

(I am posting this issue in the hopes that someone who understands CMake and Clang options more than me will consider updating the build scripts; I am attempting to come up with a solution on my own as well, though.)

iliana:
## Changes:

The `-addresses` command-line option added in 64be99d4 helps autosplitters on platforms where VVVVVV is not built as a position-independent executable. macOS has made it increasingly difficult, or impossible, to build binaries without PIE.

Adding an obvious string to search for will help tools that need to deal with versions of VVVVVV built with PIE. The bytestring to search for is `[vVvVvV]game`, followed by four null bytes (to avoid finding it in the program code section). This identifies the beginning of the game object; addresses to other objects can be figured out by relative offsets printed by `-addresses`, since ASLR can only change where the globals begin.

Partially f i x e s #928; it may still be advisable to figure out how to explicitly disable PIE on Windows/Linux.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -158,6 +158,8 @@ end:
 
 void Game::init(void)
 {
+    SDL_strlcpy(magic, "[vVvVvV]game", sizeof(magic));
+
     roomx = 0;
     roomy = 0;
     prevroomx = 0;
--- a/desktop_version/src/Game.h
+++ b/desktop_version/src/Game.h
@@ -132,6 +132,8 @@ struct CustomLevelStat
 
 class Game
 {
+    char magic[16];
+
 public:
     void init(void);
 


InfoTeddy:
I'll merge if you reword the description so it doesn't automatically close #928.

iliana:
Done (although I think GitHub lets you clear that over on the sidebar on the PR, too!)

InfoTeddy:
Ah, I didn't know that. I'll merge when CI passes.

PR #932, opened 2023-02-23
Add new UTF-8 library, remove utfcpp submodule, fix UTF-8 fixme in next_wrap()

Dav999-v:
## Changes:

This adds a small library I wrote to handle UTF-8.

Usage is meant to be as simple as possible - see for example decoding a UTF-8 string:

```c
const char* str = "asdf";
uint32_t codepoint;
while ((codepoint = UTF8_next(&str)))
{
    // you have a codepoint congrats
}
```

Or encoding a single codepoint to add it to a string:

```cpp
std::string result;
result.append(UTF8_encode(0x1234).bytes);
```

There are some other functions (`UTF8_total_codepoints()` to get the total number of codepoints in a string, `UTF8_backspace()` to get the length of a string after backspacing one character, and `UTF8_peek_next()` as a slightly less fancy version of `UTF8_next()`), but more functions could always be added if we need them.

This allows us to replace utfcpp (utf8::unchecked) - which is also done in this PR - and also fix some less-than-ideal code:
- Some places had to resort to ignoring UTF-8 (`next_wrap`) or using UCS-4→UTF-8 functions (VFormat had to use PHYSFS ones, and one other place had four lines of code including a `std::back_inserter` just for one character)
- The iterator stuff is kinda confusing and verbose anyway.

The reason why I'm doing this now is also because my next button glyphs PR will most likely need it to encode the characters used for the button glyphs (and because `next_wrap` being incomplete might matter for localization).

## Legal Stuff:

  for these changes

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "third_party/tinyxml2"]
 	path = third_party/tinyxml2
 	url = https://github.com/leethomason/tinyxml2/
-[submodule "third_party/utfcpp"]
-	path = third_party/utfcpp
-	url = https://github.com/nemtrif/utfcpp
 [submodule "third_party/FAudio"]
 	path = third_party/FAudio
 	url = https://github.com/FNA-XNA/FAudio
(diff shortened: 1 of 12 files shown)


InfoTeddy (approved):
Thanks!

PR #938, opened 2023-03-04
Fix possible dangling pointers in FontContainer.map_name_idx

Dav999-v:
## Changes:

The hashmap would get populated with the name of each font, as each font was being added. Unfortunately, adding a font would also realloc the storage for fonts, in which the names are also stored... Possibly invalidating the pointers to the names. This is now fixed by populating the hashmap after all the fonts are added.

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Thanks!

PR #939, opened 2023-03-04
Remove old Spanish language files

Dav999-v:
## Changes:

These are from a fan translation that was originally made in 2020. The files were kept around as a possible base for the future Spanish translators, and now that a first version of the new Spanish translation is being tested, it's time to remove this one. These files are making it very annoying for me to test all the new translations and then jump around between different commits (and they're already not in the round 2 translator pack).

Actually, this won't really help with the jumping around different commits part... But the sooner it's removed, the less confusing it will be when different versions of the language pack are floating around and the latest version needs to be added, and the less "Changes not staged for commit" problems you'll get when testing the new language packs. (Or two different españols being on the language screen)

## Legal Stuff:

  for these changes


InfoTeddy (approved):
k

PR #940, opened 2023-03-05
Change `font::len`, `font::print` and `font::print_wrap` text argument from `std::string` to `const char*`

Dav999-v:
## Changes:

Taking `const char*` instead of `const std::string&` means two things:
- If you already had a `const char*` (which is very common especially because `loc::gettext` gives them, and also string literals), your string is no longer converted into a `std::string` when calling `font::len`, `font::print` or `font::print_wrap`.
- If you had a `std::string` (for example due to string concatenation), you now need to call these functions with `.c_str()` (doesn't apply to `font::print` specifically).

In the case of `font::len` and `font::print_wrap`, they were pretty easy to change to taking only `const char*`. Any code that calls these functions needs to add `.c_str()` to any `std::string`s passed, and I didn't need to add them enough times to feel really worried about that (just 4 times for `font::print_wrap` out of the roughly 213 times this function is used, and 9 times for `font::len` out of the rougly 42 times).

On the other hand, `font::print` had so many calls that needed `.c_str()` to be added, that I decided to overload `font::print` instead. This means it can take either a `const char*` (the main function) or a `std::string` (a wrapper). This means C strings don't need to be needlessly converted into `std::string` objects, and existing `font::print` code doesn't need to be modified.

I don't have any metrics or really any idea whether this improves performance in any way and how much, but in my mind, I was imagining needless memory allocations and copying of data many times every frame (because I don't know if the compiler can optimize that away well enough), and now we can be sure that won't unnecessarily happen.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -1181,7 +1181,7 @@ void editorrender(void)
         graphics.fill_rect(0,238-textheight,320,240, graphics.getRGB(32,32,32));
         graphics.fill_rect(0,239-textheight,320,240, graphics.getRGB(0,0,0));
 
-        font::print_wrap(0, 4, 240-textheight, message, 255,255,255, 8, 312);
+        font::print_wrap(0, 4, 240-textheight, message.c_str(), 255,255,255, 8, 312);
     }
     else if(ed.scripteditmod)
     {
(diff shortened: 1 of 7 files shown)


InfoTeddy (approved):
Thanks!

PR #941, opened 2023-03-05
Fix hardestroom not being translated

AllyTally:
## Changes:

`hardestroom` currently stores the current roomname, but it was missing a call to get the translated string. This has been added, fixing the hardest room appearing as untranslated. This bug appeared in 5beaf973cefa0aa5a19655575d89b52febf21149.

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Thanks!

PR #942, opened 2023-03-05
Update c-hashmap submodule

Dav999-v:
## Changes:

- Hashmap keys are now const everywhere except in callbacks, meaning we don't need (char*) and (void*) casts whenever passing a const pointer anymore - so I cleaned up all these casts. (Guess C is a lot less strict with const than C++ is, and string literals aren't even const-qualified in C, but the reason hashmap keys aren't just const everywhere is because you may want to free a key based on a hashmap iteration).
- "no newline at end of file" warnings on macOS are fixed (for some reason, specifically macOS complained about that)
- "'return': conversion from 'uint64_t' to 'uint32_t', possible loss of data" warning on Windows is fixed (though Windows gives lots of these warnings, I guess /W4 is a bit too much...)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Font.cpp
+++ b/desktop_version/src/Font.cpp
@@ -465,7 +465,7 @@ static bool find_font_by_name(FontContainer* container, const char* name, uint8_
     }
 
     uintptr_t i;
-    if (hashmap_get(container->map_name_idx, (char*) name, SDL_strlen(name), &i))
+    if (hashmap_get(container->map_name_idx, name, SDL_strlen(name), &i))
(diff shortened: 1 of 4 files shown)


InfoTeddy:
> * "no newline at end of file" warnings on macOS are fixed (for some reason, specifically macOS complained about that)

This is because of Clang's `-Wnewline-eof`, and macOS uses Clang by default.

InfoTeddy (approved):
Thanks!

Issue #834, opened 2021-09-01
Switch to action sets


leo60228:
This was already mentioned in #829, but I thought it was deserving of its own issue, especially since it:
* Blocks #759 (which itself partially blocks #832)
* Blocks #784 
* Blocks(?) #590 

This might have to wait for https://github.com/libsdl-org/SDL/issues/4464.

flibitijibibo:
Hypothetically we can do this without SDL_ActionSet, but it is absolutely 100% worth the effort to try and develop this in SDL first. I would read _everything_ in the proposal, including all the links and videos, to get a sense of how it would work. Steam Input's native API would probably be our starting template.

flibitijibibo:
Going to scope this down slightly:

While we don't have the SDL interface yet, we can still prepare for it by fixing the way input is bound and read. We need to rework the input as follows:

- `typedef enum VVV_ActionSet` which lists all the various input modes, such as Menu, InGame, and Editor
- Enums for each VVV_ActionSet, which contains all the actions for each mode. `typedef enum VVV_Action_Menu` for example would have `Menu_Accept` and `Menu_Cancel`l.
- Two sets of arrays for each action enum, one for keyboard and one for controllers
    - Keyboard defaults should be assigned via SDL_GetKeyFromScancode (fixes issue described in #590)
    - Rewrite binding menu to support reassigning InGame array values (fixes issue described in #784)

We'll have to wire in the key/button events ourselves for now. When we rewire our input layer to SDL_ActionSet it will resolve the issue described in #759.

flibitijibibo:
This got delayed due to ActionSet's ongoing development, and the fact that it will be in SDL3 which is still in develolpment.

Issue #875, opened 2022-04-24
Steam Deck UX observations


jstuder-gh:
I just played through the game on the Steam Deck and it was a great experience! However there are a few little UX issues that I wanted to bring up (beyond the lack of Steam Deck controller glyphs due in 2.4).

While I'm approaching this mostly from the mindset of a user coming to the Steam Deck and looking for a console-like experience, I feel that some of the suggestions I'm making here would apply more broadly and could improve the overall UX for new users (on any PC platform) looking for a more console-like experience out of the box.

When installing the game for the first time via the Steam UI it installs:

	Version 2.3.6 (Native Linux build)

The default controller layout used is:

	Official Layout for VVVVVV by Terry Cavanagh

I'm not sure if this was intentional and applied by you, the developers, or whether this was set by Valve and their testing team.
This layout (created initially for Steam Controller???) uses Steam Input to map keyboard keys to the buttons and overrides the in-game SDL GameController mapping.

This detail was confusing for me as a user as it was not apparent that the game was not using the in-game controller mappings at first. This lead to some head-scratching moments for me as my attempts to remap controls in the in-game menu did nothing and certain actions (like Right Bumper to restart) were inaccessible given that Steam input was overriding the actual button press in-game.

Also confusing is that the Steam Input layout binds B to Space/flip. This was not bad in-game, but made for some confusion in navigating the menus as the B button would Accept rather than Go back as is standard.

I can imagine that there are quite a few Deck users who will be confused when first booting the game due to conflict due to it using this layout.

---

There are a few bindings that the Steam input mapping provides that I really liked that I feel would be nice to have in the in-game configuration by default. The Steam input mapping had the left "Back" button (above left thumbstick) bound to Menu and the right "Start" button (above right thumbstick) bound to Enter. By default in the in-game controller configuration, the Back and Start buttons do nothing which feels a bit odd to me as a user who has become accustomed to games using those buttons.

* Perhaps those buttons (SDL_GameController_Back and SDL_GameController_Start) could be bound to the Menu and Enter actions respectively?

### Other UX concerns (when not using the Steam input layout):

The option to separate Interact from Enter was not intuitive to me as a user. It was unintuitive to go to the speedrunning options menu in order to activate. I also found it a bit confusing to see the "bind interact" option in the gamepad bindings menu when the option to separate Interact from Enter is false. It led to a bit of confusion when the mapping that I had for interact had no affect in-game and I was still prompted to hit Enter.

[...]

PR #943, opened 2023-03-08
Change text to accommodate button glyphs

Dav999-v:
## Changes:

This PR adds most of the remaining handling needed for #859.

This adds a function that converts an action (such as interacting in-game) to the corresponding button text ("ENTER", "E") or button glyph (PlayStation triangle, Steam Deck Y, etc).

To identify the actions that currently need to be displayed, this commit also adds the initial enums for action sets as described in https://github.com/TerryCavanagh/VVVVVV/issues/834#issuecomment-1015692161.

This also adds a `textbuttons()` scripting command to handle Violet's "Remember that you can press ENTER" dialogue.

Some screenshots to give an impression (again, with hardcoded mappings):

Press (triangle) to talk to Vermilion

Press (X) to Teleport

Remember that you can press (view) to check where you are on the map!

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -71,6 +71,7 @@ endif()
 set(VVV_SRC
     src/BinaryBlob.cpp
     src/BlockV.cpp
+    src/ButtonGlyphs.cpp
     src/CWrappers.cpp
     src/Ent.cpp
     src/Entity.cpp
--- a/desktop_version/lang/ca/cutscenes.xml
+++ b/desktop_version/lang/ca/cutscenes.xml
@@ -76,10 +76,11 @@
     <cutscene id="bigopenworldskip" explanation="">
         <dialogue speaker="purple" english="I&apos;ll be right here if you need any help!" translation="Si necessites ajuda, seré aquí!"/>
     </cutscene>
-    <cutscene id="talkpurple_intro" explanation="">
+    <cutscene id="talkpurple_intro" explanation="***ENTER is OUTDATED***">
         <dialogue speaker="player" english="I&apos;m feeling a bit overwhelmed, Doctor." translation="Tot això m’aclapara una mica, doctora."/>
         <dialogue speaker="player" english="Where do I begin?" translation="Per on començo?"/>
         <dialogue speaker="purple" english="Remember that you can press ENTER to check where you are on the map!" translation="Recorda que pots prémer Retorn per a veure on ets del mapa!"/>
+        <dialogue speaker="purple" english="Remember that you can press {b_map} to check where you are on the map!" translation=""/>
         <dialogue speaker="purple" english="Look for areas where the rest of the crew might be..." translation="Cerca zones on pugui haver-hi altres tripulants..."/>
         <dialogue speaker="purple" english="If you get lost, you can get back to the ship from any teleporter." translation="Si et perds, pots tornar a la nau a partir de qualsevol teletransportador."/>
         <dialogue speaker="purple" english="And don&apos;t worry! We&apos;ll find everyone!" translation="I no pateixis! Segur que trobarem tothom!"/>
--- a/desktop_version/lang/ca/strings.xml
+++ b/desktop_version/lang/ca/strings.xml
@@ -213,6 +213,8 @@
     <string english="Toggle whether you interact with prompts using ENTER or E." translation="Commuta si s’interacciona|fent servir Retorn o E." explanation="prompts: see the `Press {button} to talk to .../activate terminal/teleport` below" max="38*3"/>
     <string english="E" translation="E" explanation="keyboard key E. Speedrunner options menu"/>
     <string english="ENTER" translation="Retorn" explanation="keyboard key ENTER. Speedrunner options menu"/>
+    <string english="ESC" translation="" explanation="keyboard key ESC"/>
+    <string english="ACTION" translation="" explanation="the ACTION key is either the SPACE key, Z or V (this is explained on the title screen). It&apos;s used in strings like `Press ACTION to advance text`"/>
     <string english="Interact button: {button}" translation="Botó d’interacció: {button}" explanation="keyboard key (E or ENTER) is filled in for {button}. Speedrunner options menu" max="38*2"/>
     <string english="fake load screen" translation="pantalla de càrrega" explanation="menu option"/>
     <string english="Fake Load Screen" translation="Pantalla de càrrega" explanation="title, allows the loading screen which counts to 100% to be turned off" max="20"/>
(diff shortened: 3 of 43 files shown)


Dav999-v:
@flibitijibibo there are now functions in `ButtonGlyphs.cpp` for deciding on the correct text and controller glyphs to use instead of keyboard keys, but the functions still need to be written to detect what the input method is and what icons to return when. I'm assuming the main priority right now is showing Steam Deck icons on Steam Deck, but I don't know how to detect whether the game is running on a Steam Deck and what specific buttons need to be displayed there, or which of the D-pad and thumbsticks allow you to move and which don't, for example.

The functions are:
- `bool BUTTONGLYPHS_keyboard_is_available(void)` - returns false if a keyboard is not (easily) accessible, so we can hide some non-vital keyboard-only hints like muting with M and N
- `bool BUTTONGLYPHS_keyboard_is_active(void)` - returns false if the player is actively using a controller instead of a keyboard, so we can hide the "you can press UP or DOWN instead of ACTION to flip"
- `const char* BUTTONGLYPHS_get_wasd_text(void)` - returns "Press arrow keys or WASD to move", or something else (D-pads, thumbsticks, circle pads, ...)
- `const char* BUTTONGLYPHS_get_button(VVV_ActionSet, VVV_Action)` - takes an action (like `VVV_ActionSet_InGame` + `VVV_Action_InGame_Interact`) and returns either a keyboard key string ("ENTER", "E") or a button glyph (these are defined in an enum, for example `return glyph[GLYPH_PLAYSTATION_TRIANGLE];`). The initial action set enums can be found in `ActionSets.h`. (There are almost no actions defined yet, but I never expected _I'd_ be the one to add this file to the game :P)

Some more questions I have (especially for Steam Deck, since that will go through a verification process):
- I left ACTION alone for now - the title screen says "ACTION = Space, Z, or V" and then only uses ACTION throughout the game. Would this be fine for Steam Deck verification to just change the buttons on the title screen? Or would Valve complain about unclear button labeling? It'd be nice if we didn't have to change all the ACTION strings everywhere, as there are quite a few of them, and the reason "ACTION" is used in the first place is the fact that it represents many buttons at the same time... so would it be changed into something like "Press (A)"? If it needs to be "Press any button" or "Press (A), (B) or (X)", then we'd be duplicating all these strings due to grammar issues in translations...
- The level editor is very keyboard-and-mouse focused, will that cause problems for Steam Deck review? How thorough will the review be in needing to be able to access every option and hotkey? Or do they consider an editor to be more out of scope of game content? If it will cause issues, what if we add a warning screen saying the editor is not (yet) optimized for controller usage? It'd be pretty lame to have to completely hide the editor on Steam Deck just for a green checkmark... Or to now have to focus lots of effort on adding controller support.

flibitijibibo:
I usually write detection as:

```
if (controller_connected)
{
    // Rip off FNA
}
else
{
    return KEYBOARD;
}
```

The FNA sample looks like this: https://github.com/FNA-XNA/FNA/wiki/5:-FNA-Extensions#getguidext

"GUID" is just the device product and vendor mashed together.

The editor we may have to hide when the "SteamDeck" environment variable is set to "1", I don't like this idea but Escape Goat 2 got dinged for this exact situation, so it may be our only option without redoing the editor...

flibitijibibo:
Oh also: Action can be replaced with the first binding available - this is something I'm thinking about for SDL_ActionSet too; multiple bindings might have to be squished with / rather than , to avoid grammatical weirdness.

InfoTeddy:
I would be up for adding controller support to the editor. I have access to Super Mario Maker 2, and it's a good example of how to have a controller scheme for editing levels.

> The editor we may have to hide when the "SteamDeck" environment variable is set to "1", I don't like this idea but Escape Goat 2 got dinged for this exact situation, so it may be our only option without redoing the editor...

Does this actually work? Like, did Escape Goat 2 pass review simply by hiding the editor on Deck?

flibitijibibo:
At the moment EG2 is Playable with the only note being the editor - per the notes if we removed the editor at runtime it would get a Verified rating (haven't pitched this to Ian/Kate yet though).

Gateways had a more blunt workaround which was to show an error message while keeping the option where it was, which was also considered acceptable.

flibitijibibo:
Yeah, `ACTION` has to be replaced. The error message makes sense too; probably easier than making another cascading mess of menu enum values :P

Note that `SteamDeck` is mostly meant to be treated like `SteamTenfoot`; as long as we make sure that glyph rendering starts correctly (i.e. don't assume keyboard and then wait for controller input to change the glyphs) we should be able to keep the dynamic support there. It's not on the cert requirements but I try to keep them fully dynamic so people using external controllers still get accurate glyphs; so far I've only had to hardcode it once across my catalog and that was a Unity thing.

I don't know if they've fixed it yet, but the way it's _supposed_ to work is SteamTenfoot/SteamDeck are enabled in game mode, and desktop mode has both disabled, so desktop mode _should_ allow for the editor to work (and we can assume kb/m as much as we want in that case).

Dav999-v:
> as long as we make sure that glyph rendering starts correctly (i.e. don't assume keyboard and then wait for controller input to change the glyphs) we should be able to keep the dynamic support there. It's not on the cert requirements but I try to keep them fully dynamic so people using external controllers still get accurate glyphs; so far I've only had to hardcode it once across my catalog and that was a Unity thing.

So, part of my confusion was because I kinda forgot about VVVVVV already having a menu for controller binding:
game pad menu

As in, I knew it existed, but it hasn't really registered while working on this, also because I've never had any controllers I could see this menu in action with :P And also because according to #875, the game was not actually using the in-game binding settings on Steam Deck, and instead it was using another "official" mapping provided via Steam. So these are the bindings that should be displayed on Steam Deck as well, right?

I guess the game could switch to showing controller glyphs when it sees a controller event, and to keyboard glyphs when it sees a keyboard-related event? (Starting with controller glyphs on Deck, of course)

flibitijibibo:
The Steam bindings are virtualized, so it has the potential to virtually swap input bindings without the game knowing that something has changed (which is partially why it breaks so often). When we actually integrate the native Steam Input API via SDL3 the Steam client UI will change to bind "actions" rather than physical device emulation, at which point the in-game menu will be redone to act as a fallback for non-Steam versions.

Portal 2 is a good reference for the native integration; with Steam Input enabled go to the input settings and try to change the controller bindings, it _should_ spawn the Steam overlay on Deck and Big Picture mode.

Dav999-v:
So for the time being, do we need to hardcode the button glyphs for the Deck? Or will the Deck work with SDL GameController (and the in-game configured bindings) by the time 2.4 comes out, even if it doesn't have SDL3/action sets yet? What I'm reading is that Steam's input virtualizes _keyboard_ keys - not controller buttons?

I have no Steam Deck to verify how VVVVVV's controls work on there (what specific buttons are interact, map menu, etc), nor any controller I can hook up to a computer to check Steam's controller configuration (I can't even open the controller configuration menu; I have now seen Portal 2's integration via a screenshare though). If I know the correct button layout to use, I can add it into the dynamically rendered text.

flibitijibibo:
Deck can safely assume Xbox layout. If we have a special Steam Deck glyph set, that can be saved for when we can actually identify the hardware via Steam Input, because unfortunately Valve _completely obscures_ the physical device information when SI is enabled, so Xbox glyphs are sufficient for passing cert.

Dav999-v:
This is the glyphset we have:

Or at least the 12x12 version of it

It was made to include Steam Deck glyphs, and there is currently no code for detecting specific controllers or showing these glyphs yet anyway, so I'm thinking, might as well use the Steam Deck icons on Steam Deck until we add support for controller glyphs on _all_ systems. That's something this PR doesn't really aim to add yet - it's mainly to make it easy to change which buttons show up in all the text everywhere. Since Steam Deck is the only place that really needs button glyphs right now, I figured it'd be best to optimize for that, and then not add button glyphs for all the regular PCs until we can do it via action sets.

Dav999-v:
I still need to know:

1. Does it sound good to start out by only showing controller glyphs _on Steam Deck_ for now, leaving PC with just the existing keyboard strings (like ENTER), until we have SDL3/action sets? (Unless someone wants to implement controller vendor detection in VVVVVV pre-SDL3?)
2. Given that the Deck uses a VVVVVV-specific mapping defined outside the game that simulates keyboard presses (is all of that accurate?) is the correct course of action to hardcode the glyphs for that mapping (until we can unhardcode it with the SDL3 changes)?
3. What is that specific button->key mapping? As in, would it be `Press (X) to teleport`, `Press (Y) to teleport`, `Press (B) to teleport`... I don't know the correct buttons right now, whether XBox layout or Deck layout.

flibitijibibo:
If you can put in an enum of all the possible devices we support, I'll write in the detection. The important thing is that it must be possible to update the glyphs at _any_ point in the game; a good test is to just have some debug keys where pressing 1-5 (for example) lets you quickly swap glyph sets to see how they look in-game.

We do _not_ want to hardcode glyphs for _any_ platform; even on Deck we want them to be dynamic. When 2.4 launches I'll be going into the Steamworks admin panel and changing the default to be a standard gamepad, so the Xbox glyphs will line up.

For buttons, just assume the first available binding - for example, if Action is bound to A, LS, and DpadUp, it should just print A.

Dav999-v:
There's an enum for all the button glyphs at the top of `ButtonGlyphs.cpp`, with vendors in the identifiers - this is a mix of `NINTENDO`, `DECK`, `PLAYSTATION`, `XBOX`, and `GENERIC` (which has unlabeled directional action buttons and plaintext "START/SELECT").

Changing the glyphs at any point in the game is mostly covered - as long as `BUTTONGLYPHS_get_button(actionset, action)` returns an up-to-date button, every string should just update on the fly. Now that I think of it, Violet's dialogue with a button label in it might be the exception, since it's stored in the textbox until dismissed:

afbeelding

If needed, I can look into making that insta-change as well. Other than that, where completely formatted strings are not kept around in textboxes, they call `BUTTONGLYPHS_get_button()` frame-by-frame.

I mainly suggested hardcoding the glyphs on Deck because of the Deck apparently completely ignoring the in-game binding configuration, so we'd just be showing the wrong glyphs by basing them off the in-game bindings. But if that can be fixed, we don't have to hardcode them.

flibitijibibo:
Yeah, Steam Input is really obnoxious in that it completely obscures the real device information, _unless_ you integrate their API - so what's normally accessible anywhere else is actively taken away from the user/developer until the proprietary API is used. Inversely, however, this means we can just assume Xbox and if the user tries to modify the virtualized layout, they know what they've signed up for, so the stock Xbox glyphs are enough for Deck.

The Violet dialogue can be changed to support dynamic updates; probably will save someone the trouble of figuring out why a specific textbox acts differently for some reason.

Filed #945 and assigned it to me, so no need to worry about that part for this patchset.

Dav999-v:
This should be ready for review now!

As a quick overview of ButtonGlyphs.cpp (@flibitijibibo):
- `typedef enum ButtonGlyphLayout` is the enum for the different controller devices we definitely have icons for - the layouts themselves are in the `glyph_layout` table below (which should be mostly accurate, but feel free to fix things if needed).
- `static ButtonGlyphLayout layout = LAYOUT_GENERIC;` should change to the appropriate layout dynamically (it's used in `sdlbutton_to_glyph()`)
- If a controller button is pressed, **a controller is connected** or an axis is moved, the game will switch to displaying controller glyphs. If a keyboard key is pressed or the last controller is disconnected, the game will switch back to displaying keyboard keys. The bool is initialized to keyboard keys, but when I have my controller already connected when launching the game it also causes an immediate `SDL_CONTROLLERDEVICEADDED` event, so the title screen immediately shows controller glyphs. So this should be fine for Steam Deck as far as I can tell, assuming it behaves the same way with its virtualized gamepad.
- `BUTTONGLYPHS_keyboard_is_available()` should probably be changed to return false on Steam Deck though - that's what hides some things like "Press M to mute" and could also hide the editor until it's controller-ready.

InfoTeddy (approved):
Thanks!

PR #944, opened 2023-03-09
Editor cleanup & refactor

AllyTally:
### The Cleanup
The editor's code is all over the place, spread across dozens of `else if` chains, filled with repeating logic and math, useless code and broken code which somehow works. This PR is an attempt to clean it up, and while it's not perfect, it's a huge improvement.

Tools now get registered using `register_tool`, so everything about a tool is in one place instead of hardcoded. Instead of magic numbers denoting tools, an enum has been created, and logic has been extracted into simple switch/cases, with shared
logic being deduplicated.

A state system has been created, allowing the editor's input, logic and render functions to be cleaned up and organized quite a bit. This also means that the entire editor is no longer drawn over the settings UI, which might help with performance.

While warp lines were being drawn, they also got resized to automatically fit between collision. In `editorrenderfixed`, gravity lines are resized the same way. Doing logic while drawing is very poor practice, so resizing of these has been moved into logic, and merged together. This also means logic has been moved before rendering.

The editor had right-click emulation, when you hold CTRL and click. It never worked well in the past, and even requires a right click to use, so there's not really any point to keeping it around.

Scripts used a weird "hook" system, where script names were extracted into their own list. This was completely unneeded, so it has been replaced with using the `script.customscripts` vector directly.

The script editor has been cleaned up, so the cursor's Y position is relative to the entire script, rather than what's just displaying on the screen currently. This simplifies a lot of code, and I don't know why it was done the other way in the first place.

The editor now draws tiles like how gameplay does, and it even takes advantage of foreground caching.

### Added Features And Cleanup Side-Effects

The script selector and script editor cursors have been sped up, since both lists can be massive, and waiting 6 frames per line is extremely slow and boring. This is still slow and boring, but we don't have proper input repetition yet.

Interpolation for tile placement tools has been added. It uses DDA (https://w.wiki/6RSQ) to draw a line between the previous frame's mouse position, and the current frame's mouse position. This means that there will no longer be gaps in lines of tiles if you move your mouse fast enough (which is actually rather slow, so it gets annoying quickly). The editor's timestep is no longer hardcoded to 24, as I assume that was only done so there would be less gaps in lines of tiles drawn. With interpolation, that is no longer an issue, so I've removed the editor's special case for the timestep.

While I was cleaning up edge-guides, I added red edge-guides to show where spikes are on connected screens, and also added green edge-guides for horizontal gravity line kludge, which I won't even attempt to explain, because I'm pretty bad at explanations.

Entities can now be placed in walls, because there's not really much reason why it should be disallowed, especially since workarounds existed anyway. The warp token destination can also be placed in walls.

Gravity lines can be any length (not just tiles) but the editor divided the width by 8, then later multiplied it by 8, so gravity lines always looked floored to the tile grid. This has been fixed.

### Autotiling

I did a little bit of cleanup, but it remains largely untouched. It's a massive weird system and I think a dedicated PR to clean up autotiling would be better than putting it in this one, especially since that should also modify some autotiling behavior (not saying I **WILL** make that PR, but just thought I'd mention that).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/CustomLevels.cpp
+++ b/desktop_version/src/CustomLevels.cpp
@@ -1424,10 +1424,6 @@ next:
         }
     }
 
-#ifndef NO_EDITOR
-    ed.gethooks();
-#endif
-
     loc::loadtext_custom(_path.c_str());
     font::load_custom(level_font_name.c_str());
 
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -880,7 +880,7 @@ static void editormenurender(int tr, int tg, int tb)
     }
 }
 
-void draw_background_grid(void)
+static void draw_background_grid(void)
 {
     for (int j = 0; j < 30; j++)
     {
@@ -908,7 +908,7 @@ void draw_background_grid(void)
     }
 }
 
-void draw_background(int warpdir)
+static void draw_background(int warpdir)
 {
     extern editorclass ed;
 
@@ -931,7 +931,7 @@ void draw_background(int warpdir)
     }
 }
 
-void draw_edgeguide(int type, int x, int y, bool vertical)
+static void draw_edgeguide(int type, int x, int y, bool vertical)
 {
     static const SDL_Color white = graphics.getRGB(255 - help.glow, 255, 255);
     static const SDL_Color red = graphics.getRGB(255 - help.glow, 127, 127);
@@ -951,7 +951,7 @@ void draw_edgeguide(int type, int x, int y, bool vertical)
     }
 }
 
-void draw_edgeguides(void)
+static void draw_edgeguides(void)
 {
     extern editorclass ed;
 
@@ -977,7 +977,7 @@ void draw_edgeguides(void)
     }
 }
 
-void draw_entities(void)
+static void draw_entities(void)
 {
     extern editorclass ed;
 
@@ -1280,7 +1280,7 @@ void draw_entities(void)
     }
 }
 
-void draw_ghosts(void)
+static void draw_ghosts(void)
 {
     extern editorclass ed;
 
@@ -1305,7 +1305,7 @@ void draw_ghosts(void)
     }
 }
 
-void draw_bounds(void)
+static void draw_bounds(void)
 {
     extern editorclass ed;
 
@@ -1355,7 +1355,7 @@ void draw_bounds(void)
     }
 }
 
-void draw_cursor(void)
+static void draw_cursor(void)
 {
     extern editorclass ed;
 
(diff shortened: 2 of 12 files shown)


desktop_version/src/Graphics.cpp
```diff
@@ -646,6 +645,44 @@ int Graphics::fill_rect(const int x, const int y, const int w, const int h, cons
...
+}
+
+int Graphics::draw_rect(const SDL_Rect* rect, const SDL_Color color)
+{
+    return draw_rect(rect, color.r, color.g, color.b, color.a);
+}
+
+int Graphics::draw_rect(const int x, const int y, const int w, const int h, const int r, const int g, const int b, const int a)
+{
+    const SDL_Rect rect = { x, y, w, h };
```
InfoTeddy, line 672:
```suggestion
    const SDL_Rect rect = {x, y, w, h};
```

desktop_version/src/Editor.cpp
```diff
@@ -1881,12 +1906,13 @@ void editorrenderfixed(void)
...
+        ed.currentghosts++;
+        if (ed.zmod)
+        {
             ed.currentghosts++;
-            if (ed.zmod) ed.currentghosts++;
-        } else {
-            ed.currentghosts = (int) ed.ghosts.size() - 1;
         }
+
+        ed.currentghosts = SDL_max(ed.currentghosts, ed.ghosts.size());
```
InfoTeddy: Here's a useful mnemonic when working with `max` and `min` functions: The `max` function *max*imizes the numbers given (i.e. whichever is bigger), and the `min` function *min*imizes the numbers given (i.e. whichever is smaller).

In this case, `SDL_max` is the wrong function to use, because `ed.ghosts.size()` here will always be bigger no matter what. This results in ghosts always appearing all at once instead of appearing one at a time. Use `SDL_min` instead.
```suggestion
        ed.currentghosts = SDL_min(ed.currentghosts, ed.ghosts.size());
```

desktop_version/src/Editor.cpp
```diff
@@ -2037,6 +2011,7 @@ void editorlogic(void)
         music.play(6); //should be before game.quittomenu()
         ed.settingsmod=false;
     }
+
```
InfoTeddy:
```suggestion
```

desktop_version/src/Editor.cpp
```diff
@@ -142,1876 +125,2000 @@ void editorclass::reset(void)
...
+                font::print(0, 2, 230, loc::gettext("Editor ghost trail is ON"), tr, tg, tb);
         }
         break;
-    case 10:
-        // Checkpoints
-        // If it's not textured as a checkpoint, then just leave it be
-        if (entity->p1 == 0 || entity->p1 == 1)
+    case Menu::ed_desc:
+    {
+
```
InfoTeddy, line 249:
```suggestion
```

desktop_version/src/CustomLevels.cpp
```diff
@@ -1267,15 +1267,15 @@ bool customlevelclass::load(std::string& _path)
...
+                    prev_start = start;
+                    continue;
                 }
 
                 if (headerfound)
                 {
                     script_.contents.push_back(std::string(&pText[prev_start], len));
                 }
 
-next:
```
InfoTeddy, line 1278: Why remove this? There's nothing wrong with a `goto` here.

desktop_version/src/Editor.cpp
```diff
@@ -139,117 +138,94 @@ void editorclass::register_tool(EditorTools tool, const char* name, const char*
...
+        if (script.customscripts[i].name == name)
         {
             script.customscripts.erase(script.customscripts.begin() + i);
+            break;
         }
     }
 }
 
-void editorclass::removehook(const std::string& t)
+void editorclass::create_script(const std::string& name, const std::vector<std::string> contents)
```
InfoTeddy, line 176: Small optimization here, avoids an unnecessary copy.
```suggestion
void editorclass::create_script(const std::string& name, const std::vector<std::string>& contents)
```

desktop_version/src/Editor.h
```diff
@@ -211,50 +210,47 @@ class editorclass
...
-    void addhooktoscript(const std::string& t);
-    void removehookfromscript(const std::string& t);
-    void loadhookineditor(const std::string& t);
-    void clearscriptbuffer(void);
-    void gethooks(void);
-    bool checkhook(const std::string& t);
-    std::vector<std::string> hooklist;
-
-    int hookmenupage, hookmenu;
+    void create_script(const std::string& name, const std::vector<std::string> contents);
```
InfoTeddy, line 237:
```suggestion
    void create_script(const std::string& name, const std::vector<std::string>& contents);
```
(see suggestion at the implementation of this function)

desktop_version/src/Editor.cpp
```diff
@@ -463,6 +463,47 @@ static void draw_edgeguides(void)
...
+                {
+                    // It's to the right...
+                    if (x <= 0)
+                    {
+                        // And touching the left edge!
+                        graphics.fill_rect(x + w - 2, (entity->y % 30) * 8, 2, 8, green);
+                    }
+                }
+            }
+        }
```
InfoTeddy: Please try to avoid nesting if possible, via using `continue;` and combining if-statements.
```suggestion
        const CustomEntity* entity = &customentities[i];
        const bool is_horizontal_gravity_line = entity->t == 11 && entity->p1 == 0;
        if (!is_horizontal_gravity_line)
        {
            continue;
        }

        const int x = entity->p2 * 8;
        const int w = entity->p3;
        const int room_x = (entity->x / 40);
        const int room_y = (entity->y / 30);

        if (room_y != ed.levy)
        {
            continue;
        }

        if (room_x == POS_MOD(ed.levx - 1, cl.mapwidth)
        // It's to the left...
        && x + w >= 312)
        {
            // And touching the right edge!
            graphics.fill_rect(x, (entity->y % 30) * 8, 2, 8, green);
        }
        else if (room_x == POS_MOD(ed.levx + 1, cl.mapwidth)
        // It's to the right...
        && x <= 0)
        {
            // And touching the left edge!
            graphics.fill_rect(x + w - 2, (entity->y % 30) * 8, 2, 8, green);
        }
```
I've also `const`ed the `entity` pointer and named a variable `is_horizontal_gravity_line` so a comment isn't needed.

InfoTeddy (changes requested):
Style pass.

desktop_version/src/Editor.cpp
```diff
@@ -1799,1950 +1667,1760 @@ void editorrenderfixed(void)
...
-                        ed.sby--;
-                    }
-                    key.keybuffer=ed.sb[ed.pagey+ed.sby];
-                    ed.sbx = UTF8_total_codepoints(ed.sb[ed.pagey+ed.sby].c_str());
-                    music.playef(11);
+                    // This is the first thing we found, so let's use it (unless we find a start point later)
+                    testeditor = i;
                 }
             }
         }
```
InfoTeddy, line 2477: This for-loop can be rewritten to be more self-documenting with less indentation. Unfortunately, this is not a GitHub suggestion because GitHub doesn't support suggestions on deleted lines, and with the way the overall diff works out, there's a bunch of deleted lines in this part of the diff. And suggestions are very broken on intermediate commits.
```c++
        startpoint = customentities[i].t == 16;
        const bool is_startpoint_or_checkpoint = startpoint ||
            customentities[i].t == 10;
        if (!is_startpoint_or_checkpoint)
        {
            continue;
        }

        const int tx = customentities[i].x / 40;
        const int ty = customentities[i].y / 30;
        const bool in_room = tx == ed.levx && ty == ed.levy;
        if (!in_room)
        {
            continue;
        }

        if (startpoint)
        {
            // Oh, there's a start point! Let's use this.
            testeditor = i;
            break;
        }
        else if (testeditor == -1)
        {
            // This is the first thing we found, so let's use it (unless we find a start point later)
            testeditor = i;
        }
```

InfoTeddy (changes requested):
Another style pass. I didn't do a full one before the end of the day so I might as well submit what I have before going to sleep.

InfoTeddy (changes requested):
I did not mean to hit "Approve".

InfoTeddy (approved):
Thanks!

PR #947, opened 2023-03-16
Fix font .txt files not being null-terminated

Dav999-v:
## Changes:

This was an oversight when we migrated to the new UTF-8 system - it expects a null-terminated string, but the utfcpp implementation worked with a pointer to the end of the file instead.

I also added an assert in FILESYSTEM_loadFileToMemory() so this is less likely to happen again - because there should be no valid reason to have a NULL pointer for the total file size, as well as not wanting a null terminator to be added at the end of the file.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/FileSystemUtils.cpp
+++ b/desktop_version/src/FileSystemUtils.cpp
@@ -741,6 +741,12 @@ void FILESYSTEM_loadFileToMemory(
         goto fail;
     }
 
+    if (len == NULL && !addnull)
+    {
+        vlog_warn("%s is loaded with len == NULL && !addnull", name);
+        SDL_assert(0 && "Are you sure you don't want a null terminator to be added to these loaded file contents?");
+    }
+
     /* FIXME: Dumb hack to use `special/stdin.vvvvvv` here...
      * This is also checked elsewhere... grep for `special/stdin`! */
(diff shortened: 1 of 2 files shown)


InfoTeddy:
Part of me wonders why we don't just always add a null byte. We already do that for the STDIN buffer after all. It would make things simpler and get rid of needing to deal with one argument when calling `FILESYSTEM_loadFileToMemory`/`FILESYSTEM_loadAssetToMemory`.

Dav999-v:
Yeah I agree - but apparently FIQ had a reason not to do that though. Maybe we need to read back some 2020 conversations about that.

Dav999-v:
I think his reasons were that it'd be cleaner not to add the null byte for binary files (like images), so that's why the boolean was added.

I personally still think it wouldn't be an issue to append a null byte at the end of binary files - we'd still return the file length without the extra byte to the caller, so the null should be considered outside the buffer boundaries when loading images and such anyway. Something might be said about the current way making it clear and explicit that a null byte is added by the file loading function (i.e. the caller doesn't think they have to realloc to add a null byte themselves or something). But this does result in code duplication and we can always just add a comment at the top of the function.

InfoTeddy (approved):
Guess I'll merge this for now, then.

PR #948, opened 2023-03-16
Fix loading font .txt files that contain null bytes

Dav999-v:
## Changes:

An example is Maximally Misleading Miserable Misadventure, which has a font.txt which includes all ASCII characters starting with a 0x00 byte. This would accidentally null-terminate the string too early.

Instead, we now use the total length of the file again, and keep getting the next UTF-8 codepoint until the file ends. We still need to null-terminate it - it protects against incomplete sequences getting the UTF-8 decoder to read out of bounds.

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Thanks!

PR #949, opened 2023-03-18
Fix the game texture clearing during menu shaking

AllyTally:
## Changes:

Because of how `blackout` works, screen shaking must clear the gameplay buffer. `blackout` simply pauses rendering, so if the gameplay buffer gets cleared, then the screen will just be black, otherwise it'll look like the game is "frozen". VVVVVV only uses `blackout` during screen shaking, so it works as intended. However, when reimplementing this behavior in the move to using the SDL_Renderer system, I failed to notice that since my implementation always clears the gameplay buffer when shaking, if you open the menu during a shake, instead of seeing gameplay during the transition animation, you only see black. This has been fixed with a simple `game.blackout` check before clearing the gameplay buffer.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -3173,7 +3173,7 @@ void Graphics::screenshake(void)
     clear();
 
     // Clear the gameplay texture so blackout() is actually black after a screenshake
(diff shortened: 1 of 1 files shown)


InfoTeddy (approved):
Thanks!

PR #952, opened 2023-03-19
Fix textboxes fully fading out during blackout

### LINKED ISSUE #951: Regression: Text boxes fully fade out during blackout
Previously, blackout was bugged and didn't work properly. One of these bugs was that text boxes didn't fully fade out when they were closed during a blackout. The easiest way to see this is to turn screen effects off, then play the main game intro cutscene.

A text box that has not fully faded out during a blackout.

However, since #923 was merged, this is no longer the case.

A blackout with no text box, because the text box fully faded out.

There are custom levels that rely on this effect to make certain things look good, such as teleporter menus made out of text boxes and player movement. So, this should be fixed.

AllyTally:
## Changes:

When `blackout` is active, the screen (to simplify) stops getting drawn to. The excecption is textboxes, which draw anyway. But since the screen isn't being cleared, removed textboxes stay on screen, since that texture isn't being cleared. In the SDL_Renderer PR, I accidentally broke this behavior, so this commit fixes it.

Fixes #951.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Render.cpp
+++ b/desktop_version/src/Render.cpp
@@ -1967,11 +1967,12 @@ void gamerender(void)
     graphics.cutscenebars();
     graphics.drawfade();
 
+    graphics.drawgui();
+
     graphics.set_render_target(graphics.gameTexture);
 
     graphics.copy_texture(graphics.gameplayTexture, NULL, NULL);
 
-    graphics.drawgui();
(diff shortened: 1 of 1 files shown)


InfoTeddy (approved):
Thanks!

PR #954, opened 2023-03-20
Add gamepad rumble feature

Fussmatte:
## Changes:

This PR adds a rumble/vibration feature for players who use gamepads. It's implemented as a toggleable option (currently off by default) in the game pad options submenu.

Currently these events in the game will cause haptic feedback to occur (sorted here roughly from high to low intensity):
- Deaths
- Screen shakes
- Collecting trinkets
- Touching teleporters
- Touching checkpoints
- Touching gravity lines/tokens

What still needs to be done:
- [ ] Determine all events that should have haptic feedback
- [ ] Determine optimal intensity for each event
- [ ] Test feature across multiple game pads (I may need help for this!)

## Legal Stuff:

  for these changes


flibitijibibo:
Functionally this seems fine, though we could maybe make use of HD rumble to do more of a "tap" rather than a traditional rumble spin for some of the more subtle effects. This isn't easily exposed in SDL right now but we support enough devices now that it's probably worth considering for SDL3...

Other than that, will leave the rest to Terry for design feedback.

flibitijibibo:
Filed https://github.com/libsdl-org/SDL/issues/7509

desktop_version/src/Graphics.cpp
```diff
@@ -3184,6 +3185,8 @@ void Graphics::screenshake(void)
     clear();
 
     copy_texture(menuoffTexture, NULL, NULL, 0, NULL, flipmode ? SDL_FLIP_VERTICAL : SDL_FLIP_NONE);
+
+    key.controllerRumble(0x7FFF,10);
```
AllyTally, line 3189: Please don't start rumble inside of rendering. Rendering should be rendering only.

desktop_version/src/KeyPoll.cpp
```diff
@@ -496,3 +496,20 @@ bool KeyPoll::controllerWantsDown(void)
 {
     return buttonmap[SDL_CONTROLLER_BUTTON_DPAD_DOWN] || yVel > 0;
 }
+
+int KeyPoll::controllerRumble(Uint16 intensity, Uint32 duration_ms)
+{
+    if (game.rumble)
+    {
+        SDL_Event evt;
+        SDL_GameController *toOpen = SDL_GameControllerOpen(evt.cdevice.which);
```
AllyTally, line 505: You open a controller handle, but don't close it later. You probably shouldn't open a handle every single time rumble is called, as well.

desktop_version/src/KeyPoll.cpp
```diff
@@ -496,3 +496,20 @@ bool KeyPoll::controllerWantsDown(void)
...
+        return SDL_GameControllerRumble(toOpen,intensity,intensity,duration_ms);
+    }
+}
+
+int KeyPoll::controllerRumbleStop(void)
+{
+    SDL_Event evt;
+    SDL_GameController *toOpen = SDL_GameControllerOpen(evt.cdevice.which);
```
AllyTally, line 513: See previous comment.

AllyTally (changes requested):
Quick review.

desktop_version/src/KeyPoll.cpp
flibitijibibo: This _slightly_ works because the handle, based on the instance ID, should already exist, and SDL_Quit clears out all refcounts from what I remember.

That said: This `evt` value isn't initialized anyway, so instead what we can do is loop through the existing controllers opened via SDL_GameControllerOpen in the hotplugging routines, avoiding the refcount and the uninitialized evt.

PR #956, opened 2023-03-22
Make Violet's button dialogue work in cutscene test

Dav999-v:
## Changes:

This was easier than I expected - just add an optional buttons="1" attribute to cutscenes.xml. It's treated like the speaker attribute - it's only there as context for the translator, and for the cutscene test.

Before:

öb_mapÁ

After:

SELECT

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Thanks!

PR #957, opened 2023-03-23
Add /MP to CMakeLists for Windows for (slightly) faster CI

Dav999-v:
## Changes:

The CI has always been fairly slow on Windows, so I decided to look into it a little bit to see where the bottlenecks are. Nowadays it seems to be faster than I remember it being - Windows CI runs nowadays seem to be around 3-4 minutes. Still, the faster CI finishes, the more comfortable it is (since we get faster feedback). If every CI could finish after a minute like the Linux one, that'd be even nicer.

The biggest bottleneck on Windows seems to be the initial CMake configure step - on Linux and macOS, this step is near-instant or a couple of seconds at most. On Windows, it can take a lot longer, sometimes more than a minute, most of the time at least 40 seconds, but it varies a lot. It's always slower than on Linux and macOS by a serious amount, though (see the table below).

As for why it's so slow, I don't really know. I tried the `--debug-output` and `--trace` flags of CMake, but all it tells me is that it freezes for 10-20 seconds when switching to a different .cmake file:

cmake log

Hopefully someone can find out what's going on here. Maybe we can also cache the build folder that results from the `cmake ..`, by checking the file hash of `CMakeLists.txt` against known versions (so it only re-cmakes the project once if someone adds a new source file or library or something, and then caches that build folder again for future runs).

---

So I didn't manage to make cmake faster, however, I did find out there's an [`/MP`](https://learn.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=msvc-170) flag in in the VS compiler to enable multi-processor compilation, which we haven't been using. This flag does seem to make compilation faster after some experimenting (CI timing is a bit temperamental, but it still seems significant enough):

Build times

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -350,8 +350,9 @@ endif()
 
 
 if(MSVC)
-    # Statically link Microsoft's runtime library so end users don't have to install it
-    list(APPEND GLOBAL_COMPILE_FLAGS /MT)
+    # Statically link Microsoft's runtime library so end users don't have to install it (/MT)
+    # Also, build with multiple processors (/MP)
+    list(APPEND GLOBAL_COMPILE_FLAGS /MT /MP)
 endif()
 
 


Dav999-v:
Of course the CI for this PR wants to self-demonstrate, and take a lot longer than I've seen before lol

3m 28s

InfoTeddy (approved):
lol

PR #959, opened 2023-03-24
Attempt to make the Windows CI again a bit faster

Dav999-v:
## Changes:

Apart from the CI seemingly liking to randomly twiddle its fingers for dozens of seconds or even minutes (are Windows shared runners overloaded or something?) and CMake (and probably also Visual Studio) still spending a lot of time rebuilding things no matter what I try, I've made some more changes to improve the Windows CI:
- The SDL cache action is now upgraded from `cache@v2` to `cache@v3` (removing a Node.js 12 deprecation warning)
- It now caches the correct SDL folder (it was caching 30 bytes before, I don't know what those 30 bytes were)
- The "Download SDL" step is now fully skipped (with the cache hit return value) instead of using `Test-Path` to check for the folder
- The build folder is now cached, to avoid the initial `cmake -G -A -D -D ..` which often takes really long. It really depends on the CI's mood though - sometimes the next `cmake ..` completes in 5 seconds (success!), sometimes it decides to completely start over anyway, without saying why. And then the first build will still take the longest, even though subsequent builds can all be done in as fast as 11 seconds (I even tried splitting the build folder into one for each build type, but the pattern would stay the exact same unfortunately - first build in the run takes 50+ seconds, subsequent builds, in their own separate folders, are less than 20)

## Legal Stuff:

  for these changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,20 +125,19 @@ jobs:
         submodules: true
 
     - name: Cache SDL
-      uses: actions/cache@v2
+      id: cache-windows-sdl
+      uses: actions/cache@v3
       env:
         cache-name: cache-sdl
       with:
-        path: C:\SDL
-        key: ${{ runner.os }}-build-${{ env.cache-name }}
+        path: C:\SDL2-*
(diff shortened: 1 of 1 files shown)


InfoTeddy (approved):
Thanks!

PR #961, opened 2023-03-26
Add button glyphs support to selection hint on language screen

Dav999-v:
## Changes:

The language screen has a "Press Space, Z, or V to select" hint, which I forgot to update for supporting button glyphs in #943, so this commit does.

```xml
    <action_hint>Press Space, Z, or V to select</action_hint>
    <gamepad_hint>Press {button} to select</gamepad_hint>
```

Press (A) to select

## Legal Stuff:

  for these changes


Dav999-v:
@TerryCavanagh this is probably an easy string to overlook when existing translations get updated, since it's in meta.xml and prefilled in English - so I think translators will need to be specifically informed about this.

InfoTeddy (approved):

PR #974, opened 2023-05-12
Fix lab background's "backboxint" array only using the first generated value

kittrz9:
## Changes:

This change makes boxes drawn in the background for The Lab have a slight variation in brightness to give a slight sense of depth similar to the starfield background. I assume that's what was intended, as the `backboxint` variable is used in that way but would only use the first value that is generated despite creating a value for each box. This change just makes it use the value corresponding to that box.

This change also sorts that array so that darker boxes will not appear over brighter ones. I'm not too sure if this is what would've been intended since it is unsorted in the original code, but this helps to give off that slight depth feeling (though I do feel like it's a very slight difference). The sorting shouldn't have too much of an affect on performance since it's only ran once during initialization and is only ran through for each box.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -1,8 +1,8 @@
 #define GRAPHICS_DEFINITION
 #include "Graphics.h"
 
-#include <SDL.h>
 #include <algorithm>
+#include <SDL.h>
 
 #include "Alloc.h"
 #include "Constants.h"
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -2,6 +2,7 @@
 #include "Graphics.h"
 
 #include <SDL.h>
+#include <algorithm>
 
 #include "Alloc.h"
 #include "Constants.h"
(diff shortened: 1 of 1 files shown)


kittrz9:
Sorry if I interpreted the code wrongly, I just stumbled across it and it seemed like a weird bug that was never caught.

Here's a comparison between the two, with the first image being with the fix (the change in background color is not part of this pull request and is just from the color being chosen randomly). The change is very subtle so I understand if you don't think it should be merged

desktop_version/src/Graphics.cpp
```diff
@@ -2,6 +2,7 @@
 #include "Graphics.h"
 
 #include <SDL.h>
+#include <algorithm>
```
InfoTeddy, line 5: Includes need to be in alphabetical order.
```suggestion
#include <algorithm>
#include <SDL.h>
```

InfoTeddy (changes requested):
Small nit. Still leaving it up to Terry if it should be merged.

TerryCavanagh:
Hmm, yeah, I'm sorry, but I think it's less distracting when the background is all the one colour!

PR #979, opened 2023-05-23
Rewrite autotiling completely

AllyTally:
## Changes:

Autotiling was a mess of functions and if chains and switch statements. This commit makes autotiling better by assigning each direction to one bit in a byte, giving each different combination its own value. This value is then fed into a lookup table to give fine control on which tiles get placed where.

The lab tileset can now use the single tiles which were before unused in the autotiler, and the warp zone's background tool now places the fill used in the main game.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -4,6 +4,7 @@
 #include "Editor.h"
 
 #include <string>
+#include <vector>
 
 #include "Constants.h"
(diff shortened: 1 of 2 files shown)


InfoTeddy (changes requested):
Attempting to place backing in Lab or Warp Zone tileset results in a segfault.

InfoTeddy (approved):

PR #986, opened 2023-07-13
Language updates - Simplified Chinese and Irish edition

Dav999-v:
## Changes:

- New languages: Simplified Chinese and Irish!
- First version of the Simplified Chinese font added at `desktop_version/fonts/font_sc.png` and `font_sc.fontmeta`
  - This includes improvements to the `…` and `;` characters, which weren't in the versions that were posted on Discord and sent to the translator
- Fixed various CJK positioning problems
  - ERROR/WARNING screen title was overlapping with message
  - Crewmate screen names and rescued statuses were overlapping with each other
  - Textboxes on Level Complete screen were overlapping with each other and the crewmate was not vertically centered in the box
  - Some strings were running into each other in flip mode, instead of being moved out of each other (PR_CJK_HIGH and PR_CJK_LOW worked the wrong way around because of FLIP macros being applied to Y coords)
  - In-game esc menu was "bouncy" with selected menu options because of a hardcoded 16 pixel offset
  - Bindings in the gamepad menu were overlapping with each other
  - Fix "+1 Rank!" possibly overlapping with time trial time in CJK font
- Made some minor improvements to Dutch translation
- Fixed iflang, loadtext and setfont not working with uppercase
- Update c-hashmap submodule

## Legal Stuff:

  for these changes


desktop_version/src/Game.cpp
```diff
@@ -725,7 +726,8 @@ void Game::crewmate_textbox(const int r, const int g, const int b)
     graphics.addline("");
     graphics.textboxprintflags(PR_FONT_INTERFACE);
     graphics.textboxcentertext();
-    graphics.textboxpad(5, 2);
+    float spaces_per_8 = font::len(PR_FONT_INTERFACE, " ")/8.0;
```
InfoTeddy, line 729: You are mixing `float`s and `double`s here and it's not clear to me why you are doing so.

A decimal literal defaults to type `double` and must be suffixed with `f` to be of type `float`. You seem to be assigning a `double` literal to a `float` variable. Given that on the very next line, you're passing the variable to `SDL_ceil`, whose argument takes a `double`, you should either...

1. Change the `8.0` literal to `8.0f` and change the `SDL_ceil` calls to `SDL_ceilf`, or
2. Change the type of `spaces_per_8` to `double`.

Given that it doesn't seem like `double` precision is necessary here, I'm inclined to say you should go with (1).

InfoTeddy (changes requested):

desktop_version/src/Game.cpp
Dav999-v: Done and force-pushed.

InfoTeddy (approved):

PR #990, opened 2023-08-22
Controller layout detection

### LINKED ISSUE #945: Controller glyph detection support
Not much to see here, but I've done this a million times for other games so I'll write this in once #943 is merged.

flibitijibibo:
The only weird-ish thing here is that we're using the Valve vendor ID to assign the Deck layout, but that technically applies to _all_ virtual gamepads. This will eventually fix itself when we migrate to a system that uses Steam Input directly and we can get the hardware information again.

Fixes #945

### DIFF
--- a/desktop_version/src/ButtonGlyphs.cpp
+++ b/desktop_version/src/ButtonGlyphs.cpp
@@ -167,6 +167,44 @@ void BUTTONGLYPHS_keyboard_set_active(bool active)
     keyboard_is_active = active;
 }
 
+void BUTTONGLYPHS_update_layout(Uint16 vendor, Uint16 product)
+{
+    if (vendor == 0x054c)
+    {
+        layout = LAYOUT_PLAYSTATION;
+    }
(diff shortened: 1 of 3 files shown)


InfoTeddy:
So right now, my PowerA Xbox controller won't be detected as Xbox. Will that get fixed when we migrate to Steam Input?

flibitijibibo:
That's the idea, yeah - at the same time, we can add to our own vendor/product database here as needed; normally I just fall back to Xbox and don't have a generic layout at all so I don't have a good database of Xbox vs. Generic.

flibitijibibo:
Alrighty, for now let's do that - generic will be worth keeping when we have more thorough hardware detection in the future.

(If anyone wants to beat me to this, feel free to push directly to this branch and merge, won't be able to push for a few hours)

flibitijibibo:
Made the change to assume Xbox as the fallback, and added a vendor ID for 8BitDo devices. Will merge once CI passes!

PR #991, opened 2023-08-23
Fix texture cache missing in certain situations after resizing the window

### LINKED ISSUE #977: Toggling fullscreen while the map is open, then closing the map shows black
Another issue caused by invalidated cached textures (due to #923), if you toggle fullscreen while the map is open, then close the map, the gameplay that appears underneath is missing -- it's black instead.

### LINKED ISSUE #988: Regression: Editor tiles stop rendering after resizing window in game options
If you go into the game options in the editor (not just the level options screen, the game options screen that's the same as the one on the title screen), then resize the window or use the 'resize to nearest' graphics option, then exit back to the editor, tiles will stop rendering. If you then either resize the window on this screen, or update the tiles, then they will start rendering again.

`git bisect` has determined that this is a regression from @AllyTally's commit 34cc15505b9f62340c58c10a471136c0a4d94d92, which is from @AllyTally's PR #944.

AllyTally:
## Changes:

This PR fixes an issue where resizing the window can cause textures to not properly recache, rendering invisible.

Fixes #988, fixes #977.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Screen.cpp
+++ b/desktop_version/src/Screen.cpp
@@ -170,9 +170,9 @@ void Screen::ResizeScreen(int x, int y)
     else
     {
         int result = SDL_SetWindowFullscreen(m_window, 0);
-        recacheTextures();
         if (result != 0)
         {
+            recacheTextures();
             vlog_error("Error: could not set the game to windowed mode: %s", SDL_GetError());
             return;
         }
@@ -185,6 +185,7 @@ void Screen::ResizeScreen(int x, int y)
                 SDL_WINDOWPOS_CENTERED_DISPLAY(windowDisplay)
             );
         }
+        recacheTextures();
     }
 }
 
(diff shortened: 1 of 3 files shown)


desktop_version/src/Screen.cpp
```diff
@@ -170,9 +170,9 @@ void Screen::ResizeScreen(int x, int y)
     else
     {
         int result = SDL_SetWindowFullscreen(m_window, 0);
-        recacheTextures();
         if (result != 0)
         {
+            recacheTextures();
```
flibitijibibo, line 175: nit: We could probably remove the `return;` and instead make the `if` just below it an `else if`, so that we effectively move recacheTextures to the bottom and avoid any lurking early return bugs.

flibitijibibo (approved):
Makes sense to me - even if there are still other places we need to recache this should be okay.

PR #993, opened 2023-08-23
Remove NO_EDITOR/NO_CUSTOM_LEVELS, disable editor on Steam Deck

### LINKED ISSUE #992: Make NO_EDITOR a runtime option
We pretty much have a "no editor" mode already, but it's done as a build-time option. Eventually the editor might be usable outside of PC, but for platforms like Deck where it depends on the current OS mode being used it helps to be able to support hiding the editor at runtime.

AFAIK this option is basically a subset of NO_CUSTOM_LEVELS, so it shouldn't interfere with existing builds _too_ much, it just means builds with custom levels that don't want the editor will just be hiding it instead of stripping it out entirely (and users can bring it back if they really really want to).

AllyTally:
## Changes:

This PR removes the `NO_EDITOR` and `NO_CUSTOM_LEVELS` defines, which cleans up the code a lot, and they weren't really needed anyways.

This commit also disables the editor on the Steam Deck, ~~and adds a program argument to re-enable the editor, `-enable-editor`.~~ This has been removed in favor of detecting if a keyboard is available when you enter the menu.

When the editor is disabled, this is the screen which gets shown:

Screenshot of the levels menu

Closes #992.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -195,7 +195,6 @@ void scriptclass::run(void)
                 }
                 scriptdelay = 1;
             }
-#if !defined(NO_CUSTOM_LEVELS)
             if (words[0] == "setroomname")
             {
                 ++position;
@@ -268,7 +267,6 @@ void scriptclass::run(void)
                     position--;
                 }
             }
-#endif
             if (words[0] == "destroy")
             {
                 if(words[1]=="gravitylines"){
@@ -1412,14 +1410,12 @@ void scriptclass::run(void)
             }
             else if (words[0] == "rollcredits")
             {
-#if !defined(NO_CUSTOM_LEVELS) && !defined(NO_EDITOR)
                 if (map.custommode && !map.custommodeforreal)
                 {
                     game.returntoeditor();
                     ed.show_note(loc::gettext("Rolled credits"));
                 }
                 else
-#endif
                 {
                     game.gamestate = GAMECOMPLETE;
                     graphics.fademode = FADE_START_FADEIN;
@@ -1768,13 +1764,11 @@ void scriptclass::run(void)
 
                 int max_trinkets;
 
-#if !defined(NO_CUSTOM_LEVELS)
                 if (map.custommode)
                 {
                     max_trinkets = cl.numtrinkets();
                 }
                 else
-#endif
                 {
                     max_trinkets = 20;
                 }
@@ -2423,7 +2417,6 @@ void scriptclass::run(void)
             }
             else if (words[0] == "setfont")
             {
-#ifndef NO_CUSTOM_LEVELS
                 if (words[1] == "")
                 {
                     font::set_level_font(cl.level_font_name.c_str());
@@ -2432,7 +2425,6 @@ void scriptclass::run(void)
                 {
                     font::set_level_font(raw_words[1].c_str());
                 }
-#endif
             }
 
             position++;
@@ -2783,10 +2775,6 @@ void scriptclass::startgamemode(const enum StartMode mode)
         }
         break;
 
-#ifdef NO_CUSTOM_LEVELS
-        UNUSED(gotoerrorloadinglevel);
-#else
-# ifndef NO_EDITOR
     case Start_EDITOR:
         cl.reset();
         ed.reset();
@@ -2824,7 +2812,6 @@ void scriptclass::startgamemode(const enum StartMode mode)
             music.currentsong = -1;
         }
         break;
-# endif /* NO_EDITOR */
 
     case Start_CUSTOM:
     case Start_CUSTOM_QUICKSAVE:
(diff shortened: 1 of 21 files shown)


desktop_version/src/main.cpp
```diff
@@ -380,6 +369,7 @@ int main(int argc, char *argv[])
     char* langDir = NULL;
     char* fontsDir = NULL;
     bool seed_use_sdl_getticks = false;
+    bool editor_disabled = !BUTTONGLYPHS_keyboard_is_available();
```
flibitijibibo, line 372: Note for me: Make sure this checks the SteamDeck environment variable, we can maybe adjust this to keyboard_is_preferred since consoles very very quietly support kb/m 🤫

flibitijibibo (approved):
Seems pretty straightforward, will wait for a +1 in case I missed an issue for no-custom-level configs

desktop_version/src/main.cpp
AllyTally: Oh I didn't see this before committing -- I added it to `keyboard_is_available`, is that okay?

flibitijibibo: lgtm!

TerryCavanagh:
This is a good idea, but maybe the wording on the warning should be changed? Not sure what to, but the current warning doesn't make it clear why the level editor is disabled.

desktop_version/src/main.cpp
Dav999-v: > we can maybe adjust this to keyboard_is_preferred since consoles very very quietly support kb/m 🤫

I guess the conditional in `BUTTONGLYPHS_keyboard_is_available()` could always be changed to also return true if keyboard_is_active, even if SteamDeck?

flibitijibibo: Something along those lines - the general guideline is to start by assuming controller input, and if keyboard/mouse events occur then we can assume it is explicitly requested by the user.

AllyTally:
Does the following seem better: "This version of the game does not support the level editor due to it currently being designed for keyboard and mouse only."

TerryCavanagh:
Not sure! I'll have a think about it and get back to you!

desktop_version/src/main.cpp
```diff
@@ -424,12 +408,6 @@ int main(int argc, char *argv[])
         }
         else if (ARG("-addresses"))
         {
-#ifndef NO_CUSTOM_LEVELS
-            printf("cl         : %p\n", (void*) &cl);
-# ifndef NO_EDITOR
-            printf("ed         : %p\n", (void*) &ed);
-# endif
-#endif
```
InfoTeddy, line 432: Why delete these print lines?

InfoTeddy (changes requested):

InfoTeddy:
A less wordy version, in my opinion, would be "This version of the game does not support the level editor due to needing a keyboard and mouse." We can always change the message later, so I don't think the exact wording should be blocking this from merge.

AllyTally:
I'd say "does not *currently* support" to softly imply that it might be supported in the future. And while we can always change it later, we need to keep in mind translations and the hassle that comes with asking for new translation strings. I think we should agree on a good string before contacting translators.

InfoTeddy:
Yes, but that time of contact doesn't have to be when this gets merged. We can (and should) hold off on contacting them until this string is finalized.

TerryCavanagh:
Sorry for the hold up! Hmm, ok, let's just go with:

> The level editor is not currently supported on Steam Deck, as it requires a keyboard and mouse to use.

and also, for futureproofing:

> The level editor is not currently supported on this device, as it requires a keyboard and mouse to use.

desktop_version/src/Script.cpp
```diff
@@ -2783,10 +2775,6 @@ void scriptclass::startgamemode(const enum StartMode mode)
         }
         break;
 
-#ifdef NO_CUSTOM_LEVELS
-        UNUSED(gotoerrorloadinglevel);
```
InfoTeddy, line 2787: Since the `UNUSED` macro is now unused in this file, you will need to remove `#include "Unused.h"` from the top of the file.

InfoTeddy (changes requested):

Dav999-v:
You should probably also make `BUTTONGLYPHS_keyboard_is_available()` return true if `keyboard_is_active` (or if `BUTTONGLYPHS_keyboard_is_active()`, as discussed in these comments: https://github.com/TerryCavanagh/VVVVVV/pull/993#pullrequestreview-1592152233

desktop_version/src/main.cpp
```diff
@@ -626,6 +609,7 @@ int main(int argc, char *argv[])
 
     game.init();
     game.seed_use_sdl_getticks = seed_use_sdl_getticks;
+    game.editor_disabled = editor_disabled;
```
InfoTeddy, line 612: So as I understand it, as this is the only time that `game.editor_disabled` is assigned, the keyboard must be plugged in on startup to enable the editor, and cannot be plugged in later? Is this desirable behavior?

I don't own a Steam Deck, but I wouldn't find it hard to imagine a situation where a user needs to, for some reason, start the game without a keyboard but then plug it in later. It seems sufficiently not-officially-supported enough that such jank could occur.

AllyTally:
`editor_disabled` is now updated whenever you enter the levels menu, and `-enable-editor` has been removed in favor of that.

InfoTeddy (approved):

PR #995, opened 2023-08-23
Fix UB where image loading fails in certain cases

AllyTally:
## Changes:

Looks like I forgot to set `data` to `NULL`, and uh, there's a double free somewhere. Not anymore.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/GraphicsResources.cpp
+++ b/desktop_version/src/GraphicsResources.cpp
@@ -21,7 +21,9 @@ extern "C"
 
 static SDL_Surface* LoadImageRaw(const char* filename, unsigned char** data)
 {
-    //Temporary storage for the image that's loaded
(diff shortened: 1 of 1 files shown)


desktop_version/src/GraphicsResources.cpp
```diff
@@ -84,7 +86,6 @@ SDL_Surface* LoadImageSurface(const char* filename)
 
     if (optimizedImage == NULL)
     {
-        VVV_free(data);
```
InfoTeddy, line 87: This is not actually a double-free, because `VVV_free` checks for `NULL` first and will set it to `NULL` afterwards. But it's still good to remove this.

InfoTeddy (approved):

PR #997, opened 2023-08-24
Add finalstretch support for custom levels

AllyTally:
## Changes:

This PR adds all of the necessary code to support `finalstretch` in custom levels, including a new command, `finalstretch(on/off)`. This is visual.

This PR also changes the autotile fill for the gray warp zone to be tile 740, to support `finalstretch` better.

## Legal Stuff:

  for these changes


InfoTeddy (changes requested):
This introduces a regression of the bug that #682 (3c5ee78e8aa0b14fe131da597852c73e21d4b899) fixed.

InfoTeddy:
At a glance, this looks okay. My only concern as of now is that people would be annoyed by changing the gray backing tile to 740 and that this would create a split in the look of auto-tiled gray Warp Zone rooms in custom levels before and after 2.5.

PR #1000, opened 2023-08-28
Fix `setfont()` changing font of fading out text boxes

### LINKED ISSUE #925: `setfont()` changes font of fading out text boxes
As reported by @Fussmatte on the VVVVVV Discord while #922 was being reviewed:

https://user-images.githubusercontent.com/59748578/218668448-bcbdcded-31cc-431f-9b78-9ffa7bf8790f.mp4

`setfont()` is called right as the new text box appears, but this changes the font of the previous text box which is already fading out. This looks weird.

I told @Dav999-v that it would be fine to PR the fix separately so filing this just to keep track.

Dav999-v:
## Changes:

Closes #925.

My fix here is to delay the font change until all fading-out textboxes have disappeared. See it as adding a sort of `untilbars` or `untilfade` for text box fadeout, into setfont.

Before:

https://github.com/TerryCavanagh/VVVVVV/assets/44736680/30ed2d7c-4292-41b5-af86-382ae2df0ec1

After:

https://github.com/TerryCavanagh/VVVVVV/assets/44736680/ca5ca402-899a-49cb-9292-4a2225d4ad9b

This doesn't prevent every possible way to change the font of an existing textbox, but you would need to use internal scripting to still do it (and basically be doing it on purpose) - the problem in simplified scripting when you simply do textbox-setfont-textbox is gone.

Also, #1000!

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -2416,13 +2416,29 @@ void scriptclass::run(void)
             }
             else if (words[0] == "setfont")
             {
-                if (words[1] == "")
+                // If any textbox is currently fading out, wait for that first
+                bool blocked = false;
+                for (size_t i = 0; i < graphics.textboxes.size(); i++)
                 {
-                    font::set_level_font(cl.level_font_name.c_str());
+                    if (graphics.textboxes[i].tm == 2)
+                    {
+                        scriptdelay = 1;
+                        position--;
+                        blocked = true;
+                        break;
+                    }
                 }
-                else
+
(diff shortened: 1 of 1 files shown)


InfoTeddy:
As described, I consider the intended effect of this kludge acceptable and don't see a problem with still allowing people to intentionally change the font of an existing text box on purpose, merely just fixing it for simplified scripting, since it would otherwise force people to add the delay themselves.

InfoTeddy (approved):

PR #1002, opened 2023-08-31
Fix some bugs related to switching languages

Daaaav:
## Changes:

This PR fixes several bugs and makes some improvements (described in more detail in the commit messages):

- Fixed getting sent to the title screen if you have no language files and have never had, and open the language screen while in-game
- Added warning messages for missing fonts/lang folders in the bottom left of the main menu
- Fixed activity zone prompt language not getting updated when switching languages while in-game
- Prevent changing language if a script is running (to prevent the hard-to-fix language and font mixups in dialogues)

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/lang/ca/strings.xml
+++ b/desktop_version/lang/ca/strings.xml
@@ -51,7 +51,8 @@
     <string english="Change controller options." translation="Canvia les opcions del controlador." explanation="" max="38*2"/>
     <string english="language" translation="llengua" explanation="menu option"/>
     <string english="Language" translation="Llengua" explanation="title" max="20"/>
-    <string english="Change the language." translation="Canvia la llengua." explanation="" max="38*5"/>
+    <string english="Change the language." translation="Canvia la llengua." explanation="" max="38*2"/>
+    <string english="Can not change the language while a textbox is displayed in-game." translation="" explanation="" max="38*3"/>
     <string english="clear main game data" translation="esborra dades de partides" explanation="menu option"/>
     <string english="clear custom level data" translation="esborra dades de niv. person." explanation="menu option"/>
     <string english="Clear Data" translation="Esborra dades" explanation="title" max="20"/>
--- a/desktop_version/lang/cy/strings.xml
+++ b/desktop_version/lang/cy/strings.xml
@@ -51,7 +51,8 @@
     <string english="Change controller options." translation="Newid opsiynau rheolydd." explanation="" max="38*2"/>
     <string english="language" translation="iaith" explanation="menu option"/>
     <string english="Language" translation="Iaith" explanation="title" max="20"/>
-    <string english="Change the language." translation="Newid yr iaith." explanation="" max="38*5"/>
+    <string english="Change the language." translation="Newid yr iaith." explanation="" max="38*2"/>
+    <string english="Can not change the language while a textbox is displayed in-game." translation="" explanation="" max="38*3"/>
     <string english="clear main game data" translation="clirio data prif gêm" explanation="menu option"/>
     <string english="clear custom level data" translation="clirio data lefel cwstwm" explanation="menu option"/>
     <string english="Clear Data" translation="Clirio Data" explanation="title" max="20"/>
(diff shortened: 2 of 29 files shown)


InfoTeddy (changes requested):
You can still change languages during a gamestate.
Starting in Chinese then changing to English

Dav999-v:
Had a hunch there would still be something like that, but at least it's still an improvement over being able to mess up just any text box...

InfoTeddy:
How about you can't change languages if a text box is on screen at all?

This would fix it without needing to add another check, while also covering cases (if custom level localization is added) where custom levels have a text box on screen not during a script, and also still being minimally intrusive and letting players change language during the rest of gameplay.

Dav999-v:
Yeah that's what I was just thinking about - replacing the script check by a textbox check. But I'll do that tomorrow.

InfoTeddy:
I would also like it so that the menu explains why "language" is disabled when the user hovers over it. Like exactly what happens when time trials or No Death Mode are disabled when invincibility is enabled.

Dav999-v:
Done:

Can not change the language while a textbox is displayed in-game.

InfoTeddy (changes requested):
> * Fixed getting sent to the title screen if you have no language files and have never had, and open the language screen while in-game

This bug is not fixed. It's still possible to go to the title screen from in-game and trigger the assert if you have language files but your language hasn't been set yet.

This situation is possible on a fresh installation without external modification to the save files, if the user bypasses the title screen and gets in-game through command-line playtesting (e.g. by using Ved).

Dav999-v:
@InfoTeddy Fixed in 4d0809da029d3d915d6d88ff64163c802786ce46.

InfoTeddy (commented):
I guess adding the `!startinplaytest` check works, but it seems kind of fragile.

Personally, I would much rather have `Menu::language` add a `game.ingame_titlemode` check, or even better, don't `game.createmenu(Menu::mainmenu);` at all, and instead rely on `game.returnmenu();` to go to the main menu (as it should always be the first one in the stack anyways).

Dav999-v:
`game.ingame_titlemode` is only `true` if you're actually accessing the menu in-game - so the problem would shift to starting command-line playtesting, quitting to the main menu, and going to the language screen from there.

The `game.returnmenu();` almost works, but you'll still see the title screen when setting the language from in-game. It's just you land back in the in-game options menu instead of the main menu, and the in-game music changes to Presenting VVVVVV. Removing `game.menustart = false;` fixes that, but that also completely removes the title screen when first launching the game, which looks and sounds very jarring.

Idea: pressing a "language" menu option should just set `loc::pre_title_lang_menu` to `false`.

InfoTeddy:
> `game.ingame_titlemode` is only `true` if you're actually accessing the menu in-game - so the problem would shift to starting command-line playtesting, quitting to the main menu, and going to the language screen from there.

Yes, but it fixes the problem where it boots you out of gameplay undesirably, which is the main problem with the bug. The other stuff besides booting you out of gameplay are more visual glitches, which of course shouldn't happen, but booting you out of gameplay is a pretty game-breaking bug that I wanted addressed first. And hence why I suggested the other safeguards.

I was suggesting them to fix it in the general case and prevent potential other cases where the main menu could be created from in-game. In my opinion, we should _never_ `game.createmenu(Menu::mainmenu);`, and prefer `game.returnmenu();`; and both paths in `Menu::language` should use `game.returnmenu();`.

Really, a lot of this comes from the fact that the language menu isn't shown at the start of command-line playtesting. I wonder if we should do that, and make it so that in general, no matter how you're launching the game, you will get the language menu if you haven't set a language yet.

> Idea: pressing a "language" menu option should just set `loc::pre_title_lang_menu` to `false`.

That probably works too.

Dav999-v:
> In my opinion, we should _never_ `game.createmenu(Menu::mainmenu);`, and prefer `game.returnmenu();`; and both paths in `Menu::language` should use `game.returnmenu();`.

Since it seems to work, I'll do that too.

> Really, a lot of this comes from the fact that the language menu isn't shown at the start of command-line playtesting. I wonder if we should do that, and make it so that in general, no matter how you're launching the game, you will get the language menu if you haven't set a language yet.

It would probably be nice to do that, if only from the perspective of making sure people don't miss out on knowing their language is now an option, but it's probably not a huge deal. I can't imagine many scenarios where you'd be indefinitely postponing the language screen, it involves only ever starting the game via playtesting, but as soon as you ever start the game normally you will get the screen - or not anymore if you've ever set the language during one of your playtesting sessions.

InfoTeddy (approved):

PR #1004, opened 2023-09-02
Android port (for desktop_version)

leo60228:
## Changes:

This adds an Android port. It does *not* solve #832, however, so an external keyboard or gamepad is required. The project setup is based on [SDL's README-android.md](https://github.com/libsdl-org/SDL/blob/main/docs/README-android.md), which requires a symlink to SDL to be placed inside the game's source tree, meaning that SDL has to be cloned next to VVVVVV.

It can be built by opening the `desktop_version/VVVVVV-android` folder in Android Studio and building it like any other Android app.

A prebuilt version for testing is at <https://drive.google.com/drive/folders/19Z1RSDvR0crhDhcOGC5gFAHKO7zbOZCa>.

## Legal Stuff:

  for these changes


leo60228:
Forgot to mention this in the PR description, but data.zip should be put at `desktop_version/VVVVVV-android/app/src/main/assets/data.zip` in the repo.

desktop_version/src/Vlogging.c
```diff
@@ -5,6 +5,9 @@
 #ifdef _WIN32
 #   define WIN32_LEAN_AND_MEAN
 #   include <windows.h>
+#elif defined(__ANDROID__)
```
flibitijibibo, line 8: In retrospect, I wonder if we could make the SDL_Log path a non-Android-specific path (`VLOG_USE_SDL`?) and for now Android will be the only user of that backend. Windows in particular might benefit from SDL's ability to route to both stdout as well as OutputDebugString.

desktop_version/src/Screen.cpp
```diff
@@ -363,6 +363,12 @@ void Screen::recacheTextures(void)
     }
 }
 
+#ifdef __ANDROID__
+bool Screen::isForcedFullscreen(void)
+{
+    return true;
```
flibitijibibo, line 369: We can probably move this def to _inside_ the function, so that we don't accidentally end up with symbol conflicts if this function prototype ever changes.

flibitijibibo (commented):
A couple nits but otherwise this looks okay from the multiplat end. I'll leave it to the Android peoples to verify the rest!

leo60228:
Three minor notes:
* Google requires using a "Play Asset Delivery" SDK to install assets separately from the app itself when the app bundle is over 150MB. Even with all four architectures, an app bundle is currently only 70MB, so I don't think this is necessary.
* It might be worth adding a Google Play Games backend along with the Steam and GOG ones. I don't have a developer account for Google Play, though.
* I believe SDL3 will be published to Maven, which would avoid the need to copy/paste the Java code and symlink the native code.

leo60228:
I played through the whole game on Android. The only issue I ran into is that the left stick Y axis is reversed on a GameCube controller. This is likely an SDL issue, and the other controllers I've tested (Steam Controller and Xbox One controller, both over USB) worked fine, so I don't think it blocks much.

flibitijibibo:
The GameCube issue may be https://github.com/libsdl-org/SDL/pull/7930

leo60228:
I added a DocumentsProvider to allow cleanly accessing VVVVVV's data from a file manager. The implementation was adapted from the one used by system apps, and I kept the Apache 2.0 license header intact on `FileSystemProvider.java` (which is almost entirely AOSP source).

InfoTeddy:

flibitijibibo (approved):
Shared source changes still look good to me.

@leo60228, if the fix proposed in https://github.com/libsdl-org/SDL/pull/7930 fixes the GameCube inversion issue, definitely give that PR a bump so we can prioritize it for 3.0 and 2.28.4.

InfoTeddy (approved):
Was able to build using Android Studio. Tested as working on my Google Pixel 6 running GrapheneOS with an Xbox USB controller.

PR #1005, opened 2023-09-04
Fix bug with `destroy(moving)`/`destroy(disappear)`

AllyTally:
## Changes:

This commit fixes an obscure bug with `destroy(moving)` and `destroy(disappear)` where, when looping through entities, the code doesn't actually check what the entity is before trying to destroy the block underneath it.

To fix this, we just put the block-destroying code *inside* of the check, instead of being outside of it.

I also fixed the code style because it was horrible.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -290,11 +290,13 @@ void scriptclass::run(void)
                 }
                 else if (words[1] == "platforms" || words[1] == "moving")
                 {
+                    const bool fixed = words[1] == "moving";
+
                     for (size_t edi = 0; edi < obj.entities.size(); edi++)
                     {
                         if (obj.entities[edi].rule == 2 && obj.entities[edi].animate == 100)
                         {
-                            if (words[1] == "moving")
+                            if (fixed)
                             {
                                 obj.disableblockat(obj.entities[edi].xp, obj.entities[edi].yp);
                             }
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -280,7 +280,8 @@ void scriptclass::run(void)
                 }
                 else if (words[1] == "warptokens")
                 {
-                    for (size_t edi = 0; edi < obj.entities.size(); edi++) {
+                    for (size_t edi = 0; edi < obj.entities.size(); edi++)
+                    {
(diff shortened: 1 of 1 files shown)


desktop_version/src/Script.cpp
```diff
@@ -268,25 +268,48 @@ void scriptclass::run(void)
...
+                            obj.disableentity(edi);
+                        }
                     }
-                }else if(words[1]=="warptokens"){
-                    for(size_t edi=0; edi<obj.entities.size(); edi++){
-                        if(obj.entities[edi].type==11) obj.disableentity(edi);
+                }
+                else if (words[1] == "warptokens")
+                {
+                    for (size_t edi = 0; edi < obj.entities.size(); edi++) {
```
InfoTeddy, line 283:
```suggestion
                    for (size_t edi = 0; edi < obj.entities.size(); edi++)
                    {
```

InfoTeddy (changes requested):

desktop_version/src/Script.cpp
```diff
@@ -268,25 +268,49 @@ void scriptclass::run(void)
...
-                        if(fixed) obj.disableblockat(obj.entities[edi].xp, obj.entities[edi].yp);
-                        if(obj.entities[edi].rule==2 && obj.entities[edi].animate==100) obj.disableentity(edi);
+                }
+                else if (words[1] == "platforms" || words[1] == "moving")
+                {
+                    for (size_t edi = 0; edi < obj.entities.size(); edi++)
+                    {
+                        if (obj.entities[edi].rule == 2 && obj.entities[edi].animate == 100)
+                        {
+                            if (words[1] == "moving")
```
InfoTeddy, line 297: Please keep the `bool`, as an explicit optimization in case the compiler can't see it, and as self-documenting code so readers know that `moving` is specifically the fixed version of `platforms`.
```suggestion
                    const bool fixed = words[1] == "moving";

                    for (size_t edi = 0; edi < obj.entities.size(); edi++)
                    {
                        if (obj.entities[edi].rule == 2 && obj.entities[edi].animate == 100)
                        {
                            if (fixed)
```

InfoTeddy (changes requested):

InfoTeddy (approved):

PR #1007, opened 2023-09-04
Convert entity `type`s to an enum

AllyTally:
## Changes:

In an effort to remove magic numbers, I've given every entity type a name. Hopefully I didn't miss anywhere.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Ent.h
+++ b/desktop_version/src/Ent.h
@@ -5,6 +5,34 @@
 
 #define        rn( rx,  ry) ((rx) + ((ry) * 100))
 
+enum EntityTypes
+{
+    EntityType_INVALID = -1,
+    EntityType_PLAYER,
+    EntityType_MOVING,
+    EntityType_DISAPPEARING_PLATFORM,
+    EntityType_QUICKSAND,
+    EntityType_GRAVITY_TOKEN,
+    EntityType_PARTICLE,
+    EntityType_COIN,
+    EntityType_TRINKET,
+    EntityType_CHECKPOINT,
+    EntityType_HORIZONTAL_GRAVITY_LINE,
+    EntityType_VERTICAL_GRAVITY_LINE,
+    EntityType_WARP_TOKEN,
+    EntityType_CREWMATE,
+    EntityType_TERMINAL,
+    EntityType_SUPERCREWMATE,
+    EntityType_TROPHY,
+    EntityType_GRAVITRON_ENEMY = 23,
+    EntityType_WARP_LINE_LEFT = 51,
+    EntityType_WARP_LINE_RIGHT = 52,
+    EntityType_WARP_LINE_TOP = 53,
+    EntityType_WARP_LINE_BOTTOM = 54,
+    EntityType_COLLECTABLE_CREWMATE = 55,
+    EntityType_TELEPORTER = 100
+};
+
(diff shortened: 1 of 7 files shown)


InfoTeddy:
This has conflicts.

Since I can't test this due to conflicts, may I ask what happens if you spawn an invalid entry with `createentity`?

AllyTally:
> may I ask what happens if you spawn an invalid entry with `createentity`?

It spawns the default entity, being the entity which takes player input, does not have a speed cap, uses tile 0, is cyan, etc. Commonly known as a "glitchy Viridian".

InfoTeddy:
So I don't think this was known before, but if you `createentity` 100, that invalid entity is still treated as a teleporter by `entityclass::getteleporter`, and can, for example, fling Viridian out of it with teleporter gamestates.

I don't think any custom levels use this, though.

desktop_version/src/Ent.h
```diff
@@ -5,6 +5,34 @@
 
 #define        rn( rx,  ry) ((rx) + ((ry) * 100))
 
+enum EntityTypes
```
Daaaav, line 8: I'd personally call this `EntityType`, so that a variable of that type isn't an "entity types".

Daaaav (changes requested):
Found a couple of places where the enum values aren't used yet:

Script.cpp:
```cpp
                        if (obj.entities[edi].type == 11)
```

LevelDebugger.cpp:
```cpp
            // For gravity lines, show the true hitbox.
            if (obj.entities[i].type == 9)
            {
                graphics.draw_rect(bounding_box.x - 1, bounding_box.y + 1, bounding_box.w + 2, bounding_box.h, graphics.getRGB(90, 90, 15));
            }
            else if (obj.entities[i].type == 10)
            {
                graphics.fill_rect(bounding_box.x - 2, bounding_box.y - 1, bounding_box.w + 1, bounding_box.h + 2, graphics.getRGB(90, 90, 15));
            }
```
```cpp
                // Mostly contains duplicates, but for ease of use
                switch (entity->type)
                {
                case 0:
                    // ...
                case 1:
                    // ...
                case 7:
                    // ...
                case 8:
                    // ...
                case 9:
                    // ...
                case 10:
                    // ...
                }
```
That last `switch` should probably also have a `default:` in it, since my compiler is now warning about missing cases.

Logic.cpp:
```cpp
                if (obj.entities[i].type >= 51
                && obj.entities[i].type <= 54) /* Don't warp warp lines */
```
```cpp
                if ((obj.entities[i].type >= 51
                && obj.entities[i].type <= 54) /* Don't warp warp lines */
```

NyakoFox:
Fixed and force pushed.

Daaaav (changes requested):
One remaining thing:
```
Entity.cpp: In member function ‘bool entityclass::updateentities(int)’:
Entity.cpp:2245:15: warning: enumeration value ‘EntityType_INVALID’ not handled in switch [-Wswitch]
 2245 |         switch(entities[i].type)
      |               ^
```

Otherwise, lgtm! (I changed `entclass.type` to a wrapper struct and manually edited the patch itself to check we're not missing any of them)

NyakoFox:
Fixed that as well!

PR #1008, opened 2023-09-04
Fix yellow lab autotiling

AllyTally:
## Changes:

A few autotile entries were missing. Here's an example of the issue it caused. This is holding down X and placing down tiles:

image

## Legal Stuff:

  for these changes


InfoTeddy (approved):
Retagged as required because this issue actually results in out-of-bounds indexing. I couldn't reproduce the messed-up tiles on Linux, but I could verify the OoB by running the game through Valgrind.

PR #1009, opened 2023-09-04
Fix blurry images when starting the game filtered

AllyTally:
## Changes:

If the game started in linear mode, the graphics would be created with linear filtering on. This means, when you switch off of linear filtering, all of the assets are still blurry.

This commit makes sure that the assets are loaded as nearest, then it toggles linear on only after.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -3422,6 +3422,9 @@ bool Graphics::reloadresources(void)
     grphx.destroy();
     grphx.init();
 
+    gameScreen.isFiltered = !gameScreen.isFiltered;
+    gameScreen.toggleLinearFilter();
+
     MAYBE_FAIL(checktexturesize("tiles.png", grphx.im_tiles, 8, 8));
     MAYBE_FAIL(checktexturesize("tiles2.png", grphx.im_tiles2, 8, 8));
     MAYBE_FAIL(checktexturesize("tiles3.png", grphx.im_tiles3, 8, 8));
--- a/desktop_version/src/GraphicsResources.cpp
+++ b/desktop_version/src/GraphicsResources.cpp
@@ -262,6 +262,8 @@ static void LoadSprites(const char* filename, SDL_Texture** texture, SDL_Surface
 
 void GraphicsResources::init(void)
 {
+    SDL_SetHintWithPriority(SDL_HINT_RENDER_SCALE_QUALITY, "nearest", SDL_HINT_OVERRIDE);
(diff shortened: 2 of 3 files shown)


desktop_version/src/Graphics.cpp
```diff
@@ -3422,6 +3422,9 @@ bool Graphics::reloadresources(void)
     grphx.destroy();
     grphx.init();
 
+    gameScreen.isFiltered = !gameScreen.isFiltered;
+    gameScreen.toggleLinearFilter();
```
InfoTeddy, line 3426: Not a fan of this double-toggle hack. If `gameTexture` needs to be created differently than other textures, then it should be made more explicit than this.

desktop_version/src/GraphicsResources.cpp
```diff
@@ -262,6 +262,8 @@ static void LoadSprites(const char* filename, SDL_Texture** texture, SDL_Surface
 
 void GraphicsResources::init(void)
 {
+    SDL_SetHintWithPriority(SDL_HINT_RENDER_SCALE_QUALITY, "nearest", SDL_HINT_OVERRIDE);
```
InfoTeddy, line 265: This doesn't seem to be necessary.

desktop_version/src/Screen.cpp
```diff
@@ -39,11 +39,8 @@ void Screen::init(const struct ScreenSettings* settings)
     badSignalEffect = settings->badSignal;
     vsync = settings->useVsync;
 
-    SDL_SetHintWithPriority(
-        SDL_HINT_RENDER_SCALE_QUALITY,
-        isFiltered ? "linear" : "nearest",
-        SDL_HINT_OVERRIDE
-    );
+    SDL_SetHintWithPriority(SDL_HINT_RENDER_SCALE_QUALITY, "nearest", SDL_HINT_OVERRIDE);
```
InfoTeddy, line 42: An `SDL_SetHintWithPriority` call here at all seems to be unnecessary.

InfoTeddy (changes requested):
I'm not sure the understanding of the linear filter hint here is correct. What the correct understanding is, I can't say, because I don't think I really understand it either. But I don't think this is it.

My working theory, though, is that the problem lies with the final texture that gets rendered to the screen (`gameTexture`), and not necessarily any of the asset textures, which would explain why the bug seems to happen only when textures are drawn scaled. 

Also, this doesn't fix the same bug happening when the screen shakes.

PR #1020, opened 2023-09-18
Language patch - Update French to near-latest strings and fix issues

Daaaav:
## Changes:

This PR applies the latest delivery of strings from the French translator to fix issues and translate all untranslated strings (up to and not including the Steam Deck ones).

## Legal Stuff:

  for these changes


InfoTeddy:

Daaaav:
Conflict resolved.

desktop_version/lang/fr/numbers.xml
```diff
@@ -95,12 +95,12 @@
...
-    <number value="94" form="0" english="Ninety Four" translation="quatre-vingt-quatorze"/>
+    <number value="94" form="0" english="Ninety Four" translation="94"/>
     <number value="95" form="0" english="Ninety Five" translation="quatre-vingt-quinze"/>
     <number value="96" form="0" english="Ninety Six" translation="quatre-vingt-seize"/>
-    <number value="97" form="0" english="Ninety Seven" translation="quatre-vingt-dix-sept"/>
-    <number value="98" form="0" english="Ninety Eight" translation="quatre-vingt-dix-huit"/>
-    <number value="99" form="0" english="Ninety Nine" translation="quatre-vingt-dix-neuf"/>
+    <number value="97" form="0" english="Ninety Seven" translation="97"/>
+    <number value="98" form="0" english="Ninety Eight" translation="98"/>
+    <number value="99" form="0" english="Ninety Nine" translation="99"/>
```
InfoTeddy, line 103: Were 94, 97, 98, and 99 specifically removed for being too long?

(In general, please make some notes about these changes or issues fixed, so I don't have to guess at the reasons myself.)

Daaaav: Yep, "Il en reste quatre-vingt-quatorze" broke the limit by 1 character, and these were the specific numbers which were all the longest.

InfoTeddy: Then why not just make it so that text box can have multiple lines?

Daaaav: Because it already can.

---

I would've said it wouldn't be worth the effort just for custom levels that have 95 crewmates in them, but I checked and the textbox _does_ support multiple lines. I do think the single-line limit was there on purpose though - to keep translations just a single line if at all possible, because multiple would make it look worse (like the "Press ENTER to talk to crewmate" strings _can_ be multiline but really shouldn't, and also shouldn't for just one name like Vitellary and not for the others). At this point though, most languages are already in, so that goal has thus been achieved. So we could probably raise the limit now to get rid of the false alarm (and restore the wordy numbers as they were).

InfoTeddy: > I would've said it wouldn't be worth the effort just for custom levels that have 95 crewmates in them

Well, there's also "Quatre-vingt-quatorze sur quatre-vingt-quatorze" for trinkets, and I think levels are likelier to have 100 trinkets than crewmates.

> So we could probably raise the limit now to get rid of the false alarm (and restore the wordy numbers as they were).

Yes, I'd like that to happen.

Daaaav: Alright, then I think let's restore the numbers here, and then I'll make a new PR for raising both this limit and another one after a false alarm in Polish.

PR #1036, opened 2023-10-09
Add translingual map legend

mothbeanie:
## Changes:

This pr changes the map legend to a translingual one with representative icons. They are in the font as 🪙 for trinkets, 💿 for teleporters, and ❓ for teleporters that have not been discovered yet. It also adds the PR_BOR8 flag to generate a border with filled in corners for the map legend, and adjusts their positioning to accommodate for the new icon sizes at different map scales.

## Legal Stuff:

  for these changes


desktop_version/src/Font.h
```diff
@@ -45,6 +45,7 @@
 #define PR_FONT_IDX(idx) ((SDL_clamp(idx, 0, 28) + 3) << 3) /* use given font index */
 #define PR_BRIGHTNESS(value) /* use this brightness 0-255 for the text (accounts for button glyphs correctly) */\
     (((~SDL_clamp((int)(value), 0, 255) & 0xff) << 8))
+#define PR_BOR8 (1 << 16) /* draw a black border around the text, filling in the corners (for the map legend) */
```
InfoTeddy, line 48: Is `PR_BOR8`/`border8` really the best name for this? The effect is indeed very similar to `PR_BOR`/`border` but I feel like a better name is `PR_FULLBOR`/`full_border`. When I first read it, I was wondering where the '8' came from and what it meant (it means the black text making up the border is drawn 8 times instead of 4). Easy to semantically confuse with `PR_FONT_8X8`.

mothbeanie: changed! i was thinking 8 as in 8-directional which seemed super clear to me at the time, but `PR_FULLBOR` and `full_border` are definitely clearer as to what it does :]

InfoTeddy (approved):

PR #1043, opened 2023-10-25
Allow UTF-8 console output on Windows (for -console)

Daaaav:
## Changes:

Right now, Windows assumes all our console output is code page ????. That means our UTF-8 output appears mangled. (I ran into this while testing IME text input by outputting strings to the console)

For a moment I was scared we'd need to do UTF-16 conversions and call Windows-specific print functions like WriteConsoleW() in our vlog functions, but fortunately SetConsoleOutputCP(CP_UTF8) works just fine.

Example of `vlog_warn(loc::gettext("return"))` in German (just to make sure the encoding of the source file is not a factor) before this change:

zurCOUGHck

After:

zurück

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Vlogging.c
+++ b/desktop_version/src/Vlogging.c
@@ -181,6 +181,11 @@ void vlog_open_console(void)
     }
 
     check_color_support();
+
+    if (!SetConsoleOutputCP(CP_UTF8))
+    {
+        vlog_warn("Could not set code page for console output to UTF-8.");
+    }
 }
 #endif /* _WIN32 */
 


flibitijibibo (approved):

InfoTeddy (commented):
Would like some error checking on this to be consistent with the other Win32 functions in `Vlogging.c` having error checking.

Daaaav:
I'm not sure I really agree there to be honest - it doesn't seem critical and "alarm-worthy" if for example some older version of Windows only supports legacy console output (what's the user gonna do about it?) and for nearly all of our output it wouldn't make a difference since it's still in English. Adding an error check there does make the code a little less clean, and I don't think error checks being needed for some functions necessarily makes them needed for all functions.

If you'd like me to do it anyway, just say so.

InfoTeddy:
Yes, add the check.

It's not alarm-worthy, but it's at least a warning. And it should be there to make it clear to the user that it's not our fault if UTF-8 output is mangled.

Daaaav:
Okay well... For the error message to even show up, I had to move it down further, because there's no console output visible before the `freopen`s. Furthermore, the message will be hidden away by all the other output (like the Viridian) because the window will scroll down instantly. Either way, done.

InfoTeddy (approved):

PR #1046, opened 2023-11-01
Add screenshot functions

InfoTeddy:
This adds a screenshot keybind (F6) and hooks Steam's screenshot function to send our own screenshots. This is so that people can easily grab screenshots of pixel perfect internal resolution (320x240) without needing to downscale it themselves.

Issues:

* ~~We need to change the existing F12 keybind, which is currently bound to reloading language files in translator mode. This conflicts with Steam's default screenshot keybind. In this draft, I've just disabled it instead.~~
* ~~Still need to add timestamps to filenames for F6 screenshots.~~
* ~~Flip Mode needs to be manually corrected but it doesn't work as of now.~~

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -316,7 +316,6 @@ add_library(lodepng-static STATIC ${PNG_SRC})
 target_compile_definitions(lodepng-static PRIVATE
     -DLODEPNG_NO_COMPILE_ALLOCATORS
     -DLODEPNG_NO_COMPILE_DISK
-    -DLODEPNG_NO_COMPILE_ENCODER
 )
 
 add_library(c-hashmap-static STATIC ${CHM_SRC})
--- a/desktop_version/src/CWrappers.cpp
+++ b/desktop_version/src/CWrappers.cpp
@@ -2,6 +2,8 @@
 
 #include <SDL.h>
 
+#include "Graphics.h"
+#include "GraphicsUtil.h"
 #include "Localization.h"
 #include "UtilityClass.h"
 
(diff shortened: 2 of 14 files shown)


Daaaav:
I wonder if 640x480 isn't a better resolution for screenshots. In an ideal world, pixel art never gets upscaled by blurry linear scaling, but in practice, it's applied everywhere. And 320x240 is small enough on most screens that people will want their screenshots to be bigger (for example, when uploading your level). For example, if I take a 320x240 screenshot and put it on a forum with `[img width=640]`, it becomes a blurry mess. Same with the standard zoom buttons in browsers and image viewers.

Maybe as a compromise we should store both 320x240 and 640x480 versions at the same time? Which we probably can't cleanly do with the Steam integration... But in that case I'd personally go with just 640x480.

TerryCavanagh:
(yeah, I think I prefer 640x480 as a more practical screenshot size too)

NyakoFox:
I think it'd be best if 640x480 was uploaded to Steam, but we'd also keep around the 320x240 screenshot in a folder somewhere for things like [level database](https://vsix.dev/) thumbnails.

InfoTeddy:
> I think it'd be best if 640x480 was uploaded to Steam, but we'd also keep around the 320x240 screenshot in a folder somewhere for things like [level database](https://vsix.dev/levels/) thumbnails.

So would people be fine with both 1x and 2x screenshots being saved (in different directories)? Just trying to head off any concerns about wasted disk space or extra file management.

Daaaav:
Yeah, that sounds fine by me. VVVVVV's resolution is small enough that storing both 640x480 and 320x240 is still much less pixels than would go into a screenshot of most other games (that have fullscreen resolution). If you don't like the space they take up anyway, just delete them again.

flibitijibibo:
After compression the 1x in particular is probably microscopic, so yeah having both should be okay.

PR #1085, opened 2023-12-21
Add an undoing & redoing system to the editor

AllyTally:
## Changes:

It's very easy to make mistakes in VVVVVV's built-in level editor, with no way to undo them. This PR adds an undo and redo system, bound to CTRL+Z for undo and CTRL+Y for redo. The undo and redo stacks don't have any limits, but could easily happen in the future. VVVVVV's data is small enough where this should never be an issue, however.

No notes show up for undoing and redoing, because keeping track of what specific action you're doing may bloat the system, and would get annoying for the end-user. Notes are a bit annoying in general, even, but that's not relevant for this PR.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -414,6 +414,7 @@ void editorclass::reset(void)
     substate = EditorSubState_MAIN;
 
     undo_buffer.clear();
+    redo_buffer.clear();
 }
 
 void editorclass::show_note(const char* text)
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -2340,7 +2340,7 @@ static void commit_tiles()
     info.room_x = ed.levx;
     info.room_y = ed.levy;
     info.type = EditorUndoType_TILES;
-    SDL_memcpy(&info.tiles, &ed.old_tiles, sizeof(ed.old_tiles));
+    SDL_memcpy(info.tiles, ed.old_tiles, sizeof(ed.old_tiles));
 
     ed.undo_buffer.push_back(info);
     ed.redo_buffer.clear();
@@ -2371,7 +2371,7 @@ static void commit_roomdata_tiles_change()
     info.room_y = ed.levy;
     info.type = EditorUndoType_ROOMDATA_TILES;
     update_old_tiles();
-    SDL_memcpy(&info.tiles, &ed.old_tiles, sizeof(ed.old_tiles));
+    SDL_memcpy(info.tiles, ed.old_tiles, sizeof(ed.old_tiles));
     info.room_data = *cl.getroomprop(ed.levx, ed.levy);
 
     ed.undo_buffer.push_back(info);
@@ -3377,7 +3377,7 @@ void process_editor_buffer(const bool undo)
             cl.settile(ed.levx, ed.levy, x, y, info.tiles[i]);
         }
 
-        SDL_memcpy(&new_info.tiles, &ed.old_tiles, sizeof(ed.old_tiles));
+        SDL_memcpy(new_info.tiles, ed.old_tiles, sizeof(ed.old_tiles));
 
         graphics.foregrounddrawn = false;
         break;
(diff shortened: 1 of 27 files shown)


InfoTeddy:

desktop_version/src/Editor.cpp
```diff
@@ -3192,6 +3320,123 @@ void editorclass::get_input_line(const enum TextMode mode, const std::string& pr
...
 
+void process_editor_buffer(const bool undo)
+{
+    extern editorclass ed;
+
+    std::vector<EditorUndoInfo>* buffer = undo ? &ed.undo_buffer : &ed.redo_buffer;
+
+    if (buffer->size() == 0)
+    {
+        ed.show_note(loc::gettext(undo ? "ERROR: Nothing to undo" : "ERROR: Nothing to redo"));
```
InfoTeddy, line 3331: In general we try to keep all `loc::gettext` strings easily greppable. Please move the `loc::gettext` calls into the ternary, not surrounding it.

Speaking of which, you need to add these strings to the language files too.

InfoTeddy (changes requested):
Style pass.

desktop_version/src/Editor.cpp
```diff
@@ -412,6 +412,8 @@ void editorclass::reset(void)
 
     state = EditorState_DRAW;
     substate = EditorSubState_MAIN;
+
+    undo_buffer.clear();
```
InfoTeddy, line 416: When a new level is loaded, the undo buffer is cleared but not the redo buffer. So if there's anything in the redo buffer, you can press Ctrl+Y and that will be applied in the new level when it shouldn't be.

InfoTeddy (changes requested):

NyakoFox:
Rebased and force-pushed, no longer conflicts (and fixes an issue)

desktop_version/src/Editor.cpp
```diff
@@ -3544,6 +3862,9 @@ void editorinput(void)
                     if (ed.old_entity_text == "")
                     {
                         ed.remove_entity(ed.text_entity);
+                        // Uncommit the past two actions
```
flibitijibibo: This could be improved with a 'why', in case someone looks at this and thinks it's just temporary kludge or something

flibitijibibo (approved):
At a high level this looks good - if anybody's tested this would be good to get a +1 from them as well before merging.

desktop_version/src/Editor.cpp
NyakoFox: Added a few comments to explain the logic, hopefully they makes sense!

flibitijibibo (approved):

NyakoFox:
I asked a few people to test; seems like we're getting some weird memory corruption to do with cross-room undos, so I'm looking into that right now.

NyakoFox:
Fixed the issue; it was just a missed conditional somewhere which caused the stack to inflate & save tile states which didn't change and relied on old data (which made things work strangely) -- not the "memory corruption" I assumed it was. I'm waiting on people to test now and I'll report back once no issues have been found!

desktop_version/src/Editor.cpp
```diff
@@ -2476,27 +2592,34 @@ void editorclass::entity_clicked(const int index)
...
             entity->p1 = (entity->p1 + 1) % 2;
         }
-        SDL_FALLTHROUGH;
+        get_input_line(TEXT_SCRIPT, loc::gettext("Enter script name:"), &entity->scriptname);
+        text_entity = index;
+        break;
     case 19:
         // Script Boxes (and terminals)
+        commit_entity(index);
         get_input_line(TEXT_SCRIPT, "Enter script name:", &entity->scriptname);
```
Daaaav: These two `get_input_line`s are inconsistent with each other - the new one adds a `loc::gettext` that the others didn't have (`get_input_line` is a rare case where the `loc::gettext` is done somewhere later)

desktop_version/src/Editor.cpp
```diff
@@ -3194,6 +3344,127 @@ void editorclass::get_input_line(const enum TextMode mode, const std::string& pr
     old_entity_text = key.keybuffer;
 }
 
+void process_editor_buffer(const bool undo)
```
Daaaav: `process_editor_buffer` is not a very clear name for a function that (I think) handles specifically the action of undoing or redoing.

Daaaav (changes requested):
Few really small things, otherwise looks good I think!

desktop_version/src/Editor.cpp
NyakoFox: Changed to `handle_undo`.

desktop_version/src/Editor.cpp
NyakoFox: Solved.

NyakoFox:
Fixed those; and I've gotten a few people to test again and they also say it looks good.

Daaaav (approved):

flibitijibibo (approved):

PR #1096, opened 2024-01-07
Indicate modes when loading in to gameplay

### LINKED ISSUE #1091: Indicate which modes are active when loading in to gameplay
Many times I (and many others) have forgotten that we have, say, invincibility mode enabled when loading in. There should be text onscreen in the upper-left saying "Invincibility mode enabled" that fades out after a few seconds.

Ditto for other modes that might not be immediately obvious to the user, like glitchrunner. (Flip Mode and slowdown ~~should be sufficiently obvious that they don't need to be indicated~~ should be indicated too for consistency.)

InfoTeddy:
If you load in to gameplay with invincibility mode, glitchrunner mode, Flip Mode, or slowdown enabled, then there will be text displayed on screen for a few seconds that says so.

This is to serve as a useful reminder. A common pitfall with using  invincibility is forgetting to turn it off when you don't want it anymore. What usually happens is that players forget that they have it on until they encounter a hazard. Now, they can realize it as soon as they load in.

~~The only reason this is a draft pull request is because there is an issue with the Japanese translation where some menu option strings I would like to reuse, e.g. "Glitchrunner mode is {version}", can't be reused on account of them being genericized in Japanese (i.e. they say "Mode: OFF" and that gets reused across multiple strings). Waiting on this to be resolved before merging.~~ Actually instead I'll use the header strings which definitely can't be generic.

Fixes #1091.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Render.cpp
+++ b/desktop_version/src/Render.cpp
@@ -2175,6 +2175,87 @@ static const char* interact_prompt(
     return buffer;
 }
 
+static void mode_indicator_text(const int alpha)
+{
+    const int flags = PR_BRIGHTNESS(alpha) | PR_BOR | PR_RTL_XFLIP;
+    const int r = 220 - help.glow;
+    const int g = 220 - help.glow;
+    const int b = 255 - help.glow/2;
+    const int x = 5;
+    const int spacing = font::height(flags) + 2;
+    int y = 5;
+    if (game.advancetext)
+    {
+        /* Prevent clashing */
+        y += 15;
+    }
+
+    /* FIXME: Some strings have not yet been translated. In order to not have
+     * English text in other languages, they are substituted with existing
+     * ones. Remove all substitute text when they're fully translated. */
+
+    if (map.invincibility)
+    {
+        const char* english = "Invincibility mode enabled";
+        const char* text = loc::gettext(english);
+        if (loc::lang != "en" && SDL_strcmp(english, text) == 0)
+        {
+            /* Substitute text */
+            text = loc::gettext("Invincibility");
+        }
+        font::print(flags, x, y, text, r, g, b);
+        y += spacing;
+    }
+
+    enum GlitchrunnerMode mode = GlitchrunnerMode_get();
+    if (mode != GlitchrunnerNone)
+    {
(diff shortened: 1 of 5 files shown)


desktop_version/src/Render.cpp
```diff
@@ -2158,6 +2158,90 @@ static const char* interact_prompt(
...
+        const char* text = loc::gettext(english);
+        char buffer[SCREEN_WIDTH_CHARS + 1];
+        if (loc::lang != "en" && SDL_strcmp(english, text) == 0)
+        {
+            /* Substitute text */
+            SDL_snprintf(
+                buffer, sizeof(buffer), "%s %s",
+                loc::gettext("Flip Mode"),
+                loc::gettext("Currently ENABLED!")
+            );
```
Daaaav, line 2219: Concatenating translations together in new contexts is a bad idea.

In this case, speaking for Dutch and other languages, starting the word "Currently" with a capital letter, in the middle of the sentence, is painful.

I wouldn't be surprised if you get weird wordings like "Flip mode This is currently ENABLED!" in some languages, and I also wouldn't be surprised if it runs offscreen sometimes. Just "Flip Mode" should be fine - by itself it shows that it's enabled.

Daaaav (changes requested):

desktop_version/src/Render.cpp
```diff
@@ -2158,6 +2158,91 @@ static const char* interact_prompt(
...
+            SDL_strlcpy(buffer, loc::gettext("Glitchrunner Mode"), sizeof(buffer));
+        }
+        else
+        {
+            const char* mode_string = loc::gettext(GlitchrunnerMode_enum_to_string(mode));
+            vformat_buf(
+                buffer, sizeof(buffer),
+                loc::gettext("Glitchrunner mode is {version}"),
+                "version:str", mode_string
+            );
```
Daaaav, line 2211: Is this format string not suppposed to be `text` instead?

Daaaav (changes requested):

desktop_version/src/Render.cpp
InfoTeddy: lol whoops

PR #1099, opened 2024-01-08
Fix another incorrect autotile

AllyTally:
## Changes:

Once again, another tile is incorrect in the autotiling system. This fixes it.

## Legal Stuff:

  for these changes


InfoTeddy (approved):
For future reference, since this wasn't made clear. This fixes an incorrect autotile in the pink Lab tileset.

Incorrect:
incorrect

Fixed:
fixed

We should probably figure out a way to fuzz the autotiler for edge cases like this.

PR #1118, opened 2024-01-14
Properly disable exceptions on MSVC

### LINKED ISSUE #1117: Desktop build for Windows doesn't actually disable exceptions
The CMake code that is supposed to disable exceptions for MSVC build instead enables fully conforming exception handling. The correct approach is to justnot have any `/EH` flags at all, but if you want to be explicit about it you can add `/EHs-` instead.

https://github.com/TerryCavanagh/VVVVVV/blob/8dd5414e8defb6c0bc425acb22c25215012615aa/desktop_version/CMakeLists.txt#L290-L292

InfoTeddy:
Apparently the actual way to do this is to use `-`, not `c`.

Fixes #1117.

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -289,7 +289,7 @@ if(MSVC)
 
     # Disable exceptions
     string(REGEX REPLACE "/EH[a-z]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


InfoTeddy:
Closing due to the fact that this doesn't actually work. See https://github.com/TerryCavanagh/VVVVVV/issues/1117#issuecomment-1904711381.

PR #1122, opened 2024-01-17
Add support for GameCube glyphs

### LINKED ISSUE #1116: Add GameCube glyphs
How on Earth did I forget this one...

Detection is easy since the adapter uses the Nintendo vendor and its own product ID, so I think it just comes down to assets. I can handle the code if anyone is willing to handle asset updates.

flibitijibibo:
It's kind of a bummer that L/R don't actually do anything... we should add ZL/ZR support at some point, my fault for not handling the trigger axes properly.

Also note that GameCube binds X to 'back' rather than B, this will be fixed by SDL_ActionSet for 2.5.

CC @Fussmatte @Daaaav, in case I got the font data wrong!

Fixes #1116


Daaaav:
I can't test with a GameCube controller myself so I can't verify the layout, but the font and code changes look correct to me!

PR #1129, opened 2024-01-22
Add missing Viridian squeaks to cutscene test menu

InfoTeddy:
These mildly annoyed me. It should be consistent that whenever you press ACTION on a menu option, you get a Viridian squeak (at least if something happens).

## Legal Stuff:

  for these changes

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -1338,6 +1338,7 @@ static void menuactionpress(void)
         else if (game.currentmenuoption == (int)game.menuoptions.size()-2)
         {
             // play the cutscene, from clipboard
+            music.playef(Sound_VIRIDIAN);
             game.cutscenetest_menu_play_id = std::string(SDL_GetClipboardText());
             startmode(Start_CUTSCENETEST);
         }
@@ -1351,6 +1352,7 @@ static void menuactionpress(void)
         else
         {
             // play the cutscene!
+            music.playef(Sound_VIRIDIAN);
             game.cutscenetest_menu_play_id = loc::testable_script_ids[(game.cutscenetest_menu_page*14)+game.currentmenuoption];
             startmode(Start_CUTSCENETEST);
         }


Daaaav:
I don't agree - the squeaks are missing on purpose because if you have a lot of cutscenes to test, it could become annoying to hear all the squeaks. So this makes it easier to bear. Also, as a translator menu option, consistency should be less of a concern anyway.

flibitijibibo:
It also makes me think of the earlier textbox change, and I wonder if we should start being more strict about userspace breakages - obviously we'll still have lots of editor changes but minimizing those changes for existing features may be a good idea at this point.

TerryCavanagh:
I guess I'm very mildly with @InfoTeddy on this one, just because it's something I've noticed too. Consistency is probably good. But I usually have the game muted when I'm working on it, so I don't care that much! (sorry, this is not a useful comment)

flibitijibibo:
I guess this would be the determining factor: If the localizers are okay with mashing the built-in mute key that would be enough to merge this, since people dealing with that screen frequently could ignore it if it becomes a problem. I'll +1/-1 whatever they decide in this case!

InfoTeddy:
> the squeaks are missing on purpose because if you have a lot of cutscenes to test, it could become annoying to hear all the squeaks.

I would just like to point out that the cutscenes themselves still have all the squeaks.

> If the localizers are okay with mashing the built-in mute key that would be enough to merge this

Presumably, they are already doing this, since the cutscenes already contain squeaks.

Daaaav:
> I would just like to point out that the cutscenes themselves still have all the squeaks.

Yes, that's kind of my point - there are already so many squeaks in the cutscenes themselves that making every cutscene start with an additional squeak only adds to the pile of squeaks. And I feel you can probably get immersed in the cutscenes better if all squeaks you hear belong _only_ to the cutscenes, and you can clearly hear when a textbox has appeared for you to read?

flibitijibibo:
I guess to reiterate: Totally happy to let the localizers make the final call - if they say no, that's enough to block this change in favor of documenting the history of why this quirk exists.

flibitijibibo:
Going to close this since it's not a _huge_ deal and localization didn't seem to like it. One other thing that came to mind is the fact that cutscene audio could conflict with the menu sound if there was something early enough in the cutscene, so it'd be hard to test audio as a result.

If localization changes their minds we can bring this back.

PR #1130, opened 2024-01-22
PR template: Clarify compensation

InfoTeddy:
Translators have been getting compensated for localization. Sometimes they submit pull requests for their changes, but just because they do doesn't mean that they won't get compensated.

The intent of this line was to make sure that any random contributor wouldn't expect any compensation for their changes, so adding another clause here still keeps that expectation while making it clear that it's not like @TerryCavanagh _never_ compensates people. Even if, as a Swedish man once sung, "Terry is a monster, I think we all know that".

## Legal Stuff:

^^^ This is the updated line!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -34,4 +34,4 @@ By submitting this pull request, I confirm that...
 - [ ] My changes may be used in a future commercial release of VVVVVV
 - [ ] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
   section of the credits for all of said releases, but will NOT be compensated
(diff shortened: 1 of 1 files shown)


TerryCavanagh:
> "... unless there is a prior written agreement" comes to mind, I guess it's implied but as with all agreements being more explicit always helps!

yep, this is a good change!

InfoTeddy:
Suggestion applied.

flibitijibibo (approved):
lgtm - will let Terry push the button since this is licensing-related!

TerryCavanagh:
Done, thanks!

PR #1132, opened 2024-01-22
Properly fix using `setfont()`/`setrtl()` script commands in between text boxes

InfoTeddy:
There used to be a problem with the `setfont()` and `setrtl()` script commands. Namely, if you used them in between text boxes naïvely, without any careful thought, then the fading out text box would suddenly gain the font of the new one. A kludge solution to this was implemented by simply blocking the script until the existing text box faded out before switching the font or RTL, and shipped for 2.4.0.

However, a better solution is to simply bake the font flags in to the text box, so that way, if the level font switches, then the text box keeps its font.

This is only for custom levels, because in the main game, the font in a text box needs to be able to change depending on language. But it seems like custom level translations weren't much on the roadmap, and so even the existing hack didn't support changing the font based on translation (even though translation of custom level cutscenes is supported). So baking the font flags into the text box here doesn't make things any worse.

It also makes things better, arguably, by allowing multiple text boxes to exist on screen at once with different fonts.

Maybe in the future we'll need a flag that specifies that the font should change depending on language if a translation in said language exists for the text box, or something like that.

For people that want to override the fonts of every existing text box on screen, you can specify `all` as the second parameter of `setfont` or `setrtl` to do so.

Here is an example of what you can do now.

2024-01-22_12-15-34_1x

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Font.cpp
+++ b/desktop_version/src/Font.cpp
@@ -95,8 +95,8 @@ uint8_t font_idx_options_n = 0;
 uint8_t font_idx_options[20];
 
 static bool font_level_is_interface = false;
-static bool font_idx_level_is_custom = false;
-static uint8_t font_idx_level = 0;
+bool font_idx_level_is_custom = false;
+uint8_t font_idx_level = 0;
 
 static void codepoint_split(
     const uint32_t codepoint,
@@ -765,13 +765,14 @@ static Font* container_get(FontContainer* container, uint8_t idx)
     return NULL;
 }
 
-static Font* fontsel_to_font(int sel, bool* rtl)
+static Font* fontsel_to_font(int sel, bool* rtl, bool custom)
 {
     /* Take font selection integer (0-31) and turn it into the correct Font
      * 0: PR_FONT_INTERFACE - use interface font
      * 1: PR_FONT_LEVEL     - use level font
      * 2: PR_FONT_8X8       - use 8x8 font no matter what
-     * 3-31:                - use (main) font index 0-28
+     * 3-31:                - use font index 0-28 (depending on custom, from
+     *                        PR_FONT_IDX_IS_CUSTOM)
      *
      * rtl will be set depending on whether we're requesting the interface
      * font (take it from the lang attributes) or level font (take it from
@@ -808,6 +809,10 @@ static Font* fontsel_to_font(int sel, bool* rtl)
         return container_get(&fonts_main, font_idx_8x8);
     }
 
+    if (custom)
+    {
+        return container_get(&fonts_custom, sel-3);
+    }
     return container_get(&fonts_main, sel-3);
 }
 
@@ -816,7 +821,9 @@ static PrintFlags decode_print_flags(uint32_t flags)
 {
     PrintFlags pf;
     pf.scale = FLAG_PART(0, 3) + 1;
-    pf.font_sel = fontsel_to_font(FLAG_PART(3, 5), &pf.rtl);
+    pf.font_sel = fontsel_to_font(
+        FLAG_PART(3, 5), &pf.rtl, flags & PR_FONT_IDX_IS_CUSTOM
+    );
     if (flags & PR_RTL_FORCE)
     {
         pf.rtl = true;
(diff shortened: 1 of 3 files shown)


Daaaav:
This is broken for custom level fonts. Print flags can currently encode individual _main_ fonts (built-in 8x8, Japanese, Chinese, etc...), but not individual fonts from _level assets_, which are stored in a separate `FontContainer` with their own index. When you use `setfont()`, you're changing what `PR_FONT_LEVEL` means, and for fonts that come from level assets, `PR_FONT_LEVEL` is the only way to display them. So without further changes, there can only be one level asset font onscreen at the same time.

Print bitflag

Font selection:
- 0: Use whatever the INTERFACE language/font is
- 1: [PR_FONT_LEVEL] Use whatever the LEVEL language/font is (used for room names, cutscene dialogue, roomtext, etc, but not “Press ENTER to...”)
- 2: [PR_FONT_8X8] Use 8x8 font no matter what
- 3-31: Use specific font (used for language screen, levels list... use function to convert from langcode like "ja" to this number) [PR_FONT_IDX(idx, rtl)]

So if you want to actually make it work this way, you'd have to extend the bitflag with a way to also encode individual level-specific fonts. Something like 32-63 for level-specific fonts 0-31, or a separate bit that just changes the meaning of the existing 0-31.

Also, I think there's actually a nice side effect of the existing system - with internal scripting you could currently make text animate by switching fonts while keeping a textbox displayed! A change like this would make that impossible. Maybe it'd be cool to support both usecases somehow...

desktop_version/src/Font.cpp
```diff
@@ -765,13 +765,13 @@ static Font* container_get(FontContainer* container, uint8_t idx)
...
 
-static Font* fontsel_to_font(int sel, bool* rtl)
+static Font* fontsel_to_font(int sel, bool* rtl, bool custom)
 {
     /* Take font selection integer (0-31) and turn it into the correct Font
      * 0: PR_FONT_INTERFACE - use interface font
      * 1: PR_FONT_LEVEL     - use level font
      * 2: PR_FONT_8X8       - use 8x8 font no matter what
-     * 3-31:                - use (main) font index 0-28
+     * 3-31:                - use font index 0-28
```
Daaaav, line 774: I would update this comment with some more explanation what "font index" now means, namely that it's affected by the `custom` argument (which in turn depends on the newly added flag)

desktop_version/src/Font.h
```diff
@@ -56,6 +56,7 @@
 #define PR_CJK_HIGH (2 << 20) /* larger fonts should stick out fully on the top */
 #define PR_RTL_FORCE (1 << 22) /* force the RTL flag, not needed if the font is set to INTERFACE or LEVEL */
 #define PR_RTL_XFLIP (1 << 23) /* in RTL languages, mirror the X axis, so left is 320 and right is 0, and invert the meaning of PR_LEFT and PR_RIGHT */
+#define PR_FONT_CUSTOM (1 << 24) /* with PR_FONT_IDX, mark that the font index is of a custom font */
```
Daaaav, line 59: This name is a little confusing, since we now have the following under `PR_FONT_*`:
- `PR_FONT_INTERFACE`
- `PR_FONT_LEVEL`
- `PR_FONT_8X8`
- `PR_FONT_IDX(idx, rtl)`
- `PR_FONT_CUSTOM`

It reads like it's just another possible font option. Maybe `PR_FONT_IDX_IS_CUSTOM`?

desktop_version/src/Script.cpp
```diff
@@ -798,6 +798,16 @@ void scriptclass::run(void)
                     graphics.textboxcentery();
                 }
 
+                if (map.custommode)
+                {
+                    int flags = PR_FONT_IDX(font::font_idx_level, cl.rtl);
```
Daaaav, line 803: Print flags are always `uint32_t`, so:

```suggestion
                    uint32_t flags = PR_FONT_IDX(font::font_idx_level, cl.rtl);
```

desktop_version/src/Script.cpp
```diff
@@ -2481,39 +2491,38 @@ void scriptclass::run(void)
...
+                    font::set_level_font(cl.level_font_name.c_str());
+                }
+                else
+                {
+                    font::set_level_font(raw_words[1].c_str());
+                }
+                if (argexists[2] && words[2] == "all")
+                {
+                    /* Immediately update all text boxes. */
+                    int flags = PR_FONT_IDX(font::font_idx_level, cl.rtl);
```
Daaaav, line 2516:
```suggestion
                    uint32_t flags = PR_FONT_IDX(font::font_idx_level, cl.rtl);
```

Daaaav (changes requested):
I think it looks mostly alright from the code side? Just a couple small things, and I haven't tested it out yet.

I was honestly expecting a change like this to be way more invasive 😄

InfoTeddy:
Fixed.

Daaaav (approved):

Issue #1136, opened 2024-01-23
Wiimote support


mardy:
Hi all!
  I want to add support for showing the right button names on the Wii ("1", "2" at least), and I'm planning to go the lazy way: since the only supported controller will be the wiimote, I would just replace the button images in the `buttons_<size>.png` textures with the right images.

What I don't like about this solution is that I wouldn't be able to merge it upstream. So, the question: what's the likelyhood of a PR adding some of the Wiimote buttons being accepted?

InfoTeddy:
I don't see any problems with adding these buttons. The Wiimote can also be used on PC, after all.

Fussmatte:
Here are some buttonsets with the Wii Remote and Nunchuk buttons. Hopefully this will help with the implementation.

![VVVVVVbutton_8squared](https://github.com/TerryCavanagh/VVVVVV/assets/44736084/ce39b24a-af3e-47dd-89aa-2a6b6a961eab)
![VVVVVVbutton_10squared](https://github.com/TerryCavanagh/VVVVVV/assets/44736084/9b642ee0-7be0-4e6d-93cd-927a488be4e6)
![VVVVVVbutton_12squared](https://github.com/TerryCavanagh/VVVVVV/assets/44736084/f0deea30-e719-464b-b077-cb0a432ef300)

mardy:
> Here are some buttonsets with the Wii Remote and Nunchuk buttons. Hopefully this will help with the implementation.

Thanks @Fussmatte, you have been extremely helpful! I will remove the buttons Z and C, because they are located on the nunchuck, and IMHO it doesn't make sense to use the nunchuck with VVVVVV (the Wiimote alone is good enough). I should have a PR ready in the next few days :-)

PR #1141, opened 2024-01-29
Add support for the Wiimote controller

mardy:
Thanks @Fussmatte for the button pictures. The Product IDs added here are those reported by libSDL when running on a Wii.

Fixes: #1136

## Legal Stuff:


flibitijibibo (approved):
Will wait for the the localization team to approve but the technical side lgtm

Daaaav (approved):
Looks good to me as well!

Fussmatte:
On Windows the game isn't picking up on my Wii Remote - I've tried both a native connection and a connection through Dolphin's continuous Wii Remote scanning feature. Is any specific driver required to use the controller in VVVVVV?

InfoTeddy (changes requested):
On the Discord server, we have two reports of the button icons being incorrect. For example, B displays as 1.

mardy:
> On the Discord server, we have two reports of the button icons being incorrect. For example, B displays as 1.

That's because the buttons are reported differently on a PC and on a Wii. I've added a PC-specific layout, can you please ask those players to test this new patch?

flibitijibibo:
This might be worth punting to SDL, we're supposed to have a hidapi backend for Wii at this point...

https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_wii.c

... but if we're not aligned with the _actual_ Wii port of SDL then we might need to sync those configs up.

PR #1160, opened 2024-03-23
Initial touch controls implementation

NyakoFox:
## Changes:

This PR adds touch controls to the game. While not being the prettiest UI-wise, it makes it possible to play the game using touch controls, making testing on mobile devices much easier. This PR will be followed up by other PRs which implement more features, such as:

- Menus converted to a mobile-friendly format
- Additional gameplay control styles (Not just D-pad)
- The touch control tutorial
- Saving when you touch a checkpoint
- Level editor controls

Supersedes #1142.

## Legal Stuff:


NyakoFox:
I've rebased and pushed -- do note that I don't have any apple devices so I'm unable to test for iOS.

desktop_version/CMakeLists.txt
leo60228, line 1: This needs to be changed to include the touch graphics in iOS builds:
```diff
@@ -146,7 +146,7 @@ if(WIN32)
 elseif(ANDROID)
     add_library(VVVVVV SHARED ${VVV_SRC})
 elseif(CMAKE_SYSTEM_NAME STREQUAL "iOS")
-    file(GLOB_RECURSE REPO_RESOURCES "fonts/*" "lang/*")
+    file(GLOB_RECURSE REPO_RESOURCES "fonts/*" "lang/*" "touch/*")
 
     add_executable(VVVVVV MACOSX_BUNDLE ${VVV_SRC} ${DATA_ZIP} AppIcon.xcassets ${REPO_RESOURCES})
     set_target_properties(VVVVVV PROPERTIES
```

leo60228 (changes requested):

Daaaav (changes requested):
Went through this whole PR - it looks pretty solid I think! I mostly have some nits here and there.

In addition to the in-context code comments:
- In `Touch.cpp`, maybe `init(void)` should be at the top? Maybe `get_rect(button, rect)` should be right above `render(void)`? (close to where it's used)
- Maybe the `Touch.cpp` functions could use comments at the top of the functions to document what the functions mean. Like, `touch::get_rect(button, rect)` gets a rect from a `TouchButton`. Then you see `touch::get_scale(void)` and it gets... the scale? From? A button as well? Is it an API for external use? Oh wait, this one doesn't actually take a button, it's the global graphics scale. (Also, if functions aren't used outside of `Touch.cpp` then it helps to make them static to make clear that they're local helper functions and not 'API')
- The names of functions could also make things clearer:
  - The `get_rect(button, rect)` I mentioned above could maybe be clearer as `rect_from_button(button, rect)`
  - `touching_right()` sounds a little bit ambiguous - you're not touching the right arrow button, but the right half of the screen
  - The 3 functions `touch::button_tapped(id)`, `touch::touching_right()`, and `touch::screen_down()` serve similar purposes (to ask whether something is touched) but all are named completely differently

PR #1171, opened 2024-06-04
Load an unlimited amount of loose music

### LINKED ISSUE #965: Custom music from loose files
The ability to load loose music files was added, and VVVVVV now loads all 128 music headers if it can, meaning a lot more room for custom music. But it's not possible to load custom music from loose music files -- you're stuck with the 16 original filenames.

It would be good if there was a way to play custom loose music -- maybe along the lines of `play(my_song)` to play `music/my_song.ogg`, or some sort of filename-to-index map in the level file, so `play(18)` would always play a specific song and wouldn't shift around, etc.

NyakoFox:
## Changes:

Binary blobs can now store 128 tracks, and music can now be read from loose files. Unfortunately, the two cannot be used together -- loose music can still only use 16 tracks. This commit attempts to fix that, allowing for music playback by filename.

Closes #965.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -3,7 +3,6 @@
 
 #include <SDL.h>
 #include <FAudio.h>
-#include <physfsrwops.h>
 
 #include "Alloc.h"
 #include "BinaryBlob.h"
@@ -390,19 +389,6 @@ public:
 
         id = SDL_strdup(_id);
 
-        // Strip "music/" prefix if it exists
-        if (SDL_strncmp(id, "music/", 6) == 0)
-        {
-            SDL_memmove(id, id + 6, SDL_strlen(id) - 5);
-        }
-
-        // Strip file extension if any
-        char* dot = SDL_strrchr(id, '.');
-        if (dot != NULL)
-        {
-            *dot = '\0';
-        }
-
         loose_extra = _loose_extra;
         vorbis = stb_vorbis_open_memory(read_buf, rw->size(rw), &err, NULL);
         if (vorbis == NULL)
@@ -763,6 +749,27 @@ static void add_builtin_sound(const char* id)
     soundTracks.push_back(SoundTrack(asset_filename, id, false));
 }
 
+static void add_builtin_track(SDL_RWops* rw, const char* track_name)
+{
+    // Make an ID from the track name
+    char* id = SDL_strdup(track_name);
+
+    // Strip "music/" prefix if it exists
+    if (SDL_strncmp(id, "music/", 6) == 0)
+    {
+        SDL_memmove(id, id + 6, SDL_strlen(id) - 5);
+    }
+
+    // Strip file extension if any
+    char* dot = SDL_strrchr(id, '.');
+    if (dot != NULL)
+    {
+        *dot = '\0';
+    }
+
+    musicTracks.push_back(MusicTrack(rw, id, false));
+}
+
 void musicclass::init(void)
 {
     if (FAudioCreate(&faudioctx, FAUDIO_1024_QUANTUM, FAUDIO_DEFAULT_PROCESSOR))
@@ -872,28 +879,41 @@ void musicclass::init(void)
 
     if (!mmmmmm_blob.unPackBinary("mmmmmm.vvv"))
     {
+        // If mmmmmm.vvv is invalid, or doesn't exist...
+
+        SDL_RWops* rw;
+
         if (pppppp_blob.unPackBinary("vvvvvvmusic.vvv"))
         {
             vlog_info("Loading music from PPPPPP blob...");
 
             mmmmmm = false;
-            usingmmmmmm=false;
+            usingmmmmmm = false;
 
             int index;
-            SDL_RWops* rw;
 
 #define TRACK_LOAD_BLOB(blob, track_name) \
-    index = blob.getIndex("data/" track_name); \
-    if (index >= 0 && index < blob.max_headers) \
+    vlog_debug("Searching for track " track_name " as loose file"); \
+    rw = FILESYSTEM_loadAssetRWops(track_name); \
+    if (rw != NULL) \
     { \
-        rw = SDL_RWFromConstMem(blob.getAddress(index), blob.getSize(index)); \
-        if (rw == NULL) \
+        vlog_debug("Found loose music file " track_name); \
+        add_builtin_track(rw, track_name); \
+    } \
+    else \
+    { \
+        index = blob.getIndex("data/" track_name); \
+        if (index >= 0 && index < blob.max_headers) \
         { \
-            vlog_error("Unable to read music file header: %s", SDL_GetError()); \
-        } \
-        else \
-        { \
-            musicTracks.push_back(MusicTrack(rw, track_name, false)); \
+            rw = SDL_RWFromConstMem(blob.getAddress(index), blob.getSize(index)); \
+            if (rw == NULL) \
+            { \
+                vlog_error("Unable to read music file header: %s", SDL_GetError()); \
+            } \
+            else \
+            { \
+                add_builtin_track(rw, track_name); \
+            } \
         } \
     }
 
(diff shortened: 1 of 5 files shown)


NyakoFox:
todo: on line 383 of music.cpp, needs a `SDL_strdup(name)` instead of just `name`

desktop_version/src/Music.cpp
```diff
@@ -372,14 +372,16 @@ float SoundTrack::volume = 0.0f;
...
-    MusicTrack(SDL_RWops *rw)
+    MusicTrack(SDL_RWops *rw, const char* name, bool loose_extra)
     {
         SDL_zerop(this);
         read_buf = (Uint8*) SDL_malloc(rw->size(rw));
         SDL_RWread(rw, read_buf, rw->size(rw), 1);
         int err;
         stb_vorbis_info vorbis_info;
         stb_vorbis_comment vorbis_comment;
+        filename = SDL_strdup(name);
```
Daaaav, line 383: `name`, `filename` and `id` are used a bit confusingly. In the enumeration code, the file path (with folder and extension) is first written to a buffer `name`, then a plain version without folder or extension is written to a buffer `id`, this is passed to the `MusicTrack` constructor as `name` and then written to the `filename` attribute, which can be requested with a function called `getid`...

Daaaav (changes requested):
I've only looked at the code thus far and it doesn't seem to do very strange things. I also still wanna test it, and see how it behaves in edge cases (like when songs have numbers as filenames, or how accessing `vvvvvvmusic.vvv` songs by name works or overwriting the same names, and stuff like that). Unless someone wants to beat me to that 😛

desktop_version/src/Music.cpp
NyakoFox: What do you suggest?

Daaaav: Sorry for the late reply!

I'd say:
- `char name[256]` (stores e.g. `music/mysong.ogg`) -> `asset_filename`? It's used as an argument to `FILESYSTEM_loadAssetToMemory()`
- `char id[256]` (stores e.g. `mysong`) -> probably fine as-is
- `char* filename` in `MusicTrack` -> `id` to match
- `const char* name` in the `MusicTrack` constructor -> `id` to match

The names of the constructor argument and class attribute will clash - I'm not too used to C++ OO but I've usually seen the solution in other languages be something like `this.id = id;`. Otherwise, naming the argument `_id` and doing `id = SDL_strdup(_id);` would in my opinion be better than working around it with `filename`/`name`, `loose_extra`/`looseextra`, etc. (And that kind of `_` prefixing is done in other places in the code as well anyway)

desktop_version/src/Music.cpp
```diff
@@ -881,11 +885,53 @@ void musicclass::init(void)
...
+        if (mem == NULL)
+        {
+            vlog_error("Unable to read loose music file: %s", SDL_GetError());
+        }
+        else
+        {
+            rw = SDL_RWFromConstMem(mem, len);
+            if (rw == NULL)
+            {
+                vlog_error("Unable to read loose music file: %s", SDL_GetError());
```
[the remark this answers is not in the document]
Daaaav: Oh, I didn't realize there were actually 3 such messages, hah. (On lines 837, 918 and 925).

I'd just add a little bit of specifics like "Unable to load asset for loose music file", "Unable to read memory for loose music file", etc, just so the messages are uniquely findable. (Alternatively we could always do something like "Unable to read, number 1", "number 2"... but that seems a bit barbaric lol)

flibitijibibo (approved):
Basically the same review as the sounds patch, so consider this approved at the same time as the other.

Daaaav:
Went through this - and an interesting case came up: you can load built-in music (from the binary blob) by filename as well, but currently you have to use the ID `music/4passionforexploring.ogg`.

The standard binary blob uses the following names (that the game looks for for the first 16 songs):

```
data/music/0levelcomplete.ogg
data/music/1pushingonwards.ogg
data/music/2positiveforce.ogg
data/music/3potentialforanything.ogg
data/music/4passionforexploring.ogg
data/music/5intermission.ogg
data/music/6presentingvvvvvv.ogg
data/music/7gamecomplete.ogg
data/music/8predestinedfate.ogg
data/music/9positiveforcereversed.ogg
data/music/10popularpotpourri.ogg
data/music/11pipedream.ogg
data/music/12pressurecooker.ogg
data/music/13pacedenergy.ogg
data/music/14piercingthesky.ogg
data/music/predestinedfatefinallevel.ogg
```

For boring macro-related reasons (mostly because this was never accounted for), the ID for those built-in songs only has `data/` removed from them, not `music/` and `.ogg`.

You _can_ actually set the filename attribute for extra songs in the binary blob, and use that in a scripting command to load that song. It would probably be best to strip both `data/music/` and `.ogg` if present from the filename for the ID, so you can use `play(5intermission)` if you like, and also add your own files with custom names (without having to have `data/music/` prefixes or `.ogg` suffixes, even though they'd be allowed)

NyakoFox:
I have now made it possible to play built-in songs with the internal filename, like `play(5intermission)`; and also added a warning like the sounds PR also got. Filenames are also converted to IDs the same way that the extra sounds are.

While working on this PR, I did notice that loose music files don't work as I expected; They are only loaded if `vvvvvvmusic.vvv` is invalid or missing (which is what the original PR DID say) and they also are not read from custom levels. I might make a followup PR which allows them to be level assets as expected.

flibitijibibo (approved):
Makes sense to me, but will wait for a +1 just in case the remaining case @NyakoFox pointed out is important for this feature.

Daaaav (approved):
It took me a bit to understand what exactly the remaining problem case was, but essentially: loose files in `music/` cannot override songs that are also defined in a `vvvvvvmusic.vvv`. You _can_ have custom-named songs in `vvvvvvmusic.vvv` _and_ have custom-named songs as loose files in `music/`, and both work, except for the ones that have the same name across both, then the one from `vvvvvvmusic.vvv` takes priority, and it's probably ideal for that to be the other way around.

For an overview, I tested different combinations of locations where a song can be stored, and scripting commands to access these songs:

Image


Testing levels: [pr1171-tests.zip](https://github.com/user-attachments/files/27418148/pr1171-tests.zip)

(Testing level 3 also has three different files where I tested variants of including a custom song filename in a binary blob: `horses`, `horses.ogg` and `music/horses.ogg`.)

I also tested `data/music/horses.ogg` in the binary blob, which does not work even though it follows the same pattern as the other songs, but I don't think it's a problem. We can just store IDs without `data/`.

So overall: +1

mothbeanie:
for the record, the original case that ally mentioned is a result of #412 from 5 years ago - it only comes into play if `vvvvvvmusic.vvv` is missing from data.zip, e.g. the asset is malformed or intentionally removed by a user to reduce heap usage.
a custom `vvvvvvmusic.vvv` overriding a custom loose ogg file with the same name, like in dav's graph, is actually a second *additional* case. i agree that in both cases the behavior is unexpected - loose music files should probably always take precedence over `vvvvvvmusic.vvv` regardless of what's in it or if it even exists - the only way they'll be loaded in the first place is if they were intentionally placed by a user, as the game normally ships with just the binary blob `vvvvvvmusic.vvv` in data.zip.

NyakoFox:
I was in the middle of writing something but I was distracted by @Daaaav -- I'm fixing both the issue mentioned and the original issue from PR #412 as both need to do some similar fixes and change some fundemental logic issues.

NyakoFox:
I've pushed a pretty sizeable commit which addresses the initial bug I was talking about (case 4 in Dav's image I think); Dav's case 6 is not something you'll actually encounter and I've gone into more detail in the commit description. Personally, I think it's good to go after another review!

desktop_version/src/Music.cpp
```diff
@@ -743,6 +749,27 @@ static void add_builtin_sound(const char* id)
...
+    }
+
+    // Strip file extension if any
+    char* dot = SDL_strrchr(id, '.');
+    if (dot != NULL)
+    {
+        *dot = '\0';
+    }
+
+    musicTracks.push_back(MusicTrack(rw, id, false));
```
Daaaav, line 807: `MusicTrack` strdup's the `id` itself, so it's not used after this point and needs to be free'd.

```suggestion
    musicTracks.push_back(MusicTrack(rw, id, false));
    VVV_free(id);
```

desktop_version/src/Music.cpp
```diff
@@ -945,11 +987,82 @@ void musicclass::init(void)
...
+
+        // Null-terminate the string
+        id[current_char] = '\0';
+
+        // Chop off the extension!
+        char* dot = SDL_strrchr(id, '.');
+        if (dot != NULL)
+        {
+            *dot = '\0';
+        }
```
Daaaav: Maybe this could be a function - lines 1004-1033 look the same as lines 825-854 for sound effects.

Daaaav (changes requested):
Two quick code things but other than that I think this PR is pretty much ready!

Like you said, case 6 remains, but that case only happens when you did it to yourself (adding `horses` to both music/ and vvvvvvmusic.vvv)

New table:

Image

desktop_version/src/Music.cpp
NyakoFox: Oh, shoot -- apparently I half-reverted this code; add_builtin_song isn't supposed to `strdup` at all but instead make a temporary buffer...

desktop_version/src/Music.cpp
NyakoFox: Made it a function.

NyakoFox:
Solved the issues mentioned!

desktop_version/src/Music.cpp
```diff
@@ -742,6 +742,42 @@ musicclass::musicclass(void)
     usingmmmmmm = false;
 }
 
+static void make_id_from_filename(char* id, size_t id_size, const char* filename)
```
Daaaav, line 745: `id_size` is not used (looks like it's meant to replace the 255 below)

Daaaav (changes requested):

desktop_version/src/Music.cpp
NyakoFox: Whoops, good catch!

Daaaav (approved):
Looks good to me now!

flibitijibibo (approved):

PR #1172, opened 2024-06-04
Touch swipe movement

NyakoFox:
## Changes:

The current mobile version of VVVVVV has three movement types: swipe, d-pad and sides. Swipe is the default, but my PRs have only implemented d-pad. This commit adds swipe and adds it as the default.

Depends on #1164.

## Legal Stuff:


NyakoFox:
Apparently this has a few bugs, so I'm leaving this as a draft for now.

PR #1174, opened 2024-06-05
Checkpoint autosaving

NyakoFox:
## Changes:

This is a small PR that adds mobile VVVVVV's behavior of saving on checkpoints. It quicksaves whenever your checkpoint gets set, either by touching a checkpoint or `setcheckpoint()`. On by default on Android, off otherwise. Can be manually overridden through `settings.vvv`.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2733,17 +2733,7 @@ bool entityclass::updateentities( int i )
                 }
                 entities[i].state = 0;
 
-                if (game.checkpoint_saving)
-                {
-                    bool success = game.savequick();
-                    game.gamesaved = success;
-                    game.gamesavefailed = !success;
-
-                    if (game.gamesavefailed) {
-                        game.show_save_fail();
-                        graphics.textboxapplyposition();
-                    }
-                }
+                game.checkpoint_save();
             }
             break;
         case 9: //Gravity Lines
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2732,6 +2732,13 @@ bool entityclass::updateentities( int i )
                     game.savedir = entities[player].dir;
                 }
                 entities[i].state = 0;
+
+                if (game.checkpoint_saving)
+                {
+                    bool success = game.savequick();
+                    game.gamesaved = success;
+                    game.gamesavefailed = !success;
+                }
             }
             break;
         case 9: //Gravity Lines
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -2738,6 +2738,11 @@ bool entityclass::updateentities( int i )
                     bool success = game.savequick();
                     game.gamesaved = success;
                     game.gamesavefailed = !success;
+
+                    if (game.gamesavefailed) {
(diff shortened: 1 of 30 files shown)


InfoTeddy (commented):
If the game fails to quicksave, there is no indication to the user (besides console output) that it has failed.

Also, this adds a new settings bool but there is no way to toggle it without manually editing the file. Should we add a menu option to toggle it?

Also (and this doesn't have to be done right now), I was wondering if we should add some sort of indication for a successful quicksave. Nothing as big as the teleporter "Game Saved" but something like a rotating icon in the corner that other games do would be good. (Of course it would be slightly misleading because we block the entire game thread for the write but it would still get the point across.)

NyakoFox:
> Also, this adds a new settings bool but there is no way to toggle it without manually editing the file. Should we add a menu option to toggle it?
Honestly, it feels like something that you should manually have to toggle if you know what you're doing -- we don't want people on Android to lose their progress. However, if people disagree, I can add a menu option.

InfoTeddy:
> Honestly, it feels like something that you should manually have to toggle if you know what you're doing -- we don't want people on Android to lose their progress. However, if people disagree, I can add a menu option. 

A menu option _would_ still be manually toggling it, and Android players would know fully well what the option does based off the name and description. If they decide to toggle it, that's on them.

desktop_version/src/Entity.cpp
```diff
@@ -2732,6 +2732,18 @@ bool entityclass::updateentities( int i )
...
                 }
                 entities[i].state = 0;
+
+                if (game.checkpoint_saving)
+                {
+                    bool success = game.savequick();
+                    game.gamesaved = success;
+                    game.gamesavefailed = !success;
+
+                    if (game.gamesavefailed) {
```
InfoTeddy, line 2742:
```suggestion
                    if (game.gamesavefailed)
                    {
```

desktop_version/src/Script.cpp
```diff
@@ -1206,6 +1206,18 @@ void scriptclass::run(void)
...
                     game.savedir = obj.entities[i].dir;
                 }
+
+                if (game.checkpoint_saving)
+                {
+                    bool success = game.savequick();
+                    game.gamesaved = success;
+                    game.gamesavefailed = !success;
+
+                    if (game.gamesavefailed) {
```
InfoTeddy, line 1216:
```suggestion
                    if (game.gamesavefailed)
                    {
```

InfoTeddy (changes requested):
Attempts to quicksave are made in Time Trials, Intermission replays, No Death Mode, and custom levels, but they fail, so the error text box displays. (The Secret Lab doesn't have checkpoints so in practice it couldn't happen, but this issue also applies there.)

In particular we might want this feature to work for custom level quicksaves too. :stuck_out_tongue: Anyone know what the current mobile version behavior is for custom levels? I would hope it automatically writes custom level quicksaves too.

Daaaav:
> 😛 Anyone know what the current mobile version behavior is for custom levels? I would hope it automatically writes custom level quicksaves too.

Yep, custom levels also autosave on mobile.

InfoTeddy (approved):

PR #1177, opened 2024-06-09
Fix regression: Entities not moving

### LINKED ISSUE #1176: Regression: moving enemies aren't moving anymore
Just posting this here so it can't get missed/forgotten:

https://github.com/TerryCavanagh/VVVVVV/assets/44736680/14426589-4d58-4435-888e-e33a14b145cf

(reported by Klank in the VVVVVV Discord)

This is caused by 53d725f78a3550b750d652a62ffe9d8b51dd0e18.

InfoTeddy:
Commit 53d725f78a3550b750d652a62ffe9d8b51dd0e18, intended to fix an overzealous commit, was itself overzealous. This is because it applied to all entities when it should only apply to entity-emitting entities. To fix this, `entityclonefix` needs to no-op if the entity is not an entity emitter.

Fixes #1176.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Entity.cpp
+++ b/desktop_version/src/Entity.cpp
@@ -1235,6 +1235,12 @@ static void entityclonefix(entclass* entity)
     const bool is_lies_emitter = entity->behave == 10;
     const bool is_factory_emitter = entity->behave == 12;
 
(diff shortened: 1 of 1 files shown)


Daaaav:
From some playtesting in the main game and custom levels, this seems to be a non-overzealous fix.

PR #1186, opened 2024-07-13
Use variable timestep for `SDL_Delay`

### LINKED ISSUE #1185: Fast-forward does not work in 30-FPS-only mode
The level debugger's fast-forward functionality (accessible by holding down F with the debug menu active) only works in over-30-FPS mode. It does not work if the framerate mode is set to 30 FPS only.

<sub>Thanks to AllyTally for originally reporting this on Discord.</sub>

InfoTeddy:
This fixes a bug where fast-forward wouldn't work in 30-FPS-only mode.

This is because the 30-FPS-only code has a hardcoded check for the number 34, as in 34 milliseconds must pass before the next frame can advance. This is why slowdown still worked, because slowdown means you're waiting longer than 34 ms anyways, but fast-forward tries to wait for only 1 ms, which wouldn't work if the 34 limit was still enforced.

So instead, swap out the 34 with `game.get_timestep()` and this will be fixed.

Fixes #1185.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -866,9 +866,10 @@ int main(int argc, char *argv[])
         f_time = SDL_GetTicks64();
 
         const Uint64 f_timetaken = f_time - f_timePrev;
-        if (!game.over30mode && f_timetaken < 34)
+        const int timestep = game.get_timestep();
+        if (!game.over30mode && f_timetaken < (Uint64) timestep)
         {
-            const volatile Uint64 f_delay = 34 - f_timetaken;
+            const volatile Uint64 f_delay = timestep - f_timetaken;
             SDL_Delay((Uint32) f_delay);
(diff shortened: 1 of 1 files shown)


desktop_version/src/main.cpp
```diff
@@ -863,12 +863,14 @@ int main(int argc, char *argv[])
 #else
     while (true)
     {
+        const int timestep = game.get_timestep();
+
         f_time = SDL_GetTicks64();
 
         const Uint64 f_timetaken = f_time - f_timePrev;
```
Daaaav, line 870: Nit:
```suggestion
        f_time = SDL_GetTicks64();

        const Uint64 f_timetaken = f_time - f_timePrev;
        const int timestep = game.get_timestep();
```

NyakoFox:
lgtm

PR #1192, opened 2024-08-30
Give Master of the Universe trophy in normal mode, meeting certain conditions

InfoTeddy:
This makes it so that it is possible to obtain the Master of the Universe trophy/achievement, usually unlocked by beating No Death Mode, outside of NDM.

There are several conditions that need to be met:

1. The game needs to be started from a new game and cannot be from loading a save.
2. Accessibility modes (invincibility and slowdown) must never be enabled.

If either condition is violated, then the boolean that keeps track of NDM eligibility will be set to false.

## Legal Stuff:


TerryCavanagh:
Oh! This is a good idea!

PR #1195, opened 2024-09-16
Remove slowdown during death animation

InfoTeddy:
Some discussion on the Discord server resulted in this change. It's a quality-of-life improvement where, if the game is in slowdown mode, it will return to 100% speed for the duration of the death animation.

The reasoning is obvious. There is nothing to do during the death animation, so making it take longer during slowdown is just an annoyance to the player, almost a penalty for them using an accessibility option. This is the same reason why slowdown no longer applies in menus, etc.

## Legal Stuff:


Daaaav (approved):
Looks good to me. Slowdown certainly feels much better this way.

PR #1196, opened 2024-09-16
Add glitchless mode

InfoTeddy:
This adds a new mode (mutually incompatible with glitchrunner mode) that disables glitches that would otherwise be used in speedrunning. The intention is to facilitate speedrunning glitchless categories without accidentally triggering a glitch which would invalidate the run (e.g. line clipping).

## Legal Stuff:

### DIFF
--- a/desktop_version/lang/ar/strings.xml
+++ b/desktop_version/lang/ar/strings.xml
@@ -249,6 +249,14 @@
     <string english="none" translation="بدون" explanation="menu option, do not emulate any older version"/>
     <string english="2.0" translation="2.0" explanation="VVVVVV version number for glitchrunner mode"/>
     <string english="2.2" translation="2.2" explanation="VVVVVV version number for glitchrunner mode"/>
+    <string english="Glitchrunner mode is incompatible with glitchless mode." translation="" explanation="" max="38*2" max_local="38*2"/>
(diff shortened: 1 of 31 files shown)


Buggem:
Are you going to fix the issue of getting stuck in walls (on custom levels, travelling into the next room will get you stuck in the wall if your entrance to the next room is blocked.)

Daaaav:
> Are you going to fix the issue of getting stuck in walls (on custom levels, travelling into the next room will get you stuck in the wall if your entrance to the next room is blocked.)

That's not really what glitchless mode was intended to solve - wall sliding is more of a failsafe, and not a glitch.

Buggem:
> That's not really what glitchless mode was intended to solve - wall sliding is more of a failsafe, and not a glitch.

A real "failsafe" would be simply not letting the player enter the room. (e.g detecting if the tile in the next room is wall)

Buggem:
Unless the official VVVVVV levelset has used this failsafe, ofc.

NyakoFox:
> A real "failsafe" would be simply not letting the player enter the room.

Glitchless mode is intended for speedrunners, as the PR description states. It should not have anything to do with custom levels. Additionally, your "real failsafe" would break levels which already uses this mechanic, which matters even if it's not a built-in level.

Issue #1200, opened 2024-10-04
Update to SDL3


flibitijibibo:
The ABI has officially frozen, so this should be safe to do - if we want to keep 2.0 support until the final launch it should be easy enough to do a BUILD_SDL3 option like I did for FNA3D and FAudio.

NyakoFox:
Could this wait until all of my touch input PRs get merged? It'd probably be easier to update all in one go instead of having to update the current codebase, *then* all affected PRs.

NyakoFox:
When this happens (might be in a while due to touch input not being merged yet), we might be able to get rid of the lodepng dependency, since SDL 3.3.2 seems to add `SDL_LoadPNG()` and `SDL_LoadPNG_IO()`.  We'd have to be very careful about things like the values of the RGB channels while fully transparent, though, due to the red channel bug -- doesn't sound that hard to fix though. I would personally prefer if we got rid of lodepng for SDL 3.3.2 later on (one extra dependency means another git submodule to deal with...)

flibitijibibo:
Also up for doing this - the loader in SDL is just stbi so that'd be the only difference:

https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_stb.c#L203

flibitijibibo:
This is now live on 2.5-testing for Linux and Windows.

PR #1204, opened 2024-10-12
Add libretro port

iyzsong:
## Changes:

Add libretro port, which build VVVVVV as a shared library (static link not tested) for RetroArch.
It:
- Add SDL2 as a submodule to `third_party`
- Use SDL software renderer
- Add `BUILD_LIBRETO` and `LIBRETO_STATIC` options to cmake
- Will build `vvvvvv_libretro.so` on Linux, run via `retroarch -L vvvvvv_libretro.so`.
- game data (`data.zip`, `lang` and `fonts`) should be in a `vvvvvv` directory in the RetroArch system directory.

Thanks!

## Legal Stuff:


NyakoFox:
How does this work with custom levels with custom assets? Isn't there a problem regarding save state size being dynamic due to custom assets size being dynamic?

iyzsong:
> How does this work with custom levels with custom assets? Isn't there a problem regarding save state size being dynamic due to custom assets size being dynamic?

I haven't test it, but it can load `data.zip` as game, and use its directory to replace the system directory.  The saves are not managed  by RetroArch, and will be put in the system directory's `saves`.

InfoTeddy:
> How does this work with custom levels with custom assets? Isn't there a problem regarding save state size being dynamic due to custom assets size being dynamic?

This doesn't support save states (`retro_serialize` and `retro_unserialize` return `false`).

flibitijibibo:
The changes to the core are okay, but the added SDL files are kind of huge - we did a bunch of changes to SDL's overall design to accommodate RetroArch's expectations for program flow, has a libretro port of SDL3 been attempted yet? The custom files actually look okay so I wonder if now's the time to upstream, and then we can get this port in when #1200 is done.

PR #1207, opened 2024-10-28
CI: Migrate to SLR Sniper SDK for Linux builds

### LINKED ISSUE #1205: CI: Update Linux image to Sniper SDK
We're investigating this for FNA's libraries, should be basically the same process for VVVVVV:

https://github.com/FNA-XNA/FAudio/pull/359

flibitijibibo:
This isn't a _huge_ change, but it does bump the glibc requirement to 2.31, so this shouldn't be backported to 2.4-updates. The main benefit is that it gives us a container with access to new SDL APIs while also giving Misa one less thing to maintain locally.

Fixes #1205

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,41 +64,36 @@ jobs:
       run: ninja -C ${SRC_DIR_PATH}/build
 
   build-lin:
-    name: Build (CentOS 7)
+    name: Build (Steam Linux Runtime Sniper)
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


leo60228:
Is it possible this could cause problems for people playing itch.io/DRM-free builds on LTS distros? To be fair, I'm not sure how many people are playing games on 4-year-old distros anyway.

flibitijibibo:
I think the only mainstream distribution affected is Ubuntu 20.04 LTS, which ends mainline support next year. I was planning to bump flibitBuild to glibc 2.34 after SDL3 released, so this should be even safer by the time we start shipping with it.

PR #1208, opened 2024-11-03
Add `position(force)` and `textoutline(on/off/default)`

NyakoFox:
## Changes:

The `force` argument in `position`  forces the textbox position, meaning it won't be moved to be inside of the bounds of the screen (nor have the 10 pixel padding on each side.)

`textoutline(on/off/default)` is a new scripting command for textbox visual control, where you're able to force the transparent textbox's outline on or off. This does NOT affect roomtext.

These changes are both solely visual.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -3214,15 +3214,15 @@ SDL_Color Graphics::bigchunkygetcol(int t)
     return color;
 }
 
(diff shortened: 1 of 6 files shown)


mothbeanie:
perhaps `position(unbounded)` might be a more descriptive name, instead of `position(force)`?

NyakoFox:
Changed to `position(absolute)`.

PR #955, opened 2023-03-22
Refactor script text input

AllyTally:
## Changes

This PR adds an input system with selection, copying, pasting, and all of that fun stuff, to the script editor.

NOTE: I'm still new to text input, so if someone else sees issues then feel free to contribute.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV
- [x] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes

PR #1213, opened 2024-11-22
[Unfinished] Some patches 

Buggem:
## Changes:

So far:

  - Set FS permissions from 777 (bad, because it gives rw permissions to every person, even low level users) to 755, the better permission
  - Added text L+R scrolling to the Script Editor

...and more that I am planning to add in the future:
  - Multiplayer(?)
  - Looking around for bugs
  - OmniScale (some code from [SameBoy](https://github.com/LIJI32/SameBoy/blob/master/Shaders/AAOmniScaleLegacy.fsh), the license is pretty safe, only if there is not a substantial amount of code from the project)

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Constants.h
+++ b/desktop_version/src/Constants.h
@@ -6,6 +6,7 @@
  * FIXME: This should include more constants!
  */
 
+#define NO_DATA_ZIP "Error: data.zip missing!\nYou do not have data.zip!\nGrab it from your purchased copy of the game, or get it from the free Make and Play Edition.\n\nIf you are getting this error, and you do\n have data.zip, move it into the folder\ncontaining the executable.\n");
 #define SCREEN_WIDTH_TILES 40
 #define SCREEN_HEIGHT_TILES 30
 
--- a/desktop_version/src/Constants.h
+++ b/desktop_version/src/Constants.h
@@ -6,7 +6,8 @@
  * FIXME: This should include more constants!
  */
 
-#define NO_DATA_ZIP "Error: data.zip missing!\nYou do not have data.zip!\nGrab it from your purchased copy of the game, or get it from the free Make and Play Edition.\n\nIf you are getting this error, and you do\n have data.zip, move it into the folder\ncontaining the executable.\n");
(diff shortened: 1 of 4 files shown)


NyakoFox:
Your changes should probably be in separate PRs.

> Added text L+R scrolling to the Script Editor

I wrote #955 which has various text input improvements, including moving the cursor left/right

> Multiplayer(?)

That's uh, quite an ambitious addition, maybe some discussion would be best first.

> OmniScale (some code from [SameBoy](https://github.com/LIJI32/SameBoy/blob/master/Shaders/AAOmniScaleLegacy.fsh), the license is pretty safe, only if there is not a substantial amount of code from the project)

In my opinion, upscaler shaders have never looked good, and OmniScale doesn't seem like an exception... personal opinions aside, I don't think shaders make sense in the base repository, especially since we don't use shaders anywhere else yet.

Daaaav:
Hi! So to add a few more things:

As Nyako said, there are multiple different changes in this PR, and it's better to make one PR per (related) change. Additionally, you may want to squash/fixup the commits a bit - the 1st commit adds a constant that's only defined in the 2nd one, which has a syntax error which is resolved in the 4th one, etc. Ideally, every individual commit can be compiled, and has a message describing the change (instead of "Some polish"). Don't make the first line of the commit message longer than 72 characters though. Or any further lines technically, but especially the first 😛 

> 777 (bad, because it gives rw permissions to every person, even low level users) to 755, the better permission

777 is often seen as a "dangerous" permissions value because it gives permission to **everyone** to do **everything**, but that's not really the case. That low-level user also needs access to _every directory along the way_, including your home directory. If your entire home directory is already open for every user to access, restricting access to your VVVVVV directory is the least of your worries.

Buggem:
@Daaaav 
I thought it would create a lot of clog to open too many pull requests, and at the time of adding the individual commits, I didn't have access to `git` and was using the GitHub UI.

flibitijibibo:
These can be reviewed separately once they've been cleaned up and finished - nothing in here is _horrible_ but it's nicer to get emails for PRs that are finished (or at least a complete draft).

PR #1217, opened 2024-12-10
Add Discord Game SDK support

Buggem:
## Changes:

- Adds a Discord RPC implementation

Unfinished, I need to link the Discord library for more platforms in the CMakeLists.txt.
Also, linking it at runtime (even if it doesn't get included in the executable also is a requirement unless you want a clunky workaround.

I would **love** to hear your thoughts on this change, and any improvements or suggestions are more than welcome.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/DiscordNetwork.c
+++ b/desktop_version/src/DiscordNetwork.c
@@ -1,0 +1,85 @@
+#include "MakeAndPlay.h"
+
+#ifndef MAKEANDPLAY
+
+#include <stdint.h>
+#include <stdio.h>
+#include <assert.h>
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <unistd.h>
+#include <string.h>
+#endif
+
+// Library includes
+#include <SDL.h>
+#include "discord_game_sdk.h" // DISCORD HEADERS REQUIRE LINKING AT RUNTIME!!!!!
+
+#define DISCORD_REQUIRE(x) assert(x == DiscordResult_Ok)
+
+struct Application {
+    struct IDiscordCore* core;
+    struct IDiscordActivityManager* activityMan;
+};
+
+struct DiscordActivity activity;
+struct Application app;
+
+
+
+int32_t DISCORD_init(void)
+{
+    memset(&app, 0, sizeof(app));
+
+    struct DiscordCreateParams params;
+    params.client_id = 1315544357532729447;
+    params.flags = DiscordCreateFlags_Default;
+
+    if(!DiscordCreate(DISCORD_VERSION, &params, &app.core))
+    {
+        return 1;
+    }
+
+
+    strncpy(activity.state, "On the Main Menu", 128);
+    //strncpy(activity.details, "I'm Sorry", 128);
+    strncpy(activity.assets.large_image, "vvvvvv", 128);
+    strncpy(activity.assets.large_text, "Dimension  VVVVVV", 128);
+
+    app.activityMan = app.core->get_activity_manager(app.core);
+    app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
+
+
+
+    return 0;
+}
+
+void DISCORD_update(const char *level, const char *name)
+{
+    if(app.activityMan == NULL)
+    {
+        app.activityMan = app.core->get_activity_manager(app.core);
+    }
+    strncpy(activity.state, name, 127);
+    strncpy(activity.assets.large_image, "vvvvvv", 127);
+    strncpy(activity.assets.large_text, level, 127);
+    activity.state[127] = '\0';
+    activity.assets.large_image[127] = '\0';
+    activity.assets.large_text[127] = '\0';
+
+    app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
+
+    DISCORD_REQUIRE(app.core->run_callbacks(app.core));
+}
+void DISCORD_unlockAchievement(const char *name)
+{
+    // No "achivements" in Discord
+}
+
+void DISCORD_shutdown(void)
+{
+    // ???
+}
+
+#endif // MakeAndPlay
--- a/desktop_version/src/Network.c
+++ b/desktop_version/src/Network.c
@@ -1,4 +1,5 @@
 #include <stdint.h>
+#include <stdio.h>
 
 #include "MakeAndPlay.h"
 #include "Unused.h"
@@ -9,6 +10,9 @@
     #endif
     #ifdef GOG_NETWORK
         #undef GOG_NETWORK
+    #endif
+    #ifdef DISCORD_NETWORK
+        #undef DISCORD_NETWORK
     #endif
 #endif
 
@@ -22,18 +26,26 @@
 #else
 #define GOG_NUM 0
 #endif
+#ifdef DISCORD_NETWORK
+#define DISCORD_NUM 1
+#else
+#define DISCORD_NUM 0
+#endif
 
-#define NUM_BACKENDS (STEAM_NUM+GOG_NUM)
+#define NUM_BACKENDS (STEAM_NUM+GOG_NUM+DISCORD_NUM)
 #define DECLARE_BACKEND(name) \
     int32_t name##_init(void); \
     void name##_shutdown(void); \
-    void name##_update(void); \
+    void name##_update(const char *level, const char *name); \
     void name##_unlockAchievement(const char *name);
 #ifdef STEAM_NETWORK
 DECLARE_BACKEND(STEAM)
 #endif
 #ifdef GOG_NETWORK
 DECLARE_BACKEND(GOG)
+#endif
+#ifdef DISCORD_NETWORK
+DECLARE_BACKEND(DISCORD)
 #endif
 #undef DECLARE_BACKEND
 
@@ -42,7 +54,7 @@ typedef struct NetworkBackend
     int32_t IsInit;
     int32_t (*Init)(void);
     void (*Shutdown)(void);
-    void (*Update)(void);
+    void (*Update)(const char*, const char*);
     void (*UnlockAchievement)(const char*);
 } NetworkBackend;
 
@@ -63,6 +75,9 @@ int NETWORK_init(void)
     #endif
     #ifdef GOG_NETWORK
     ASSIGN_BACKEND(GOG, STEAM_NUM)
+    #endif
+    #ifdef DISCORD_NETWORK
+    ASSIGN_BACKEND(DISCORD, STEAM_NUM+GOG_NUM)
     #endif
     #undef ASSIGN_BACKEND
     #if NUM_BACKENDS > 0
@@ -89,14 +104,14 @@ void NETWORK_shutdown(void)
     #endif
 }
 
-void NETWORK_update(void)
+void NETWORK_update(const char *level, const char *name)
 {
     #if NUM_BACKENDS > 0
     int32_t i;
     for (i = 0; i < NUM_BACKENDS; i += 1)
     if (backends[i].IsInit)
     {
-        backends[i].Update();
+        backends[i].Update(level, name);
     }
     #endif
 }
--- a/desktop_version/src/main.cpp
+++ b/desktop_version/src/main.cpp
@@ -979,7 +979,7 @@ static enum LoopCode loop_begin(void)
     }
 
     // Update network per frame.
-    NETWORK_update();
+    NETWORK_update(map.currentarea(game.roomx, game.roomy), loc::gettext_roomname(map.custommode, game.roomx, game.roomy, map.roomname, map.roomname_special));
 
     return Loop_continue;
 }
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -11,6 +11,7 @@ option(BUNDLE_DEPENDENCIES "Use bundled TinyXML-2, PhysicsFS, and FAudio (if dis
 
 option(STEAM "Use the Steam API" OFF)
 option(GOG "Use the GOG API" OFF)
+option(DISCORD "Use the Discord API" OFF)
 
 option(OFFICIAL_BUILD "Compile an official build of the game" OFF)
 
@@ -19,6 +20,7 @@ option(MAKEANDPLAY "Compile a version of the game without the main campaign (pro
 if(OFFICIAL_BUILD AND NOT MAKEANDPLAY)
     set(STEAM ON)
     set(GOG ON)
+    set(DISCORD ON)
 endif()
 
 option(REMOVE_ABSOLUTE_PATHS "If supported by the compiler, replace all absolute paths to source directories compiled into the binary (if any) with relative paths" ON)
@@ -133,6 +135,9 @@ endif()
 if(GOG)
     list(APPEND VVV_C_SRC src/GOGNetwork.c)
 endif()
+if(DISCORD)
+    list(APPEND VVV_C_SRC src/DiscordNetwork.c)
+endif()
 if(CMAKE_SYSTEM_NAME STREQUAL "iOS")
     list(APPEND VVV_C_SRC src/SDL_uikit_main.c)
 endif()
@@ -209,6 +214,9 @@ if(STEAM)
 endif()
 if(GOG)
     target_compile_definitions(VVVVVV PRIVATE -DGOG_NETWORK)
+endif()
+if(DISCORD)
+    target_compile_definitions(VVVVVV PRIVATE -DDISCORD_NETWORK)
 endif()
 
 set(XML2_SRC
@@ -352,6 +360,10 @@ add_library(c-hashmap-static STATIC ${CHM_SRC})
 
 add_library(sheenbidi-static STATIC ${SBIDI_SRC})
 
+# Link after we did all of the other initialisation
+if(DISCORD)
+    target_link_options(VVVVVV PRIVATE ../../third_party/libdiscord_game_sdk.so)
+endif()
 target_compile_definitions(sheenbidi-static PRIVATE
     -DSB_CONFIG_UNITY
 )
--- a/desktop_version/src/GOGNetwork.c
+++ b/desktop_version/src/GOGNetwork.c
@@ -15,7 +15,7 @@ void GOG_shutdown(void)
 {
 }
 
-void GOG_update(void)
+void GOG_update(const char *level, const char *name)
 {
 }
 
--- a/desktop_version/src/Network.h
+++ b/desktop_version/src/Network.h
@@ -11,7 +11,7 @@ int32_t NETWORK_init(void);
 
 void NETWORK_shutdown(void);
 
-void NETWORK_update(void);
+void NETWORK_update(const char *level, const char *name);
 
 void NETWORK_unlockAchievement(const char *name);
 
--- a/desktop_version/src/SteamNetwork.c
+++ b/desktop_version/src/SteamNetwork.c
@@ -234,7 +234,7 @@ void STEAM_shutdown(void)
     }
 }
 
-void STEAM_update(void)
+void STEAM_update(const char *level, const char *name)
 {
     if (!libHandle)
     {
--- a/desktop_version/src/discord_game_sdk.h
+++ b/desktop_version/src/discord_game_sdk.h
@@ -1,0 +1,729 @@
+#ifndef _DISCORD_GAME_SDK_H_
+#define _DISCORD_GAME_SDK_H_
+
+#ifdef _WIN32
+#include <Windows.h>
+#include <dxgi.h>
+#endif
+
+#ifdef _WIN32
+#  ifdef _WIN64
+#    define DISCORD_API
+#  else
+#    define DISCORD_API __stdcall
+#  endif
+#else
+#  define DISCORD_API
+#endif
+
+#define DISCORD_CALLBACK DISCORD_API
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include <string.h>
+#ifndef __cplusplus
+#include <stdbool.h>
+#endif
+
+#define DISCORD_VERSION 3
+#define DISCORD_APPLICATION_MANAGER_VERSION 1
+#define DISCORD_USER_MANAGER_VERSION 1
+#define DISCORD_IMAGE_MANAGER_VERSION 1
+#define DISCORD_ACTIVITY_MANAGER_VERSION 1
+#define DISCORD_RELATIONSHIP_MANAGER_VERSION 1
+#define DISCORD_LOBBY_MANAGER_VERSION 1
+#define DISCORD_NETWORK_MANAGER_VERSION 1
+#define DISCORD_OVERLAY_MANAGER_VERSION 2
+#define DISCORD_STORAGE_MANAGER_VERSION 1
+#define DISCORD_STORE_MANAGER_VERSION 1
+#define DISCORD_VOICE_MANAGER_VERSION 1
+#define DISCORD_ACHIEVEMENT_MANAGER_VERSION 1
+
+enum EDiscordResult {
+    DiscordResult_Ok = 0,
+    DiscordResult_ServiceUnavailable = 1,
+    DiscordResult_InvalidVersion = 2,
+    DiscordResult_LockFailed = 3,
+    DiscordResult_InternalError = 4,
+    DiscordResult_InvalidPayload = 5,
+    DiscordResult_InvalidCommand = 6,
+    DiscordResult_InvalidPermissions = 7,
+    DiscordResult_NotFetched = 8,
+    DiscordResult_NotFound = 9,
+    DiscordResult_Conflict = 10,
+    DiscordResult_InvalidSecret = 11,
+    DiscordResult_InvalidJoinSecret = 12,
+    DiscordResult_NoEligibleActivity = 13,
+    DiscordResult_InvalidInvite = 14,
+    DiscordResult_NotAuthenticated = 15,
+    DiscordResult_InvalidAccessToken = 16,
+    DiscordResult_ApplicationMismatch = 17,
+    DiscordResult_InvalidDataUrl = 18,
+    DiscordResult_InvalidBase64 = 19,
+    DiscordResult_NotFiltered = 20,
+    DiscordResult_LobbyFull = 21,
+    DiscordResult_InvalidLobbySecret = 22,
+    DiscordResult_InvalidFilename = 23,
+    DiscordResult_InvalidFileSize = 24,
+    DiscordResult_InvalidEntitlement = 25,
+    DiscordResult_NotInstalled = 26,
+    DiscordResult_NotRunning = 27,
+    DiscordResult_InsufficientBuffer = 28,
+    DiscordResult_PurchaseCanceled = 29,
+    DiscordResult_InvalidGuild = 30,
+    DiscordResult_InvalidEvent = 31,
+    DiscordResult_InvalidChannel = 32,
+    DiscordResult_InvalidOrigin = 33,
+    DiscordResult_RateLimited = 34,
+    DiscordResult_OAuth2Error = 35,
+    DiscordResult_SelectChannelTimeout = 36,
+    DiscordResult_GetGuildTimeout = 37,
+    DiscordResult_SelectVoiceForceRequired = 38,
+    DiscordResult_CaptureShortcutAlreadyListening = 39,
+    DiscordResult_UnauthorizedForAchievement = 40,
+    DiscordResult_InvalidGiftCode = 41,
+    DiscordResult_PurchaseError = 42,
+    DiscordResult_TransactionAborted = 43,
+    DiscordResult_DrawingInitFailed = 44,
+};
+
+enum EDiscordCreateFlags {
+    DiscordCreateFlags_Default = 0,
+    DiscordCreateFlags_NoRequireDiscord = 1,
+};
+
+enum EDiscordLogLevel {
+    DiscordLogLevel_Error = 1,
+    DiscordLogLevel_Warn,
+    DiscordLogLevel_Info,
+    DiscordLogLevel_Debug,
+};
+
+enum EDiscordUserFlag {
+    DiscordUserFlag_Partner = 2,
+    DiscordUserFlag_HypeSquadEvents = 4,
+    DiscordUserFlag_HypeSquadHouse1 = 64,
+    DiscordUserFlag_HypeSquadHouse2 = 128,
+    DiscordUserFlag_HypeSquadHouse3 = 256,
+};
+
+enum EDiscordPremiumType {
+    DiscordPremiumType_None = 0,
+    DiscordPremiumType_Tier1 = 1,
+    DiscordPremiumType_Tier2 = 2,
+};
+
+enum EDiscordImageType {
+    DiscordImageType_User,
+};
+
+enum EDiscordActivityPartyPrivacy {
+    DiscordActivityPartyPrivacy_Private = 0,
+    DiscordActivityPartyPrivacy_Public = 1,
+};
+
+enum EDiscordActivityType {
+    DiscordActivityType_Playing,
+    DiscordActivityType_Streaming,
+    DiscordActivityType_Listening,
+    DiscordActivityType_Watching,
+};
+
+enum EDiscordActivityActionType {
+    DiscordActivityActionType_Join = 1,
+    DiscordActivityActionType_Spectate,
+};
+
+enum EDiscordActivitySupportedPlatformFlags {
+    DiscordActivitySupportedPlatformFlags_Desktop = 1,
+    DiscordActivitySupportedPlatformFlags_Android = 2,
+    DiscordActivitySupportedPlatformFlags_iOS = 4,
+};
+
+enum EDiscordActivityJoinRequestReply {
+    DiscordActivityJoinRequestReply_No,
+    DiscordActivityJoinRequestReply_Yes,
+    DiscordActivityJoinRequestReply_Ignore,
+};
+
+enum EDiscordStatus {
+    DiscordStatus_Offline = 0,
+    DiscordStatus_Online = 1,
+    DiscordStatus_Idle = 2,
+    DiscordStatus_DoNotDisturb = 3,
+};
+
+enum EDiscordRelationshipType {
+    DiscordRelationshipType_None,
+    DiscordRelationshipType_Friend,
+    DiscordRelationshipType_Blocked,
+    DiscordRelationshipType_PendingIncoming,
+    DiscordRelationshipType_PendingOutgoing,
+    DiscordRelationshipType_Implicit,
+};
+
+enum EDiscordLobbyType {
+    DiscordLobbyType_Private = 1,
+    DiscordLobbyType_Public,
+};
+
+enum EDiscordLobbySearchComparison {
+    DiscordLobbySearchComparison_LessThanOrEqual = -2,
+    DiscordLobbySearchComparison_LessThan,
+    DiscordLobbySearchComparison_Equal,
+    DiscordLobbySearchComparison_GreaterThan,
+    DiscordLobbySearchComparison_GreaterThanOrEqual,
+    DiscordLobbySearchComparison_NotEqual,
+};
+
+enum EDiscordLobbySearchCast {
+    DiscordLobbySearchCast_String = 1,
+    DiscordLobbySearchCast_Number,
+};
+
+enum EDiscordLobbySearchDistance {
+    DiscordLobbySearchDistance_Local,
+    DiscordLobbySearchDistance_Default,
+    DiscordLobbySearchDistance_Extended,
+    DiscordLobbySearchDistance_Global,
+};
+
+enum EDiscordKeyVariant {
+    DiscordKeyVariant_Normal,
+    DiscordKeyVariant_Right,
+    DiscordKeyVariant_Left,
+};
+
+enum EDiscordMouseButton {
+    DiscordMouseButton_Left,
+    DiscordMouseButton_Middle,
+    DiscordMouseButton_Right,
+};
+
+enum EDiscordEntitlementType {
+    DiscordEntitlementType_Purchase = 1,
+    DiscordEntitlementType_PremiumSubscription,
+    DiscordEntitlementType_DeveloperGift,
+    DiscordEntitlementType_TestModePurchase,
+    DiscordEntitlementType_FreePurchase,
+    DiscordEntitlementType_UserGift,
+    DiscordEntitlementType_PremiumPurchase,
+};
+
+enum EDiscordSkuType {
+    DiscordSkuType_Application = 1,
+    DiscordSkuType_DLC,
+    DiscordSkuType_Consumable,
+    DiscordSkuType_Bundle,
+};
+
+enum EDiscordInputModeType {
+    DiscordInputModeType_VoiceActivity = 0,
+    DiscordInputModeType_PushToTalk,
+};
+
+typedef int64_t DiscordClientId;
+typedef int32_t DiscordVersion;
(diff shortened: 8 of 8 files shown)


Buggem:
> Looking it over I'm surprised you can't SDL_LoadObject and then load DiscordCreate... seems like they do everything right, it's versioned properly and the rest of the API is a function table to avoid needing more exports. Do they mention any specific reason that dlopening doesn't work?
> 
> Aside from that this looks okay; we might be able to avoid a lot of the build system goo if we can dynamically load like the other network backends.

I'm a real newbie to C/C++ and honestly do not understand how to dynamically load in the way that you do with the function tables. Also, adding libdiscord_game_sdk.so to the linker options does not actually merge it into the VVVVVV executable, it just fufills the headerfiles needs for the function calls that the SDK uses.

flibitijibibo:
This is definitely worth checking out:

https://wiki.libsdl.org/SDL2/SDL_LoadObject

The idea is that instead of linking the SDK, we just use the headers to get the type declarations, and then store a pointer to the DiscordCreate function. We initialize the pointer by dynamically loading the SDK, then loading the function by its name. You can see this in the other network files as an example.

Buggem:
> This is definitely worth checking out:
> 
> https://wiki.libsdl.org/SDL2/SDL_LoadObject
> 
> The idea is that instead of linking the SDK, we just use the headers to get the type declarations, and then store a pointer to the DiscordCreate function. We initialize the pointer by dynamically loading the SDK, then loading the function by its name. You can see this in the other network files as an example.

Ahh that is interesting. So you are not talking about removing the headerfile completely, only keeping the structs. Nice, I can definitely do that.

Buggem:
@flibitijibibo I've done your requests and the PR is almost ready! I just need to fix some potential crashes!
E.g. one potential crash is closing or restarting Discord while VVVVVV is open will lose the context of `app.core` (in short, it causes a segfault), but that will be patched soon. I also need to remove the DISCORD_shutdown stub and replace it with `ClearPointers`, and other function calls. After a few patches, it will be ready for 2.5 :)

Buggem:
Very strange finding...
With the warning this build added, when I removed it, triggered this out of the blue SDL2 error. Help?
```In file included from /usr/include/SDL2/SDL_config-x86_64.h:32,
                 from /usr/include/SDL2/SDL_config.h:58,
                 from /usr/include/SDL2/SDL_stdinc.h:31,
                 from /usr/include/SDL2/SDL_main.h:25,
                 from /usr/include/SDL2/SDL.h:32,
                 from /home/max/Downloads/VVVVVV/desktop_version/src/DiscordNetwork.c:17:
/usr/include/SDL2/SDL_platform.h:259:1: error: multiple storage classes in declaration specifiers
  259 | extern DECLSPEC const char * SDLCALL SDL_GetPlatform (void);
      | ^~~~~~
/usr/include/SDL2/SDL_platform.h:259:38: warning: ‘SDL_GetPlatform’ declared ‘static’ but never defined [-Wunused-function]
  259 | extern DECLSPEC const char * SDLCALL SDL_GetPlatform (void);
      |                                      ^~~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/VVVVVV.dir/build.make:846: CMakeFiles/VVVVVV.dir/src/DiscordNetwork.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:100: CMakeFiles/VVVVVV.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
cp: cannot stat './VVVVVV': No such file or directory
```

Buggem:
Huh. Moving the sdl2 header up worked. Huh..... :/

Buggem:
Ready for review :+1:

Daaaav:
I'm wondering how this should work with localization.

Ideally, every Discord user who sees the status messages sees them in _their own_ chosen Discord language, not whatever language the _player_ (the sender) happens to use. But unfortunately Discord does not support that at all - the game can only pass _one_ string, and that string is shown worldwide. So I'm not sure that string should be in the player's language... For example, I may prefer to have my game and my Discord in Dutch, but pretty much all my online friends are from other countries and I talk to them in English.

We've always made the localization features such that players should never have to feel forced to switch their game to English because of some kind of disadvantage, or accidentally reveal their in-game language to others when they wanted to keep it anonymous where they come from.

So... I guess the lesser evil would be to force Rich Presence in English for now and call it a Discord L, and see if they add proper multi-language support [later](https://github.com/discord/discord-rpc/issues/177). Or we could maybe add a game option whether to translate Rich Presence status messages or not (we already have a lot of options, though...)

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+#define DISCORD_LIBRARY "libdiscord_game_sdk.dylib"
+#elif defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__HAIKU__) || defined(__DragonFly__)
+#define DISCORD_LIBRARY "libdiscord_game_sdk.so"
+#else
+#error DISCORD_LIBRARY: Unrecognized platform!
+#endif
+
+
+#define DISCORD_CRASH_LIMIT 10
+#define DISCORD_CLIENT_ID 1315544357532729447 // TO TERRY/FLIBIT: You can create your own Discord instance at the Discord Developer Portal. This ID belongs to me, so just be aware that if my account was to get hacked, VVVVVV RPC would too. Use your own!
```
Daaaav, line 33: I feel like this comment gets added to the `#define` and ends up commenting out the semicolon where it's used? Might be wrong though.

It's good that this comment exists either way - the ID does need to be changed for production. I wonder if it'd be even better as a compile-time flag rather than hardcoded in...

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <assert.h>
+#include <SDL.h>
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <unistd.h>
+#include <string.h>
+#endif
```
Daaaav, line 14: Are these includes actually used?

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+        ClearPointers(); \
+        return 0; \
+    }
+    FUNC_LIST
+#undef FOREACH_FUNC
+    memset(&app, 0, sizeof(app));
+
+    struct DiscordCreateParams params;
+    params.client_id = DISCORD_CLIENT_ID;
+    params.flags = DiscordCreateFlags_Default;
```
Daaaav, line 101: I don't think we want to require Discord to be running? Which is what `Default` seems to do, as opposed to `NoRequireDiscord`. https://discord.com/developers/docs/developer-tools/game-sdk#createflags-enum

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+    if(app.activityMan == NULL)
+    {
+        app.activityMan = app.core->get_activity_manager(app.core);
+    }
+    strncpy(activity.state, name, 127);
+    strncpy(activity.assets.large_image, "vvvvvv", 127);
+    strncpy(activity.assets.large_text, level, 127);
+    activity.state[127] = '\0';
+    activity.assets.large_image[127] = '\0';
+    activity.assets.large_text[127] = '\0';
```
Daaaav, line 152: Instead of `strncpy` you can use `SDL_strlcpy` - which isn't so broken and doesn't require a separate `'\0'` assignment. You should also use `sizeof` instead of copying the buffer length. i.e.:

```suggestion
    SDL_strlcpy(activity.state, name, sizeof(activity.state));
    SDL_strlcpy(activity.assets.large_image, "vvvvvv", sizeof(activity.assets.large_image));
    SDL_strlcpy(activity.assets.large_text, level, sizeof(activity.assets.large_text));
```

(ditto the other place)

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+
+    app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
+
+    DISCORD_REQUIRE(app.core->run_callbacks(app.core));
+}
+void DISCORD_unlockAchievement(const char *name)
+{
+    // No "achivements" in Discord
```
Daaaav, line 160: Minor; typo

.vscode/settings.json
Daaaav, line 1: This file probably shouldn't be added.

Daaaav (changes requested):
Just did some light code quality scanning, no in-depth review yet.

Buggem:
> So... I guess the lesser evil would be to force Rich Presence in English for now and call it a Discord L, and see if they add proper multi-language support [later](https://github.com/discord/discord-rpc/issues/177). Or we could maybe add a game option whether to translate Rich Presence status messages or not (we already have a lot of options, though...)

Yes, I could do that very easily.

.vscode/settings.json
Buggem: Oof, sorry haha. Maybe add .vscode to the .gitignore in future?

desktop_version/src/DiscordNetwork.c
Buggem: `assert.h` should be fine to remove but STDInt and STDIO, I'm not sure. And also maybe `string.h` once I use SDL_strncopy

Daaaav: I was specifically talking about the `Windows.h` included only if `_WIN32`, and the `unistd.h` and `string.h` which are included only if not `_WIN32`.

Buggem: Well I removed `string.h` and `assert.h` and it now compiles fine.

desktop_version/src/DiscordNetwork.c
Buggem: Well, I moved it just incase

desktop_version/src/DiscordNetwork.c
Buggem: fixed

desktop_version/src/DiscordNetwork.c
Daaaav: What about `Windows.h` and `unistd.h`?

Buggem:
This PR is no longer ready. On Linux, if Discord is not installed...
Screenshot from 2024-11-30 10-08-23

This would break Steam Deck compatibility.

Buggem:
Oops! Minor mistake, only happened because I had an old build of my pull request installed onto steam. It's totally fine.

Buggem:
Oops (again) when running the build without discord, BOOM! A segfault. Patch incoming, lemme debug with GDB first.

Buggem:
Well, I patched it. I think it is finally ready for a merge. Merge when you want to, for now feel free to review the code. I will play my version of VVVVVV and look for potential segfaults for now. Bye!

Buggem:
@Daaaav I have completely finished your requests. (I forgot to remove the localisation but now that's done)

desktop_version/src/Network.c
```diff
@@ -1,4 +1,5 @@
 #include <stdint.h>
+#include <stdio.h>
```
Daaaav, line 2: `stdio.h` was added but not used - I see how that happens haha

third_party/libdiscord_game_sdk.so
Daaaav, line 1: I think we're not supposed to have this in the repo, and it's up to the person compiling to have these? That's how the Steam and GOG backends seem to work, anyway.

Daaaav (changes requested):
Did a slightly more thorough review of the code - and of course we're still gonna need to test this.

third_party/libdiscord_game_sdk.so
Buggem: They were just placeholder for earlier builds anyway, I'll just go and delete them. However, I'll have to get either Terry or Flibit to add the shared object files to the Steam/GOG/Humble builds. (they are the only people who I definitely know have write access to Steam)

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+    }
+    if(app.activityMan == NULL)
+    {
+        app.activityMan = app.core->get_activity_manager(app.core);
+    }
+    SDL_strlcpy(activity.state, name, sizeof(activity.state));
+    SDL_strlcpy(activity.assets.large_image, "vvvvvv", sizeof(activity.assets.large_image));
+    SDL_strlcpy(activity.assets.large_text, level, sizeof(activity.assets.large_text));
+
+    app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
```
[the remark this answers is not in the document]
Buggem: Yes, playtesting my version I feel that the update_activity call is chopping off FPS, even though the network implementation uses threaded instances.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+    params.flags = DiscordCreateFlags_NoRequireDiscord;
+
+    if(!DiscordCreate(DISCORD_VERSION, &params, &app.core))
+    {
+        return 1;
+    }
+
+    if(app.core == NULL)
+    {
+	discordNotDetected = true;
```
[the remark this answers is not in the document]
Buggem: I prefer tabs over spaces, but as the code uses spaces, I had to use spaces. I forget sometimes, lol.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+#undef FOREACH_FUNC
+    memset(&app, 0, sizeof(app));
+
+    struct DiscordCreateParams params;
+    params.client_id = DISCORD_CLIENT_ID;
+    params.flags = DiscordCreateFlags_NoRequireDiscord;
+
+    if(!DiscordCreate(DISCORD_VERSION, &params, &app.core))
+    {
+        return 1;
```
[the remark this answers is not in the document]
Buggem: Damn, this was a fluke that it even worked. Welp... I guess I'll just patch it.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,165 @@
...
+    {
+        ++discordCrashes;
+        DISCORD_shutdown();
+        DISCORD_init();
+        return;
+    }
+    discordCrashes = 0;
+}
+
+void DISCORD_update(const char *level, const char *name)
```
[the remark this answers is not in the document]
Buggem: @Daaaav GOGNetwork does not have the type* naming convention for functions I didn't implement, so I added that too. Should I just keep to my Discord implementation? Or would this fix be acceptable as part of this PR?

Buggem:
@Daaaav Completed your requests again :)

desktop_version/src/DiscordNetwork.c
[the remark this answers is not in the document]
Daaaav: I'd say it'd be acceptable to move spaces in the GOG/Steam code so that you don't have to add new code that either introduces `type *` convention or is inconsistent with the existing code. For more unrelated parts (functions you didn't implement for Discord but which were in the same GOG/Steam files you were already making consistency fixes in... Eh. Might as well.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,162 @@
...
+        DISCORD_REQUIRE(app.core->run_callbacks(app.core));
+        return;
+    }
+    SDL_strlcpy(activity.state, roomname, sizeof(activity.state));
+    SDL_strlcpy(activity.assets.large_image, "vvvvvv", sizeof(activity.assets.large_image));
+    SDL_strlcpy(activity.assets.large_text, area, sizeof(activity.assets.large_text));
+
+    app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
+
+    DISCORD_REQUIRE(app.core->run_callbacks(app.core));
```
Daaaav, line 153: I think how the update function should work is something like this:

```cpp
if (app.core->run_callbacks(app.core) != DiscordResult_Ok)
{
    // No Discord
    return;
}

// It was Ok, so we can now do things with Discord
if (activity is outdated)
{
    update activity;
}
```

Also, with C strings, if you want to check that two strings are equal, you need to do `SDL_strcmp(a, b) == 0` instead of `a == b`. `activity.state == roomname` and `activity.assets.large_text == area` will check that the two strings are stored in exactly the same bit of memory, rather than the strings having the same text.

Buggem: @Daaaav I did consider this, but you are meant to run the run_callbacks function after the activity update (source as per Discord C example)

Daaaav: I don't think it's necessarily that you need to run `run_callbacks` after the activity update; the example code just wanted to demonstrate updating the activity only one time and then run `run_callbacks` in a main loop, as a simplified example. And even if `run_callbacks` is the thing that "confirms" an activity update... well, then it's only going to be delayed by one frame.

The main advantage of switching them around is that we don't have to bother constantly updating the activity and telling Discord about it, if we know Discord isn't ready to take it anyway.

Buggem: Alright, in my opinion 1 frame delay is way too much (lol average gamer) but I can switch it around if you really want, I'll leave it as a single commit incase anyone wants to rollback.

Buggem: Wait... I don't actually thing we should do that. We need to handle for app.core == NULL not running a callback, as if app.core was to be NULL then we would have a segfault. So instead of replacing error handling with that I'll just add it on.

Buggem: The commit fucked something up - it no longer works. Reverting.

Buggem: I put some debug logs in for different scenarios:

[ERROR] Discord API failed to initialise!

Buggem: [ERROR] DiscordResult_Ok
[ERROR] Discord API failed to initialise!
even weirder... probably an easy fix

Daaaav: Take your time to get it fully working - you don't have to post a message every step along the way 😛

Buggem: Fixed it, just incorrect handling of DISCORD_REQUIRE, I fixed that but introduced another variable. I have to do it unless of course you want discordDetected to be true by default, which would be weird.

Daaaav (changes requested):
Not really exhaustive (I did a scan and called it a day):

- Little unsure about adding the whole `DiscordResult_Strings` array... Especially if it's not actively used... I guess we do uglier things, and it seems to be the only way to not have to say "Discord error 15" if we _do_ want debug messages...
- We tend to avoid commented-out code statements, like the `//vlog_error()`s. Maybe most of them would be good as `vlog_debug()` calls - `vlog_debug` is hidden by default so it's perfect for messages that might be a little too verbose. Console spam still isn't nice though lol, so no messages when `run_callbacks` can't find a running Discord instance every frame
- Code style nits, part of which took me a bit to notice and part of which are new: we always have a space after `if` (so `if (true)` instead of `if(true)`), and braces on a new line (no `if (true) {`). We also never put comments on the same line as a `{` - in that case the comment gets its own line.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,227 @@
...
+    FOREACH_FUNC(enum EDiscordResult, DiscordCreate, (DiscordVersion, struct DiscordCreateParams*, struct IDiscordCore**))
+
+
+
+#define FOREACH_FUNC(rettype, name, params) static rettype (*name) params = NULL;
+FUNC_LIST
+#undef FOREACH_FUNC
+
+bool DISCORD_REQUIRE(int x) {
+    //vlog_error(DiscordResult_Strings[x]);
```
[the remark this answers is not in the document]
Buggem: Yes... is there any way we could workaround this?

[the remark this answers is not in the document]
Daaaav: We have `INBOUNDS_ARR(x, DiscordResult_Strings)` which will return true if index `x` is in bounds.

Since `DISCORD_REQUIRE` is (at least in the current state) used for `run_callbacks`, I feel like maybe it should only print once if an error occurs. We have another macro for that: `WHINE_ONCE`/`WHINE_ONCE_ARGS`.

In all... This should work (definitely in the `x != DiscordResult_Ok` branch below):

```cpp
const char* message = "Unknown";
if (INBOUNDS_ARR(x, DiscordResult_Strings))
{
    message = DiscordResult_Strings[x];
}
WHINE_ONCE_ARGS(("Discord error: %s (%d)", message, x));
```

Buggem:
@Daaaav finished your req again. @flibitijibibo do you want to tag this?

Daaaav:
Note that we haven't tested the code yet (I might do that within the next week but no promises). I also still see a few code style things (like `if()` instead of `if ()` or `{`s not on their own line) but we're getting pretty close I think!

flibitijibibo:
Yeah I'll try to read through everything next week - took a quick look and it's definitely coming together nicely!

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,181 @@
...
+#elif defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__HAIKU__) || defined(__DragonFly__)
+#define DISCORD_LIBRARY "libdiscord_game_sdk.so"
+#else
+#error DISCORD_LIBRARY: Unrecognized platform!
+#endif
+
+
+#define DISCORD_CLIENT_ID 1315544357532729447
+
+// TO TERRY/FLIBIT: You can create your own Discord instance at the Discord Developer Portal. This ID belongs to me, so just be aware that if my account was to get hacked, VVVVVV RPC would too. Use your own!
```
flibitijibibo, line 27: @TerryCavanagh, does this already exist...?

Daaaav: For reference, the [Discord docs](https://discord.com/developers/docs/developer-tools/game-sdk#step-2-create-your-app) do say this:

> If you're integrating our SDK into an already-released game, there's a good chance that we may already have an application in our database for your game! Reach out to our [Dev Support](https://dis.gd/devsupport) to learn more

I wonder whether the reasons to use the ID they already have are good enough reasons to bother lol (maybe so it smoothly takes over the "Playing VVVVVV", "last played 2 days ago", etc that was already showing without rich presence?)

Buggem:
> Note that we haven't tested the code yet (I might do that within the next week but no promises). I also still see a few code style things (like `if()` instead of `if ()` or `{`s not on their own line) but we're getting pretty close I think!

Well I tried to fix that, I'll use a find and replace tool to help I guess.

desktop_version/src/DiscordNetwork.c
Buggem: @Daaaav @flibitijibibo I have sent a message to Discord Support to see if this is true for VVVVVV. Hopefully I get a response before the 15th anniversary of VVVVVV. (I'm hoping this will be released then)

NyakoFox: I'm sure an application already exists for VVVVVV, as it existed on the Discord Game Store back when that existed.

Buggem:
@Daaaav @flibitijibibo Happy new year! VVVVVV's 25th anniversary is coming soon and I am REALLY hoping to get it pushed by then. I also have some fixes coming in soon as well so get ready for that! Cheers and to a good 2025, Max Parry.

desktop_version/src/DiscordNetwork.c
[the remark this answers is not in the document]
Buggem: @Daaaav I don't think we should bother with this as spitting out inaccurate error codes is just as worse as a segfault.

Buggem:
One of the nice things about this PR is that if anyone wants to implement another type of RPC (eg Steam RPC) they already got a baseplate set up for them :D

Daaaav:
> Happy new year! VVVVVV's 25th anniversary is coming soon and I am REALLY hoping to get it pushed by then.

Happy new year!

To be clear, to my knowledge it's not planned for 2.5 to be released within the coming month. Unless you meant pushed to the repo and not pushed to store shelves, that one is realistic...

desktop_version/src/DiscordNetwork.c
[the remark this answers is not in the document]
Daaaav: It's not really relevant anymore since the error printing code is now removed, but no.

One is a best-effort "Unknown (54)" in console output that almost no player will ever read (and it's not lying about what error it is), the other is the game crashing completely when it should've been playable.

Buggem:
> > Happy new year! VVVVVV's 25th anniversary is coming soon and I am REALLY hoping to get it pushed by then.
> 
> Happy new year!
> 
> To be clear, to my knowledge it's not planned for 2.5 to be released within the coming month. Unless you meant pushed to the repo and not pushed to store shelves, that one is realistic...

Alright let's just push to the repo. I would have though something would happen with VVVVVV's 15th anniversary but if not, it's totally fine.

Buggem:
@flibitijibibo @Daaaav Can we please get this pushed soon? I don’t want this PR to go stale like it has for the past 9 days. I just want at least some activity (playtesting, etc)

flibitijibibo:
I've been stuck on SDL3 for a while (RC1 coming soon!) so I've not been able to check this out - upstream doesn't move very fast so even if this takes a while to review it shouldn't have to get rebased or anything like that since I can't really merge other stuff either.

desktop_version/src/DiscordNetwork.c
flibitijibibo: We'll need to lock this in before we can merge, so whoever can find this number should patch this line ASAP

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,183 @@
...
+
+
+struct DISCORD_application {
```
flibitijibibo, line 30: Variables like this can be static, since no other build unit depends on these

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,183 @@
...
+        return;
+    }
+    if (app.core == NULL || !discordDetected)
+    {
+        // No Discord
+        return;
+    }
+    if (!DISCORD_REQUIRE(app.core->run_callbacks(app.core)))
+    {
+        // Something  or other is wrong, but do we care?
```
flibitijibibo, line 159: Probably not, but Discord support might...? Not sure what their policy is on reporting internal errors

desktop_version/src/Network.c
flibitijibibo, line 2:
```suggestion
```

This might change for SDL3 but for now we can strip this out

desktop_version/src/main.cpp
```diff
@@ -978,8 +982,25 @@ static enum LoopCode loop_begin(void)
...
+    {
+        rpcArea = "Making a Level";
+        rpcRoomname = "";
+    }
+    // Dirty fix for custom levels getting the area from Dimension VVVVVV
+    if (map.custommode) {
+        rpcArea = game.customleveltitle.c_str();
+    }
+    // Update network APIs once per frame
+    NETWORK_update(rpcArea, rpcRoomname);
```
flibitijibibo, line 1003: Rich Presence isn't a real-time-critical feature, so while I'm trapped in SDL crunch it'd be great to add a real setRichPresence API that gets called only when game state or level data changes. Best case this does a lot of unused string churning, worst case it hammers a naive network API which might get our appid rate-limited.

mothbeanie:
> @flibitijibibo @Daaaav Can we please get this pushed soon? I don’t want this PR to go stale like it has for the past 9 days. I just want at least some activity (playtesting, etc)

Not to be rude, but you need to chill out... This is an entirely nonessential feature which has required a lot of intervention from maintainers and those more familiar with the codebase to fix bugs and and follow guidelines. 9 days is not a long time at all - being this impatient is really immature

flibitijibibo (changes requested):
The new Network backend is pretty good, but how its features are integrated could be less invasive.

Separately we should implement Steam rich presence, which thankfully is a one-liner: https://partner.steamgames.com/doc/features/enhancedrichpresence

A good reason to do this is to prepare localized strings:

https://partner.steamgames.com/doc/api/ISteamFriends#richpresencelocalization

If this is too much to do in one patch then we could split this up into two parts, where one part is rich presence and the second part is Discord integration. (Discord could be done first FWIW, it just wouldn't be fancy integration at first.)

Daaaav:
> before we can merge

(By the way, we haven't playtested yet, so please also wait with merging until we've had the chance to do that.)

(We have actually attempted to start playtesting but got stopped by some UB on Windows. So we're still working on it)

Buggem:
@flibitijibibo I would be happy to expand this PR to include the Steam network RPC (you may have noticed my comment in the code lol), as the way you’ve got it set up makes it really easy. And also make it less invasive (I’m thinking run NETWORK_setRPC only when RPC is actually updated, running NETWORK_update each frame separately) However as of now I am on holidays and won’t get back to my computers for the next week or so.

Feel free to suggest changes while I’m out, all of them will be greatly thanked and considered.

Buggem:
> (We have actually attempted to start playtesting but got stopped by some UB on Windows. So we're still working on it)

I'm back and WHAT?? Undefined behavior? It's working for me...

Buggem:
@flibitijibibo 

"About that Client ID I owe ya..."

Ask Terry to do it or maybe you can do it yourself. I got a response but they said "there is a ID for VVVVVV, but we can only provide to people on the dev team". You have access to the Steamworks profile for VVVVVV, right (I saw you posting announcements) so maybe you can do it.

Buggem:
@mothbeanie My past PR's to other very niche repos have been merged in less than a day without pinging people... so... ¯\_(ツ)_/¯

Daaaav:
> I'm back and WHAT?? Undefined behavior? It's working for me...

Yes, something about incorrect calling convention that only affects Windows.

> My past PR's to other very niche repos have been merged in less than a day without pinging people... so... ¯_(ツ)_/¯

I have a feeling those past PRs might've been very small changes that were easy to instantly see they were correct?

For this, we need to set aside parts of our free time to test thoroughly. And we have more things to do. This isn't critical.

Buggem:
> For this, we need to set aside parts of our free time to test thoroughly. And we have more things to do. This isn't critical.

Yep I understand.

desktop_version/src/Network.c
```diff
@@ -89,16 +107,22 @@ void NETWORK_shutdown(void)
...
+int32_t NETWORK_update(void)
 {
+    int32_t result = 0;
     #if NUM_BACKENDS > 0
     int32_t i;
     for (i = 0; i < NUM_BACKENDS; i += 1)
     if (backends[i].IsInit)
     {
-        backends[i].Update();
+        if ( backends[i].Update() )
```
flibitijibibo, line 119: We don't need to worry about Update status too much - as long as we do a good job with SetRPC this can stay as-is.

desktop_version/src/main.cpp
```diff
@@ -978,8 +986,29 @@ static enum LoopCode loop_begin(void)
...
+        rpcArea = "Making a Level";
+        rpcRoomname = "";
+    }
+    // Dirty fix for custom levels getting the area from Dimension VVVVVV
+    if (map.custommode) {
+        rpcArea = game.customleveltitle.c_str();
+    }
+    // Update network APIs once per frame (nested in if loop)
+    if( NETWORK_update() && ( SDL_strcmp(rpcArea, old_rpcArea) != 0 || SDL_strcmp(rpcRoomname, old_rpcRoomname) != 0 ) )
+    NETWORK_setRPC(rpcArea, rpcRoomname);
```
flibitijibibo, line 1008: Definitely better but still does a lot of string stuff - we should intercept changes to avoid doing work unless we absolutely know we need to... something like this maybe?

```suggestion
    if (game.gamestate != rpcGameState)
    {
        rpcGameState = game.gamestate;
        if (game.gamestate == TITLEMODE)
        {
            rpcArea = "Exploring the Menus";
            rpcRoomname = "";
        }
        else if (game.gamestate == EDITORMODE)
        {
            rpcArea = "Making a Level";
            rpcRoomname = "";
        }
        else
        {
            rpcArea = map.currentarea(game.roomx, game.roomy);
            rpcRoomname = map.roomname;
        }
        NETWORK_setRPC(rpcArea, rpcRoomname);
    }
    else
    {
         const char *nextArea, *nextRoom;

        // Dirty fix for custom levels getting the area from Dimension VVVVVV
        if (map.custommode)
        {
            nextArea = game.customleveltitle.c_str();
        }
        else
        {
            nextArea = map.currentarea(game.roomx, game.roomy);
        }
        nextRoom = map.roomname;

        if ((SDL_strcmp(rpcArea, nextArea) != 0) || (SDL_strcmp(rpcRoomname, nextRoom) != 0))
        {
            // FIXME: Are these pointers atually safe to store?
            rpcArea = nextArea;
            rpcRoom = nextRoom;
            NETWORK_setRPC(rpcArea, rpcRoomname);
        }
    }

    // Update network per frame.
    NETWORK_update();
```

flibitijibibo (changes requested):
It's definitely getting better - here's another quick review.

desktop_version/src/main.cpp
flibitijibibo: Not at my desk but out of curiosity I looked it up and none of the strcmp implementations I found do an early check for matching pointers, so that would definitely save some cycles if the string pointers are indeed stable.

Buggem:
Hmm something you commited made it not work... I'm looking through the code right now but it all looks good... debugging.

Buggem:
OH YES I THINK I GOT IT! It's the POINTERS!!!!!!!!!!!!!!!

Buggem:
Fixed the pointers... but the fix SUCKS!

desktop_version/src/DiscordNetwork.c
Buggem: I know coming back to this a little late, but a compile time flag would be BRILLIANT sorry I kinda skimped that part

leo60228:
I've implemented Steam rich presence here: https://github.com/leo60228/VVVVVV/commit/418ff074da4a9b161379f28c1abf8b51f6269f3e

This expects localization strings added in Steamworks along these lines:
```vdf
"lang"
{
	"english"
	{
		"tokens"
		{
			"#Default"	"%area%"
			"#NamedRoom"	"%area%: %name%"
		}
	}
}
```

Ideally, this would support localized room/area names, but this would significantly complicate both implementation and deployment.

desktop_version/src/DiscordNetwork.c
leo60228: The automatically assigned application ID is `491427544134975498`, but this doesn't currently work for rich presence. I'm not sure if the Discord Store release had a separate application ID.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,188 @@
...
+    if (libHandle != NULL)
+    {
+        app.core->destroy(app.core);
+        ClearPointers();
+    }
+}
+
+int32_t DISCORD_init(void)
+{
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__HAIKU__) || defined(__DragonFly__)
```
leo60228, line 89: This should probably be rewritten to check for OSes that Discord *does* support, rather than ones it doesn't.

desktop_version/src/DiscordNetwork.c
```diff
@@ -0,0 +1,188 @@
...
+    }
+    if (SDL_strcmp(activity.state, roomname) != 0 || SDL_strcmp(activity.assets.large_text, area) != 0)
+    {
+     	SDL_strlcpy(activity.state, roomname, sizeof(activity.state));
+        SDL_strlcpy(activity.assets.large_image, "vvvvvv", sizeof(activity.assets.large_image));
+        SDL_strlcpy(activity.assets.large_text, area, sizeof(activity.assets.large_text));
+
+        app.activityMan->update_activity(app.activityMan, &activity, NULL, NULL);
+    }
+
```
leo60228, line 185: Newlines in this file seem generally inconsistent. This should probably be cleaned up.

desktop_version/src/main.cpp
```diff
@@ -978,9 +989,56 @@ static enum LoopCode loop_begin(void)
...
+        else
+        {
+            SDL_strlcpy(rpcArea, map.currentarea(game.roomx, game.roomy), RPC_SAFE_BUFFER-1);
+            SDL_strlcpy(rpcRoomname, map.roomname, RPC_SAFE_BUFFER-1);
+        }
+        NETWORK_setRPC(rpcArea, rpcRoomname);
+    }
+    else
+    {
+         char nextArea[RPC_SAFE_BUFFER] = "", nextRoom[RPC_SAFE_BUFFER] = "";
```
leo60228, line 1016: This like is indented incorrectly.

desktop_version/src/main.cpp
```diff
@@ -978,9 +989,56 @@ static enum LoopCode loop_begin(void)
...
+        // Dirty fix for custom levels getting the area from Dimension VVVVVV
+        if (map.custommode)
+        {
+            SDL_strlcpy(nextArea, game.customleveltitle.c_str(), RPC_SAFE_BUFFER-1);
+        }
+        else
+        {
+            SDL_strlcpy(nextArea, map.currentarea(game.roomx, game.roomy), RPC_SAFE_BUFFER-1);
+        }
+        SDL_strlcpy(nextRoom, map.roomname, RPC_SAFE_BUFFER-1);
```
leo60228, line 1027: This should only run in GAMEMODE.

desktop_version/src/main.cpp
```diff
@@ -978,9 +989,56 @@ static enum LoopCode loop_begin(void)
...
+            SDL_strlcpy(rpcArea, nextArea, RPC_SAFE_BUFFER-1);
+            SDL_strlcpy(rpcRoomname, nextRoom, RPC_SAFE_BUFFER-1);
+            NETWORK_setRPC(rpcArea, rpcRoomname);
+        }
+    }
+
     // Update network per frame.
     NETWORK_update();
 
+
```
leo60228, line 1041: This line seems to have been added by accident.

desktop_version/src/main.cpp
```diff
@@ -978,9 +989,56 @@ static enum LoopCode loop_begin(void)
...
+        }
+        else
+        {
+            SDL_strlcpy(nextArea, map.currentarea(game.roomx, game.roomy), RPC_SAFE_BUFFER-1);
+        }
+        SDL_strlcpy(nextRoom, map.roomname, RPC_SAFE_BUFFER-1);
+        //printf("room %s == %s, area %s == %s\n", rpcRoomname, nextRoom, rpcArea, nextArea);
+        if ((SDL_strcmp(rpcArea, nextArea) != 0) || (SDL_strcmp(rpcRoomname, nextRoom) != 0))
+        {
+            // FIXME BEFORE PUSHING!!: These pointers aren't safe to use and result in RPC not updating when it needs to.
```
leo60228, line 1031: Does this still apply?

leo60228 (changes requested):
This mostly looks good, though I found a number of minor nitpicks.

desktop_version/src/main.cpp
Buggem: no i fixed it, i just forgot to remove it

Buggem:
@leo60228 I've implemented your changes

flibitijibibo (commented):
This is looking a lot better! The only note from my end is that we should have `discord_game_sdk.h` in `third_party/`, it might add an include path to CMakeLists but it'll prevent possible tomfoolery with the official SDK files that should be in the VVV source instead.

Will let others review the main.cpp changes but with the GAMEMODE change leo mentioned this is good enough IMO. Windows/macOS testing can be done afterward, the Linux-only block looks okay for now.

PR #1219, opened 2024-12-21
Fix squeak and save spam in gamepad menu

Daaaav:
## Changes:

If you push a button to set a controller binding, you may either hear one Viridian squeak, two Viridian squeaks (a louder one), or Viridian doesn't stop squeaking until you let go of the button. While you hear the continuous squeaking, your save file is also repeatedly saved.

(Don't forget to unmute the video:)

https://github.com/user-attachments/assets/1fdf2063-fbba-4823-a2e9-0761280d01bd

There are two small bugs at play here:
- the squeak is actually played in two different places at the same time (both in `titleinput()` whenever a button is pressed, and in `updatebuttonmappings()` when a mapping is succesfully changed)
- `titleinput()` doesn't register that a button is held down and applies the button (and saves to file) every frame for as long as the button is held

This PR fixes both these issues. Now a single button press always causes one squeak, and only if the bindings actually changed. Your save file is also no longer saved repeatedly from holding down the button.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -2377,16 +2377,25 @@ void titleinput(void)
             game.jumpheld = true;
         }
 
+        static bool controller_held = false;
+
         if (    game.currentmenuname == Menu::controller &&
                 game.currentmenuoption > 0 &&
                 game.currentmenuoption < 6 &&
                 (game.separate_interact || game.currentmenuoption < 5) &&
                 key.controllerButtonDown()      )
         {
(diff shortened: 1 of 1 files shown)


Daaaav:
(I think this might be nice as a 2.4 backport by the way, since it's a small fix to a pretty noticeable bug in this menu)

flibitijibibo:
Backported via https://github.com/TerryCavanagh/VVVVVV/commit/e72d59de564577990a3bf55a94332eb4566a282d

PR #1224, opened 2025-04-10
Replace colour IDs with an enum

NyakoFox:
## Changes:

Entity colors are just integers. Their colour ID gets passed through a big switch, returning different RGB values depending on the colour ID (and can also get affected by randomness or other game state.)

But because of this, there's a bunch of random numbers floating around, with no actual description on what they are other than comments which, while most of the time are accurate, only exist in the switch.

To fix this, this commit adds a new enum which labels every colour. While we can't use it as a type (as we need to allow colours outside of what are defined, in case people want a "pure white", and scripting can set any colour ID they want), colours have to stay as `int`.

Additionally, there seemed to be an old enum for specifically colors which get used for crewmates...? Which has now been removed.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Ent.cpp
+++ b/desktop_version/src/Ent.cpp
@@ -18,7 +18,7 @@ void entclass::clear(void)
     state = 0;
     statedelay = 0;
     life = 0;
-    colour = 0;
+    colour = EntityColour_CREW_CYAN;
     para = 0;
     behave = 0;
     animate = 0;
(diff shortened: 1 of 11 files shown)


flibitijibibo (approved):
Overall looks good - will wait for a +1 _juuust_ in case a color got changed by accident and hasn't been caught yet!

Daaaav:
Planning to do a quick find-and-replace soon to make sure there's not any 5 that turned into a 6 somewhere!

Daaaav (approved):
After some find-and-replacing, this comes back clean - looks like this shouldn't cause any change in behavior!

flibitijibibo:
Excellent - once conflicts are resolved will merge ASAP.

NyakoFox:
Fixed and force pushed.

PR #1225, opened 2025-04-11
Add player colour as a level property

NyakoFox:
## Changes:

This PR adds a new XML property for the player's colour. It is 0 by default, but you can change it to any colour ID. For example, making the player use the trinket color is `<PlayerColour>3</PlayerColour>`.

This is mostly a quality-of-life addition, as the player's colour is always 0 unless changed by scripts. A lot of levels which use different player colours use an intro script which both changes the player's colour and sets their respawn colour, which works great for finished, completed levels, but makes playtesting a little more annoying as they will spawn in as the wrong colour. Adding a level property for the default player colour fixes this annoyance.

Additionally, this changes the behavior of `restoreplayercolour`. This command used to set the player's colour to 0 (ignoring the respawn colour, so when Viridian would die, they would revert to the respawn colour). Now, it sets both Viridian's colour AND the respawn colour to what was present in the level file. This way, you can temporarily change the player colour using the script commands, and then use `restoreplayercolour` to revert back to what the player colour normally is.

The start point colour has also changed, to show the player colour instead of always colour 0.

Like most changes like this, a way to change this in-editor does not yet exist, but is planned for the future.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/CustomLevels.cpp
+++ b/desktop_version/src/CustomLevels.cpp
@@ -1566,20 +1566,6 @@ bool customlevelclass::save(const std::string& _path)
         }
     }
 
-    if (player_colour != 0)
-    {
-        xml::update_tag(msg, "PlayerColour", player_colour);
-    }
-    else
-    {
-        // Get rid of this one as well, since older levels don't have this property anyways
-        tinyxml2::XMLElement* element;
-        while ((element = msg->FirstChildElement("PlayerColour")) != NULL)
-        {
-            doc.DeleteNode(element);
-        }
-    }
-
     xml::update_tag(data, "mapwidth", mapwidth);
 
     xml::update_tag(data, "mapheight", mapheight);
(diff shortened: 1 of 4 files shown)


Daaaav (changes requested):
This doesn't fully work yet:
- In `customlevelclass::load()`, the tag is loaded from the wrong place (`<Data>` I think instead of `<MetaData>`) - so the XML loading code needs to be moved into the same block as where `"rtl"` is loaded.
- When playing a level from the levels menu (or via command line argument), the default color and savecolour are still 0, because `scriptclass::hardreset()` is called before the level XML is loaded.

NyakoFox:
I've fixed the latter -- it's intentionally in `<Data>` for the former, so no change needed.

Daaaav:
The former is still an issue - the game can load the tag if it's in `<Data>`, but it doesn't save it there. (I originally said it thinking that MetaData was the intended place because that's what the save code said, but it was the other way around)

NyakoFox:
Ah, I didn't notice -- fixed.

Daaaav (approved):
Looks good to me now!

flibitijibibo (approved):

PR #1226, opened 2025-04-11
Crew screen tweaks for custom levels

mothbeanie:
## Changes:

Currently, if a custom level has zero crewmates, the "CREW" screen will look like this:

test_before

However, many, many custom levels don't actually use rescueable crewmates, so seeing "Zero crewmates remain" for a level without any is rather odd. This commit modifies the crew (and stats) screens in custom levels to be more flexible.
If a level has no rescueable crewmates, that line will no longer display and the description will be moved down, like so:

test_basic

Additionally, if a level has no trinkets, the lines about how many trinkets you have are now removed from the "STATS" screen:

test_statsbefore
test_stats
(this accounts for language, of course)
test_statszh

I've also increased the amount of space around the player head sprite that displays next to the author name to be more friendly to custom graphics. Instead of only rendering in a 10x10 bounding box, the two pixels above and 7 pixels on each side will also render now. Here are some examples with custom sprites from the most recent VVVVVV Discord contest:

Cat's First Level - stupid cat
test_first

Consuming Alkahest - mothbeanie & Allison Fleischer
test_alka

the 50 move rule - sean (seaspaces)
test_50

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Graphics.cpp
+++ b/desktop_version/src/Graphics.cpp
@@ -425,22 +425,23 @@ void Graphics::print_level_creator(
     int width_for_face = 17;
     int total_width = width_for_face + font::len(print_flags, creator.c_str());
     int face_x, text_x, sprite_x;
+    int offset_x = -7;
     if (!font::is_rtl(print_flags))
     {
         face_x = (SCREEN_WIDTH_PIXELS - total_width) / 2;
         text_x = face_x + width_for_face;
-        sprite_x = 7;
+        sprite_x = 0;
     }
     else
     {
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Haven't been able to review but it makes sense, will let Terry review the design.

TerryCavanagh:
This looks great 😃

Daaaav (approved):
Looks good to me as well!

Daaaav:
Oh: the commit message probably needs to be updated. But I think there's an option to do that during merging as well and it's a single commit so presumably that's easily fixed without a force push.

flibitijibibo (approved):

PR #1227, opened 2025-04-11
Load an unlimited amount of sounds

NyakoFox:
## Changes:

This commit allows the usage of new "extra" sounds in levels for use in custom level scripting. Including `squid.ogg` inside of your sounds folder will now allow for the usage of the command `playef(squid)`. Additionally, you can play built-in sounds by name this way too, instead of having to memorize the id (`playef(rescue)`).

However, you CANNOT play extra sounds through using their numerical IDs, since those can shift around depending on the filesystem (or if someone adds a new sound). Playing an extra sound by ID will NOT play the sound.

This is another attempt at #902, being loosely based off of that PR.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Script.cpp
+++ b/desktop_version/src/Script.cpp
@@ -448,7 +448,14 @@ void scriptclass::run(void)
             }
             if (words[0] == "playef")
             {
-                music.playef(ss_toi(words[1]));
+                if (music.soundidexists(words[1].c_str()))
+                {
+                    music.playefid(words[1].c_str());
+                }
+                else if (!music.soundisextra(ss_toi(words[1])))
+                {
+                    music.playef(ss_toi(words[1]));
+                }
             }
             if (words[0] == "play")
             {
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -93,7 +93,7 @@ static FAudioMasteringVoice* masteringvoice = NULL;
 class SoundTrack
 {
 public:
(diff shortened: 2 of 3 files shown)


Daaaav (approved):
Looks good to me I think!

desktop_version/src/Script.cpp
```diff
@@ -448,7 +448,14 @@ void scriptclass::run(void)
...
             {
-                music.playef(ss_toi(words[1]));
+                if (music.soundidexists(words[1].c_str()))
+                {
+                    music.playefid(words[1].c_str());
+                }
+                else if (!music.soundisextra(ss_toi(words[1])))
+                {
+                    music.playef(ss_toi(words[1]));
+                }
```
flibitijibibo, line 458: Should we warn if a sound was requested and not found at all?

flibitijibibo (commented):
This seems okay; I'm worried about case sensitivity but the same concerns technically apply to the original data too...

desktop_version/src/Script.cpp
NyakoFox: I tend to not leave warnings for scripting, however maybe we should -- there should most likely be a better place to warn about script errors other than the console however, but I can add a warning here for now.

NyakoFox:
I have pushed another commit which, while case sensitive filenames CAN overlap (ex. `Hi.wav` and `hi.wav`), IDs will now get lowered and if one ID already exists it will refuse to load something with the same ID. It is also now more consistent with script arguments.

PR #1228, opened 2025-04-13
Add Persian localization

Daaaav:
## Changes:

A bit overdue since this was delivered a few weeks ago (oops), but here it is!

Thanks to Amir Arzani and Masoud Varaste.

(Terry agreed a 2.4 backport would be a good idea.)

## Legal Stuff:


Daaaav:
Thanks!

By the way, we have two more things planned for the next minor 2.4.x:
- Adding missing translations that currently still have temporary placeholders in most languages, from the last-minute 2.4 changes
- Some more improvements to the gamepad menu (recently had another support conversation on where someone's settings file was... so I'm adding a "press again to confirm" and a way to remove a binding)

flibitijibibo:
This patchset applied cleanly to 2.4, so this should be set for the next update.

PR #1229, opened 2025-04-23
Add confirmation and removal to bindings menu

Daaaav:
## Changes:

This makes the following improvements to the gamepad bindings menu:

- The menu now shows a hint that you can press a button while any of the bind options are selected (or that you can navigate away from those options)
- Instead of button presses immediately setting a binding, they now ask for confirmation: press the same button a second time to confirm
- You can now remove a binding, the same way you add it (this has the same type of confirmation)
- This menu used to be inconsistent with pretty much every other menu in the game by showing a permanent title and description for the menu itself ("Game Pad", "Change controller options.") rather than showing a title and description for the currently selected option. This inconsistency is now fixed.

Demo video:

https://github.com/user-attachments/assets/9a52d350-1905-492b-860e-010a019d201f

(Terry agreed a 2.4 backport would be a good idea, see also https://github.com/TerryCavanagh/VVVVVV/pull/1228#issuecomment-2800021061)

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Input.cpp
+++ b/desktop_version/src/Input.cpp
@@ -35,6 +35,37 @@ static void updatebuttonmappings(int bind)
     ) {
         if (key.isDown(i))
         {
+            if (!game.gpmenu_confirming || i != game.gpmenu_lastbutton)
+            {
+                game.gpmenu_confirming = true;
+                game.gpmenu_lastbutton = i;
+
+                // Is this button already in the list for this action?
+                std::vector<SDL_GameControllerButton>* vec = NULL;
+                switch (bind)
+                {
+                case 1: vec = &game.controllerButton_flip; break;
+                case 2: vec = &game.controllerButton_map; break;
+                case 3: vec = &game.controllerButton_esc; break;
+                case 4: vec = &game.controllerButton_restart; break;
+                case 5: vec = &game.controllerButton_interact; break;
+                default: return;
+                }
+
+                game.gpmenu_showremove = false;
+                for (size_t j = 0; j < vec->size(); j += 1)
+                {
+                    if (i == (*vec)[j])
+                    {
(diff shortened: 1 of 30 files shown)


desktop_version/src/Input.cpp
```diff
@@ -35,6 +35,37 @@ static void updatebuttonmappings(int bind)
...
+                // Is this button already in the list for this action?
+                std::vector<SDL_GameControllerButton>* vec = NULL;
+                switch (bind)
+                {
+                case 1: vec = &game.controllerButton_flip; break;
+                case 2: vec = &game.controllerButton_map; break;
+                case 3: vec = &game.controllerButton_esc; break;
+                case 4: vec = &game.controllerButton_restart; break;
+                case 5: vec = &game.controllerButton_interact; break;
+                default: return;
```
flibitijibibo, line 52: Is this only possible when a manually edited config has an invalid value? If so it may be worth warning just in case someone's messing with this stuff and wants to know what might be wrong.

flibitijibibo (commented):
Just one thing comes to mind, otherwise this lgtm.

desktop_version/src/Input.cpp
Daaaav: It's only if an unexpected menu option is selected - `bind` comes from the current menu option and this function is only called when it's between 1 and 5.

flibitijibibo:
Will backport this tomorrow!

flibitijibibo:
This is now in 2.4-updates - if there's anything else that should be marked for 2.4 let me know, otherwise I'll likely try to put 2.4.3 together after looking at the mouse thing.

Daaaav:
Currently working on a PR for the missing translations now that both Persian and this are merged - we're still waiting for a few translators but we're pretty far along!

PR #1230, opened 2025-04-24
Add sound(x), simplified version of internal command playef(x)

MarcellaHouston:
## Changes:

Simplified scripting should now recognize sound(x) as a function with the same effect as playef(x).

## Legal Stuff:


NyakoFox:
Looks good to me -- the only thing I'm worried about is the merge commit, but that could probably be squashed out.

(Was `playremix` always missing the `customtextmode` stuff? Weird.)

Daaaav:
Looks good to me!

PR #1236, opened 2025-05-03
Fix mouse coordinates being wrong on HiDPI displays

### LINKED ISSUE #1235: Mouse may get stuck on macOS if HiDPI is enabled
Reported on [Reddit](https://www.reddit.com/r/VVVVVV/comments/1gxrbli/how_to_use_level_editor_on_mac/) and [Steam](https://steamcommunity.com/app/70300/discussions/0/592885544240827187/) (without too many details but it should be the same issue) and now by someone on Discord: it may be impossible to use the editor on macOS because when the cursor is at the bottom right of the window, the game thinks it's closer to the top left.

It seems like it has to do with having a HiDPI display - I could reproduce it once I used a 4K monitor resolution and set the monitor to be "scaled" to 1080p:

https://github.com/user-attachments/assets/232e7882-b207-49ca-a3f4-1827f51c6b74

(The "scaled" setting that appeared for a 3840x2160 monitor, and caused it to happen:)

Image

It's probably a one-line change in our mouse handling once we know how we're miscalculating it (or maybe it's an SDL hint?), and it may warrant a 2.4 backport.

NyakoFox:
## Changes:

When writing the initial stretch mode code, the existence of HiDPI displays completely slipped my mind -- or at least I didn't realize that they'd be a problem.

We implement scaling modes ourselves, so transforming the mouse coordinates to our 320x240 viewport is done manually. Unfortunately, that code did not take into account HiDPI scaling whatsoever, meaning that all of the math which assumes the window size and the renderer size is wrong.

To fix this, we use `SDL_GetWindowSizeInPixels` and `SDL_GetWindow` to find the scaling factor, and then apply that to the mouse coordinates to get the mouse coordinates in the pixel-space instead, and then we do all of our normal logic after.

Small code style thing: C-style casts without a following space (`(int)var`) was done intentionally here -- I talked to a few other contributors and we agreed that the extra space after the cast (`(int) var`) was unnecessary and annoying, and in the future it should probably be fixed when we're working on code around/near it, like how we treat other inconsistent styling.

**NOTE:** This is marked as a draft, as this is untested code. I use a Windows machine with no easy access to anything else where I can test scaling, so if someone could test the PR and report back, I'll mark this as ready and we can hopefully get this into a 2.4 patch.

**NOTE:** This **bumps the minimum SDL version to 2.26.0**, which adds `SDL_GetWindowSizeInPixels()`!

Closes #1235.

## Legal Stuff:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
     runs-on: windows-latest
 
     env:
-      SDL_VERSION: 2.24.0
+      SDL_VERSION: 2.26.0
 
     steps:
     - uses: actions/checkout@v1
--- a/desktop_version/src/KeyPoll.cpp
+++ b/desktop_version/src/KeyPoll.cpp
@@ -571,8 +571,20 @@ void KeyPoll::Poll(void)
     SDL_Rect rect;
     graphics.get_stretch_info(&rect);
 
-    mousex = (raw_mousex - rect.x) * SCREEN_WIDTH_PIXELS / rect.w;
-    mousey = (raw_mousey - rect.y) * SCREEN_HEIGHT_PIXELS / rect.h;
+    int window_width;
+    int window_height;
+    SDL_GetWindowSizeInPixels(gameScreen.m_window, &window_width, &window_height);
+
+    int scaled_window_width;
+    int scaled_window_height;
+    SDL_GetWindowSize(gameScreen.m_window, &scaled_window_width, &scaled_window_height);
+
+    float scale_x = (float)window_width / (float)scaled_window_width;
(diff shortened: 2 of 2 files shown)


Daaaav:
I have a macOS where I could test (since I could reproduce it already), but I'd need a proper .app (ideally a before and after) since I don't want to install XCode or system-wide-installed SDL there!

flibitijibibo:
Was able to reproduce on Wayland and confirmed that this fixes the issue!

flibitijibibo:
Backported to 2.4 https://github.com/TerryCavanagh/VVVVVV/commit/cd785316b6a00fde1adb8d22b998ba69a0d691c4

Issue #7, opened 2020-01-10
[Docs] License is not open source


purpleidea:
It's cool that you released the code! Unfortunately it's not actually an open source license!

I would recommend you use something like the GPLv3+ like John Carmack did when he release Quake source.

Happy to answer any licensing questions you have! Good luck!

flibitijibibo:
We definitely misused the "open source" terminology; I think we've scrubbed all the references but if we missed any, let us know. The license will probably stay as-is, however.

purpleidea:
> The license will probably stay as-is, however.

That's too bad! I don't see how this license benefits you or anyone, so I'd advise rethinking the choice, or at least explaining it here. A proper license gives your code a chance to grow and live on in Debian and other cool places. Cheers!

Kissaki:
It’s still an incredible educational, exploration and archiving resource even without a FOSS or OSI license. So I have to strongly disagree with it not benefiting anyone.

While there is an OSI definition of open source I don’t think it’s prevalent enough that people using the term open source for projects where the source code is available in the open have to change their wording. That’s established just as much. Not in all communities, but still.

Calinou:
I'm inclined to agree with OP as well. [Non-commercial clauses](https://freedomdefined.org/Licenses/NC) don't really  provide anything that a copyleft license would provide today.

Moth-Tolias:
considering that the art - and any/all other assets not in this repo - are already not licenced for public use, having the noncommercial clause [which i believe is the only thing preventing this licence from being "truly" open source?] seems rather unnecessary.

Cpt-Derp:
If you're gonna release the source code, following the example of John Carmack (id Tech 1-4 and Quake/Quakeworld engines) wouldn't do harm, and would foster a healthy modding community.

benwiley4000:
1. Who cares what OSI decided or authorized, they don't get to control the free/OSS software community.
2. I don't see how there's any harm in the license specifying that you can't sell this game as your own.
3. The modding community has full license to work with the source code and release alterations (free of charge albeit). If they want to release modifications that include original assets and they want to work out an agreement with the creators of the game, the license permits that.

The one thing I would say is that it would be nice to modders to leave a way for them to earn income from their work if they put significant effort into a mod/extension that broadens the lasting appeal of VVVVVV as a product that people spend money on. Historically this hasn't been the case for game mods but there's an increasing awareness that modders increase the lasting value of games and help bring in profits for the companies whose games they mod.

[...]

PR #1249, opened 2025-09-11
Add License Exception for Debian Distribution.

computermouth:
I'd been emailing with Terry, who said:

```
I'm more than happy to grant one to allow official linux distributions of the full game.
[...]
For the avoidance of doubt, I'm happy for you to distribute both the full version of VVVVVV or the Make and Play version of VVVVVV on debian in whatever form they support.
```

I've put kinda the "big" ones in the list here, I'm personally really only interested in getting it into the Debian repos, but Ubuntu tends to follow those. Arch users will usually package anything they find that they can, and I put in Fedora for good measure.

This list could obviously grow out of hand if everyone wants to add every dang distro to the list, so maybe a more generic description will be needed later if enough interest is there.


computermouth:
Yeah, @flibitijibibo you and I had actually had a similar conversation on the Rogue Legacy repo. While I could agree that a flatpak will work for everyone, and probably have less issues with custom levels and payloads that get passed around, my personal interest is in Debian packaging.

I did reach out to the (Debian Mentors mailing list)[https://lists.debian.org/debian-mentors/2025/09/msg00049.html], and they agreed that VVVVVV could land in the `nonfree` section of the official repository.

The Exceptions list as-is would cover the Make And Play version built for a Debian nonfree package, but Terry was gracious enough to offer the official version as well, which is what really prompted this PR.

flibitijibibo:
> Yeah, @flibitijibibo you and I had actually had a similar conversation on the Rogue Legacy repo. While I could agree that a flatpak will work for everyone, and probably have less issues with custom levels and payloads that get passed around, my personal interest is in Debian packaging.

Ah right, the RL1 asset pack! I thought I recognized your name :sweat_smile: 

> I did reach out to the [Debian Mentors mailing list](https://lists.debian.org/debian-mentors/2025/09/msg00049.html), and they agreed that VVVVVV could land in the `nonfree` section of the official repository.
> The Exceptions list as-is would cover the Make And Play version built for a Debian nonfree package, but Terry was gracious enough to offer the official version as well, which is what really prompted this PR.

This is good to hear, and I think we can make this a long-term goal. To make this easier let's chop this up a bit... we can do these as separate PRs:

- Add license exceptions to M&P today - exceptions will be for exact repository URLs, in this case Debian nonfree with others to come later. Other repos can request an exception for M&P and we can fast-track that since we already know what the context is. Being clear about the exact repository helps since a major pain point for packagers is ambiguity about the license; since ours is a fully custom license we can try to be as nitpicky as possible to ease the burden off of the packagers in this area.
- Add a screen _just_ after localization selection that points to the commercial URL, for Full Not Official builds
- Once the above is done, I can backport the additional screen to 2.4, backport the license exceptions, and then tag a new release for packagers to work with.

In addition to making M&P accessible for packaging sooner, it also gives repositories time to request an exception so that everybody has a chance to deploy ASAP once we have the full version ready for distributions to build and publish.

computermouth:
> exceptions will be for exact repository URLs, in this case Debian nonfree with others to come later

What would you like to see URL-wise? `packages.debian.org` I think is just the web frontend, and `deb.debian.org` is just the default mirror (versus one of the other hundred mirrors).

flibitijibibo:
Let's go with the web frontend, to avoid specifying any particular mirror (including the default).

boolforge:
What about license change to GPLv3 or LGPLv3?

computermouth:
> What about license change to GPLv3 or LGPLv3?

It is not in the cards, at this time: https://github.com/TerryCavanagh/VVVVVV/issues/7

flibitijibibo (approved):
This should be a good start, the in-game text can be the next step.

Since this involves licensing, need a +1 from Terry before mashing Rebase.

PR #1250, opened 2025-10-08
Enable System-installed Lodepng

computermouth:
Lodepng is packaged by Debian and Ubuntu. This PR enables the use of a system-installed Lodepng.

It functions identically to the other dynamically linked libraries, with the exception of the compile-time define in the `lodepng_wrapper.c`

## Legal Stuff:

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -240,7 +240,7 @@ if(APPLE)
     set(PFS_SRC ${PFS_SRC} ../third_party/physfs/src/physfs_platform_apple.m)
 endif()
 set(PNG_SRC src/lodepng_wrapper.c)
-set(PNG_OPT -DLODEPNG_NO_COMPILE_ALLOCATORS -DLODEPNG_NO_COMPILE_DISK)
+set(PNG_DEF -DLODEPNG_NO_COMPILE_ALLOCATORS -DLODEPNG_NO_COMPILE_DISK)
 set(CHM_SRC ../third_party/c-hashmap/map.c)
 set(SBIDI_SRC ../third_party/SheenBidi/Source/SheenBidi.c)
 


computermouth:
I see that the github actions used in this repository use `ubuntu-latest` but uses a container image `registry.gitlab.steamos.cloud/steamrt/sniper/sdk:latest` for the Linux build. I've never really wrapped my head around what `runs-on` and `container` mean in conjunction, maybe the first is just an ubuntu host capable of running docker?

The latter is a container image based on `debian:bullseye`. Bullseye is ancient, it's not even the old stable, it's the old old stable. Unfortunately lodepng started being packaged in the version after that. The `sniper/sdk` readme suggests maybe using backports, but the package isn't there either. I could assist in adding newer repositories, and setting configuration such that it only pulls in `liblodepng-dev` from the newer ones...

But up next, I'm packaging `c-hashmap` and `sheenbidi` as well, and if everything goes smoothly they would land in `debian:forky`, and if we were to try and get all those packages in-use in the steam sdk container, it'd mean installing packages from 3 different versions of Debian, with a range of up to 8 years difference between the artifacts in the repositories. Which usually ends up landing developers in a dependency hell.

At the end of the day, maybe we shouldn't merge this, nor any future changes for `c-hashmap` or `sheenbidi`, until they're all available in the desired build container. For my purposes, I can use build-time patches to change the `CMakeLists.txt` to link those with system libraries (once I've packaged them).

flibitijibibo:
The container is the SDK version of what ships as the Steam Linux Runtime - when the game is run it's basically put in a bubblewrapped container, currently based on Bullseye but will get bumped for a future release. We mostly do this to lock down on the compiler and glibc version; everything we use is still bundled in official builds.

flibitijibibo:
> But I guess what's the criteria for non-official builds, or more specifically, the criteria for non-official `BUNDLE_DEPENDENCIES=OFF` builds. Maybe given that it works on the last 2 Debian versions as well as Ubuntu's latest and Ubuntu LTS, it is good enough

I guess if it's going to be packaged all that _really_ matters is that it works with what the system has - if it doesn't, the VVV package either has to move to static linking or update the dependency upstream; we don't mind updates to the third_party folder if they don't happen _too_ often (which I wouldn't expect from a distribution using normal stable releases anyway).

computermouth:
Sounds good. I just noticed one variable name that I forgot to update, fixed now.

PR #1257, opened 2025-11-29
Remove strange entity-wrapping code

7f-galaxii:
## Changes:

This removes a block of code that checks for specifically vertical screenwrap, and then does horizontal entity wrapping. Under normal circumstances, this code is never reached, because entity bounds aren't expanded in the x-direction if horizontal wrapping is disabled.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Logic.cpp
+++ b/desktop_version/src/Logic.cpp
@@ -1125,41 +1125,6 @@ void gamelogic(void)
             }
         }
 
-        if (map.warpy && !map.warpx && !map.towermode)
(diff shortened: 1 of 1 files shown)


7f-galaxii:
I ran this PR by both @mothbeanie and @NyakoFox, and they agreed that despite changing behavior, it would not change behavior in any meaningful way. Notably, it does not affect speedrunning, no custom levels have ever used it, and this is behavior that can be replicated by using intended features (setting warp direction to "All").

PR #1258, opened 2025-12-03
Fix two-frame script delay when respawning

NyakoFox:
## Changes:

At some point, there was some "kludge" added to the game which allowed script boxes to immediately get triggered when you enter a room, rather than the update order unfortunately causing scripts to trigger after two frames.

For some reason, dying and respawning in a room does not have the same fix, causing really bad inconsistencies for anything that needs precise timing.

Despite the initial implementation of the kludge being somewhat careless and breaking levels which relied on timing, this one should not break anything, but in fact fix things, surprisingly.

## Legal Stuff:


Daaaav (approved):
This makes sense and seems to work nicely!

Issue #976, opened 2023-05-18
Editor improvements


AllyTally:
- [ ] Line scrolling for scripts
- [ ] Controller support (requires #834)
- [ ] Editable textbox colors
- [ ] Editable animated (special) room names
- [ ] Changing sprites of checkpoints/terminals
- [ ] Automatic internal script switch which does the boilerplate for you

PR #1260, opened 2025-12-27
added scrolling feature for the ingame script editor

motta-git:
## Changes:

As requested by the issue: https://github.com/TerryCavanagh/VVVVVV/issues/976

i have added the 'Line scrolling for scripts' feature by allowing the user to now scroll using the mouse scroll wheel  or with the PAGEUP and PAGEDOWN buttons on the keyboard.

If the user scrolls way below where the cursor is and then begins typing then the scroll will go up to where the cursor is.

Here is a small video demonstrating the feature:
[Screencast from 2025-12-27 12-58-18.webm](https://github.com/user-attachments/assets/d0083de9-f4e1-416e-8a5b-61dc9b9e9cfc)

This Patch greatly improves the in game script editor as it now allows to display large scripts more comfortably by simply scrolling instead of having to press the up and down buttons on the keyboard wich is a bit uncomfortable.

Thanks!

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -3815,7 +3815,6 @@ void editorinput(void)
                 int max_offset = SDL_max(0, (int) ed.script_buffer.size() - 1);
                 // Positive wheel.y -> scroll up (decrease offset)
                 ed.script_offset = SDL_clamp(ed.script_offset - key.mousewheel, 0, max_offset);
-                ed.keydelay = 6;
                 ed.script_user_scrolled = true;
             }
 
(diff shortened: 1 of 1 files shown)


desktop_version/src/Editor.cpp
```diff
@@ -3801,6 +3806,36 @@ void editorinput(void)
...
+                ed.script_user_scrolled = false;
+            }
+
+            // Mouse wheel scrolling and page keys
+            if (key.mousewheel != 0)
+            {
+                int max_offset = SDL_max(0, (int) ed.script_buffer.size() - 1);
+                // Positive wheel.y -> scroll up (decrease offset)
+                ed.script_offset = SDL_clamp(ed.script_offset - key.mousewheel, 0, max_offset);
+                ed.keydelay = 6;
```
NyakoFox: Is there a reason keydelay is set with mouse scrolling?

desktop_version/src/Editor.cpp
```diff
@@ -3801,6 +3806,36 @@ void editorinput(void)
...
+                ed.script_user_scrolled = true;
+            }
+
+            if (key.isDown(SDLK_PAGEUP) && ed.keydelay <= 0)
+            {
+                ed.keydelay = 6;
+                int max_offset = SDL_max(0, (int) ed.script_buffer.size() - 1);
+                int page = SDL_max(1, ed.lines_visible - (2 * SCRIPT_LINE_PADDING));
+                ed.script_offset = SDL_max(0, ed.script_offset - page);
+                ed.script_offset = SDL_min(ed.script_offset, max_offset);
```
NyakoFox: Could this be simplified to a single `SDL_clamp`? (that exists in the version we use, right?)

desktop_version/src/Editor.cpp
motta-git: Hello, pleasure talking to you! I initially thought that it would be better with the delay but i just tested it without the delay and it made the scroll feel a lot more responsive! In the new commit i made i fixed this!

desktop_version/src/Editor.cpp
motta-git: Yes! SDL_clamp is available in this codebase and this same cpp that we are editing allready used it in other parts of the code! In the new commit i made i adressed this!

desktop_version/src/Editor.cpp
```diff
@@ -3801,6 +3806,36 @@ void editorinput(void)
...
+                ed.script_user_scrolled = true;
+            }
+
+            if (key.isDown(SDLK_PAGEDOWN) && ed.keydelay <= 0)
+            {
+                ed.keydelay = 6;
+                int max_offset = SDL_max(0, (int) ed.script_buffer.size() - 1);
+                int page = SDL_max(1, ed.lines_visible - (2 * SCRIPT_LINE_PADDING));
+                ed.script_offset = SDL_min(max_offset, ed.script_offset + page);
+                ed.script_user_scrolled = true;
```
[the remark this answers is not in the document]
motta-git: I adressed this i my new commit!

PR #1266, opened 2026-02-26
Add statedelay argument to `gamestate` command

NyakoFox:
## Changes:

The `gamestate` command has an argument to set the gamestate, but the statedelay was always set to 0, despite the statedelay being an important part of the gamestate system. This commit adds an optional argument to set the statedelay.

## Legal Stuff:


Daaaav (approved):
Pretty simple change, lgtm!

PR #1267, opened 2026-03-24
Solve render recaching issues [NEEDS BACKPORT]

NyakoFox:
## Changes:

Past solutions were just "recache the screen textures under these known circumstances" where whenever things OUTSIDE of those known circumstances happened, the issue would reoccur.

I recently learned about an SDL event, `SDL_RENDER_TARGETS_RESET`, which is for this exact problem. I ripped out all of the other places `Screen::recacheTextures()` was called, and just slotted it in there, and it worked perfectly.

...well, the old behavior worked perfectly; but the old behavior was flawed as well, because it only checked for "ingame_titlemode" (if you're in the main menu during gameplay) and forgot to check for the map screen... plus, it didn't ever regenerate the minimap in custom levels, which is another "persistant" render target.

Hopefully, this is the last time we'll ever have to think about this one. I'm certainly sick of it.

This should 100% be backported into 2.4, as the bug occurs there as well. (The commit shouldn't conflict if cherry-picked into `2.4-updates`.)

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -3625,7 +3625,6 @@ void editorinput(void)
                 if (game.currentmenuname == Menu::ed_settings)
                 {
                     ed.state = EditorState_DRAW;
-                    gameScreen.recacheTextures();
                 }
                 else
                 {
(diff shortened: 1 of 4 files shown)


Daaaav:
I tested it and looks like it works! No more minimap becoming black because you resized the window, or started playtesting in fullscreen :)

Daaaav (approved):

flibitijibibo:
2.4 backport https://github.com/TerryCavanagh/VVVVVV/commit/4b9f26bf81e4389bd1f059d3053f3051670774b9

Finished my backlog of projects last week, so will aim for a 2.4 update on the 21st.

PR #1140, opened 2024-01-29
Implement scaling modes manually

AllyTally:
## Changes:

For future PRs, it'll be very nice to have full control over how VVVVVV gets drawn to the window. This means we can use the entire window size for things like touch input, drawing borders, or anything we want.

## Legal Stuff:

By submitting this pull request, I confirm that...

- [x] My changes may be used in a future commercial release of VVVVVV
- [x] I will be credited in a `CONTRIBUTORS` file and the "GitHub Friends"
  section of the credits for all of said releases, but will NOT be compensated
  for these changes unless there is a prior written agreement

PR #1269, opened 2026-04-11
Clamp editor mouse coordinates [NEEDS BACKPORT]

NyakoFox:
## Changes:

The editor has never clamped mouse bounds, and instead relied on SDL to return clamped mouse coordinates. Unfortunately, in #1140, this detail was missed, and the behavior changed, allowing the cursor to leave the bounds of the screen.

This isn't much of a problem most of the time because anything that uses position as array indeces has bounds checks. Unfortunately, placing entities outside of a room's bounds leads to entities you can't ever touch ever again without modifying the level externally. This is the worst with checkpoints, as pressing Enter in a room will now take you to the out-of-bounds checkpoint, causing you to immediately wrap to another room.

This fix should be backported to 2.4, as the issue can lead to level files which need manual editing to fix!

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -3206,8 +3206,8 @@ void editorinput(void)
     ed.old_tilex = ed.tilex;
     ed.old_tiley = ed.tiley;
 
-    ed.tilex = key.mousex / 8;
-    ed.tiley = key.mousey / 8;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
2.4 backport https://github.com/TerryCavanagh/VVVVVV/commit/6ed72297da752666008e363c837d421f6ead4d9a

PR #1270, opened 2026-04-30
Fix audio buffer issues on some platforms

NyakoFox:
## Changes:

On certain ports of the game, and possibly some Linux distros, the game's audio would crackle, or slow down with loud buzzing. This commit attempts to fix that, by enabling `FAUDIO_1024_QUANTUM`, which should increase the internal FAudio buffer size.

This could possibly be backported to 2.4, but I'm not sure how urgent it is. It's a shame we didn't fit this into the most recent backport version a few days ago... On the other hand, a new backport version could allow us to look into the font segfault that we completely forgot about. Either way, that's not relevant for this PR.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -745,7 +745,7 @@ static void add_builtin_sound(const char* id)
 
 void musicclass::init(void)
 {
-    if (FAudioCreate(&faudioctx, 0, FAUDIO_DEFAULT_PROCESSOR))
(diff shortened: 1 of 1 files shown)


flibitijibibo:
2.4 backport https://github.com/TerryCavanagh/VVVVVV/commit/08c3590a0a87e2824d23064d85dbab25c0479673

Once the font thing is fixed I can do another quick update with these changes.

PR #1275, opened 2026-05-07
Improve clarity and usability of room coordinates

Daaaav:
## Changes:

Room coordinates in the VVVVVV community have always been a source of misunderstandings. VVVVVV's editor has always displayed (1,1)-(20,20), but `gotoroom(x,y)` has always used (0,0)-(19,19), so Ved has always had an option to display coordinates as such, and most Ved users I know seem to have always preferred that.

The problem is that both coordinate systems look the same, and not everyone is aware there are two systems. So when an internal scripting veteran says there's a wall glitch in room (7,15), you may be left staring at (7,15) not finding anything, until you decide to check (8,16), still not finding anything... Before you realize you had it the wrong way around and you should've looked at (6,14) instead.

We can't get rid of either of the two systems either, since both are actually preferred by different people (we have had at least one person say 1-indexing was more intuitive to them!) and both systems are even eternalized in different scripting commands.

After some discussion about how to solve the confusion, people seem to agree to change how editors display coordinates as follows:
- 1-indexed coordinates are always displayed as `(1;1)`-`(20;20)` (with the `,` replaced by a `;`)
- 0-indexed coordinates are always displayed as `[0,0]`-`[19,19]` (with the `()` replaced by `[]`)

The idea behind this is that people will likely copy what the editor says - so if the editor says (7;15) you'll likely write (7;15) or 7;15, and if the editor says [6,14] you'll likely write [6,14] or 6,14.
It's not waterproof, but it seems much more likely to succeed than something like (7,15)1 or (6,14)0, which would inevitably be shortened by people who are unaware of what it means.

Something similar is done with video timecodes, where 00:00:00:00 means one counting method and 00:00:00;00 or 00;00;00;00 means another.

So, VVVVVV's editor still displays coordinates as 1-indexed, but just shows them with a semicolon. The same change was made in Ved.

Another improvement is the "go to room" input. Previously, coordinates had to be entered as `x,y`, for example `9,15`. My initial thought was to make this input accept `;` as well, and to change the label to say `x;y`. Then I discovered not all translators carried over the exact `x,y` format! Instead of changing only part of the languages (now knowing some users are going to have an exceptionally hard time with this input...) or making sure every string is consistent, I decided it would be better to make the input more like Ved's "go to room" input.
    
The input is now shown as `__;__`. Only digits are accepted as input. Which means you can enter 9,15 as 0915. Or as 9,15 or 9;15: typing a comma or semicolon in the second position inserts a 0 before it.

https://github.com/user-attachments/assets/f36f6857-aa25-411f-b2fa-cdabd51313db

Lastly, there is now a (currently config-file-only) option to switch VVVVVV's editor to 0-indexed coordinates, which, if set, will be shown in the aforementioned [0,0] style.

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Editor.cpp
+++ b/desktop_version/src/Editor.cpp
@@ -1033,7 +1033,7 @@ static void draw_entities(void)
 
                 if (i == edent_under_cursor)
                 {
-                    text = help.disp_room_coords(entity->p1 / 40, entity->p2 / 30, 1);
+                    text = help.disp_room_coords(entity->p1 / 40, entity->p2 / 30, game.coords0 ? 0 : 1);
                 }
                 else
                 {
@@ -1152,7 +1152,7 @@ static void draw_entities(void)
 
             if (ed.tilex == x / 8 && ed.tiley == y / 8)
             {
-                text = help.disp_room_coords(entity->rx, entity->ry, 1);
+                text = help.disp_room_coords(entity->rx, entity->ry, game.coords0 ? 0 : 1);
             }
             else
             {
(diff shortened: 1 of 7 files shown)


flibitijibibo:
Interesting... I didn't make maps myself so I'll take level makers' word on this one. Terry can make the final call for design changes as usual.

PR #1276, opened 2026-05-07
Add `savewarpdir(rx,ry)` and `unsavewarpdir(rx,ry)`

Ratismal:
These commands are used to persist and unpersist warpdir states, causing them to be saved to the player's save file.

Save files contain a new property `savedwarpdirs`, which is a comma separated list of ints for each room, like the `worldmap` (explored) value. Unlike `worldmap`, only select rooms should be persisted, so all unpersisted entries are left blank instead of containing a 0.

The approach for this change has been discussed with both @NyakoFox and @mothbeanie.

Demonstration of the commands:
https://github.com/user-attachments/assets/bda1a8e6-2473-4af3-81ec-11cffaa75242

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Game.cpp
+++ b/desktop_version/src/Game.cpp
@@ -5994,6 +5994,38 @@ void Game::customloadquick(const std::string& savfile)
                 map.setregion(thisid, thisrx, thisry, thisrx2, thisry2);
             }
         }
+        else if (SDL_strcmp(pKey, "savedwarpdirs") == 0)
+        {
+            char buffer[16]; 
+            size_t start = 0; 
+            size_t i = 0; 
+                
+            while (next_split_s(buffer, sizeof(buffer), &start, pText, ',')) 
+            {
+                // Use array size from map.explored, since savedwarpdirs doesn't have a fixed size
+                if (i >= SDL_arraysize(map.explored)) 
+                {
+                    break;
+                }
+                
+                // If empty, ignore line
+                if (buffer[0] == '\0')
+                {
+                    ++i;
+                    continue;
+                }
+
+                int warpdir = help.Int(buffer);
+                map.savedwarpdirs[i] = warpdir;
+
+                int temprx = i % 20;
+                int tempry = i / 20;
+
+                cl.setroomwarpdir(temprx, tempry, warpdir);
+
+                ++i; 
+            } 
+        }
     }
 }
 
@@ -6355,6 +6387,21 @@ bool Game::customsavequick(const std::string& savfile)
         customcollect += help.String((int) obj.customcollect[i]) + ",";
     }
     xml::update_tag(msgs, "customcollect", customcollect.c_str());
+
+    std::string savedwarpdirs;
+    // savedwarpdirs won't contain values for each room. Reuse room count from map.explored instead
+    for (size_t i = 0; i < SDL_arraysize(map.explored); i++)
+    {
+        if (map.savedwarpdirs.count(i) == 0)
+        {
+            savedwarpdirs += ",";
+        }
+        else
+        {
+            savedwarpdirs += help.String(map.savedwarpdirs[i]) + ",";
+        }
+    }
+    xml::update_tag(msgs, "savedwarpdirs", savedwarpdirs.c_str());
 
     //Position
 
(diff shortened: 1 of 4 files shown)


NyakoFox:
Haven't reviewed the code yet, but the design was discussed with us as mentioned.

For reference, this is, in a way, technically a fix; changing the warping state of a room is something which definitely should be saved to the save file, but it was not implemented, despite this command existing solely for custom level usage.

Now that more advanced levels have used them as temporary storage however, we can't simply make all of them save; which is where `savewarpdir(room_x,room_y)` comes in, solving the problem of "we want this to be saved, but can't break existing levels".

Because of that, though, I personally think that these commands should exist in simplified as well; I didn't think about that until now, and haven't discussed it with anyone else yet, but it would make sense to me given the intended use-case of these commands.

~~I'm not at my PC at the moment (it's 4:30 AM at the time of writing this), but the GitHub app reports a merge conflict; I assume what happened here is you just forgot to pull master before making a new branch, but that's simple enough to fix with a rebase. I can't actually... check what the conflict is, but I'm sure a simple rebase should fix it.~~

EDIT: I was just told that my "GitHub app is haunted" and "the merge conflict doesn't exist" so disregard that I guess. Thanks GitHub mobile.

Ratismal:
I've pushed an update to add these commands to simplified scripting. Please let me know if you'd like me to rebase it, not fully sure the practices there.

desktop_version/src/Game.cpp
```diff
@@ -5994,6 +5994,37 @@ void Game::customloadquick(const std::string& savfile)
...
+            while (next_split_s(buffer, sizeof(buffer), &start, pText, ',')) 
+            {
+                // Use array size from map.explored, since savedwarpdirs doesn't have a fixed size
+                if (i >= SDL_arraysize(map.explored)) 
+                { 
+                    break; 
+                } 
+                
+                // If empty, ignore line
+                if (buffer[0] == '\0') {
```
Daaaav:
```suggestion
                if (buffer[0] == '\0')
                {
```

desktop_version/src/Script.cpp
```diff
@@ -289,6 +289,19 @@ void scriptclass::run(void)
                     position--;
                 }
             }
+            if (words[0] == "savewarpdir") {
```
Daaaav:
```suggestion
            if (words[0] == "savewarpdir")
            {
```

desktop_version/src/Script.cpp
```diff
@@ -289,6 +289,19 @@ void scriptclass::run(void)
...
+                int temprx = ss_toi(words[1]) - 1;
+                int tempry = ss_toi(words[2]) - 1;
+
+                const RoomProperty* const room = cl.getroomprop(temprx, tempry);
+                map.savewarpdir(temprx, tempry, room->warpdir);
+            }
+            if (words[0] == "unsavewarpdir") {
```
Daaaav:
```suggestion
            if (words[0] == "unsavewarpdir")
            {
```

desktop_version/src/Game.cpp
```diff
@@ -6356,6 +6387,19 @@ bool Game::customsavequick(const std::string& savfile)
     }
     xml::update_tag(msgs, "customcollect", customcollect.c_str());
 
+    std::string savedwarpdirs;
+    // savedwarpdirs won't contain values for each room. Reuse room count from map.explored instead
+    for (size_t i = 0; i < SDL_arraysize(map.explored); i++)
+    {
+        if (map.savedwarpdirs.count(i) == 0) {
```
Daaaav:
```suggestion
        if (map.savedwarpdirs.count(i) == 0)
        {
```

desktop_version/src/Game.cpp
```diff
@@ -6356,6 +6387,19 @@ bool Game::customsavequick(const std::string& savfile)
...
+            savedwarpdirs += ",";
+        }
+        else {
```
Daaaav:
```suggestion
        else
        {
```

Daaaav (approved):
Overall +1! The only comments I have are for brace style consistency - we _usually_ do opening braces on a separate line, except around code that already doesn't, like the simplified scripting parser. (That's probably the code that's most in its original state since the source went public!)

Ratismal:
Thanks @Daaaav, resolved and rebased these style fixes.

PR #1277, opened 2026-05-07
Fix double script execution when entering playtesting

NyakoFox:
## Changes:

`twoframedelayfix()` now gets called in `resetplayer` due to #1258, to fix the two-frame script delay when you respawn. Unfortunately, there seems to be some strange bug where starting playtesting in a script box activates the script twice. While I'm not exactly sure WHY this happens logic-wise, we can quickly squash this issue by checking if the player is reset due to dying.

## Legal Stuff:


Daaaav (approved):
Haven't tested, but this seems simple and safe enough, because it only makes the line #1258 added conditional. So worst case, if this happens to be broken, we go back to the old behavior. I assume you also tested it before submitting so that's good enough for me!

PR #1278, opened 2026-05-14
Fix two nuisance error messages

Daaaav:
## Changes:

This PR gets rid of two messages I've seen a lot, one at runtime and one at compile time:
- `[ERROR] Unable to create Vorbis handle, error 30` if any music slots are empty (which is very common in custom levels)
- A warning about `for (int i` which should've been `for (size_t i`

## Legal Stuff:

### DIFF
--- a/desktop_version/src/Music.cpp
+++ b/desktop_version/src/Music.cpp
@@ -381,6 +381,12 @@ public:
     MusicTrack(SDL_RWops *rw, const char* _id, bool _loose_extra)
     {
         SDL_zerop(this);
+        if (rw->size(rw) <= 1)
+        {
+            // Don't bother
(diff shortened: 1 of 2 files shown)


NyakoFox:
For the former, I feel like it'd be good to do a debug print instead (so people with verbose debugging on can clearly see what's happening) but maybe that's not important.

And yeah, good idea; these ones bother me too.

Daaaav:
> For the former, I feel like it'd be good to do a debug print instead (so people with verbose debugging on can clearly see what's happening) but maybe that's not important.

Oh, that's a good idea! I'll do that

flibitijibibo (approved):

PR #1281, opened 2026-05-18
Add WINXP option to CMakeLists.txt

NyakoFox:
## Changes:

This is a new option (off by default, on for official builds) which tell Physfs to support Windows XP. Without it, it tries to use a function from Vista (existed in XP, but was an external thing).

For making a build that actually supports Windows XP, `cmake` must be invoked with `-T v141_xp`, and the toolchain must be installed in Visual Studio.

This PR was made as builds will use newer versions of Visual Studio moving forwards; but I don't think we want to give up XP yet.

Tested on VS2026 and a Windows XP machine that @Daaaav owns.

## Legal Stuff:

### DIFF
--- a/desktop_version/CMakeLists.txt
+++ b/desktop_version/CMakeLists.txt
@@ -16,9 +16,14 @@ option(OFFICIAL_BUILD "Compile an official build of the game" OFF)
 
 option(MAKEANDPLAY "Compile a version of the game without the main campaign (provided for convenience; consider modifying MakeAndPlay.h instead" OFF)
 
-if(OFFICIAL_BUILD AND NOT MAKEANDPLAY)
-    set(STEAM ON)
-    set(GOG ON)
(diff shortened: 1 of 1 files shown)


mothbeanie:
Our immense windows xp playerbase must be dancing merrily in circles like the seven dwarfs while reading this! 🤭 
(This is definitely the best of both worlds, and much better than compiling on VS2013!! Especially since this game is so old)

NyakoFox:
It looks like VS2026's installer has removed the option to install `v141_xp` (due to them adding signing to the installer for new packages or something?) so XP builds will have to be done on VS 2022 (unless an archive of the old installer exists anywhere, I guess...)

Either way, I still think it'd be good if official builds used this (until we finally want to formally drop Windows XP support.)

flibitijibibo:
I should be able to stay on 2022 for a while, so let's get this in before I forget to do it for 2.5...

PR #1283, opened 2026-08-07
SDL3 update

croscato:
## Changes:

This is my attempt at converting the source code to use SDL3, modifying the bare minimum while avoiding changes to the game logic.

Using the guide provided by SDL developers in https://wiki.libsdl.org/SDL3/README-migration, the migration process was more approachable.

My approach was:
* Change CMakeLists to use SDL3
* Use the script `rename_symbols.py` to convert names from SDL2 to SDL3 (provided in the guide)
* Fix header inclusion
* Fix compilation errors
* Fix logic errors (SDL3 return `bool` instead of `int` for most of the functions)
* Fix some narrowing warnings (from using `FRect`)

## Disclaimer:

No LLM was used at any moment in this conversion. Apart from the script `rename_symbols.py` provided in the migration guide, no automation of any kind was used.

## Legal Stuff:


NyakoFox:
I think originally we were going to wait for my touch input/mobile support, but that has stalled out and is awaiting a(nother) rewrite when I have the time to start it, so I personally have no problem with this.

flibitijibibo:
Will try to review over the weekend since I've done a bunch of updates to 3.0 at this point - one thing that'll help with initial smoke testing is fixing CI, hopefully it's just a matter of updating the package lists for each builder.

desktop_version/src/CustomLevels.cpp
```diff
@@ -1762,7 +1762,10 @@ void customlevelclass::generatecustomminimap(void)
                     if (tile >= 1)
                     {
                         // Add this pixel
-                        SDL_Point point = { ((i2 - data.startx) * 12 * data.zoom) + i, ((j2 - data.starty) * 9 * data.zoom) + j };
+                        SDL_FPoint point = {
```
flibitijibibo, line 1765: Marking as a point of interest for mapmakers since this is math change more than an SDL ABI thing

desktop_version/src/KeyPoll.cpp
```diff
@@ -317,37 +310,38 @@ void KeyPoll::Poll(void)
...
-            imebuffer_length = evt.editExt.length;
-            SDL_free(evt.editExt.text);
-            break;
+        // TODO(croscato): investigate why this is not defined in SDL3
+        // case SDL_EVENT_TEXT_EDITING_EXT:
+        //     imebuffer = evt.editExt.text;
+        //     imebuffer_start = evt.editExt.start;
+        //     imebuffer_length = evt.editExt.length;
+        //     SDL_free(evt.editExt.text);
+        //     break;
```
Daaaav: The `_EXT` version of this event was introduced at some point in SDL2 because the normal one was limited to 32 bytes and couldn't be changed for compatibility. It looks like SDL3 simply made `SDL_EVENT_TEXT_EDITING` have the behavior of  the old `SDL_TEXTEDITING_EXT`/`SDL_EVENT_TEXT_EDITING_EXT`, minus the need to free the text.

(In researching this I found documentation saying we need to `SDL_CleanupEvent()` after this and some others, but that was actually outdated information - that now happens automatically. I was wondering why that function was documented so poorly when it would be such a pitfall lol)

flibitijibibo (commented):
Everything looks okay at first glance - the changes from Rect to FRect should be tested first. Once that's been tried out and CI is happy again we should be good to go here.

desktop_version/src/GraphicsUtil.cpp
```diff
@@ -383,12 +386,36 @@ bool UpscaleScreenshot2x(SDL_Surface* src, SDL_Surface** dest)
...
+{
+    SDL_Rect result;
+
+    result.x = static_cast<int>(rect.x);
+    result.y = static_cast<int>(rect.y);
+    result.w = static_cast<int>(rect.w);
+    result.h = static_cast<int>(rect.h);
+
+    return result;
+}
```
Daaaav: Should these be defined in our code? They're documented SDL functions so I'd expect them to conflict.

desktop_version/src/KeyPoll.cpp
```diff
@@ -389,172 +383,166 @@ void KeyPoll::Poll(void)
...
-                SDL_EnableScreenSaver();
-                break;
+            if (SDL_strcmp(SDL_GetCurrentVideoDriver(), "x11") == 0)
+            {
+                wasFullscreen = !gameScreen.isWindowed;
+                gameScreen.isWindowed = true;
+                SDL_SetWindowFullscreen(
+                    SDL_GetWindowFromID(evt.window.windowID),
+                    0
+                );
```
Daaaav, line 505: This looks like a forgotten (but minor) change:

```suggestion
                SDL_SetWindowFullscreen(
                    SDL_GetWindowFromID(evt.window.windowID),
                    false
                );
```

Daaaav (commented):
First off, thanks for taking on this task!

I did a look through this as well and I only found a few minor things. I didn't check if there's a forgotten `!= 0` -> `== 0` anywhere, but the diffs make sense to me!

Other than the code comments: Apparently macros like `__APPLE__` and `__HAIKU__` and such should be renamed: https://wiki.libsdl.org/SDL3/README-migration#sdl_platformh. (I thought those were defined by the OS/compiler/etc rather than SDL, but there's probably a reason for it)

desktop_version/src/SteamNetwork.c
Daaaav, line 156: Just looked at the CI errors, and both Linux and macOS error on the line with `->BytesPerPixel` here.

`surface2x->format` is an `SDL_PixelFormat`, in SDL2 that's a struct with `BytesPerPixel` member, but in SDL3 it's an enum and needs `SDL_GetPixelFormatDetails` called on it.

(Hopefully the suggestion feature works, it's looking like it only changes the first line on my end)

```suggestion
    const SDL_PixelFormatDetails* details = SDL_GetPixelFormatDetails(surface2x->format);
    if (details == NULL)
    {
        return;
    }

    SteamAPI_ISteamScreenshots_WriteScreenshot(
        steamScreenshots,
        surface2x->pixels,
        surface2x->w * surface2x->h * details->bytes_per_pixel,
        surface2x->w,
        surface2x->h
    );
```

Not sure what's going on with the Windows CI.

desktop_version/src/KeyPoll.cpp
croscato: I had a suspicion of that being the case but couldn't find any documentation to confirm, anyway went ahead and removed the commented code.

Indeed the documentation around these events are rough.

desktop_version/src/GraphicsUtil.cpp
croscato: Didn't notice that SDL provided this function and implemented as I needed.

Makes perfect sense to use the ones provided by the library, although it don't have a `SDL_RectToFRect`. However it's only used in one place and a function for it is not justified.

Removed my conversion implementation in favor of the one provided in the library.

It didn't conflict by mere lucky, SDL function expect two pointer (SDL_Rect *, SDL_FRect *) mine only one (SDL_Rect &), it was treated as an overload.

desktop_version/src/KeyPoll.cpp
croscato: Good catch! Fixed.

desktop_version/src/SteamNetwork.c
croscato: Applied you suggestion and it fixed the compilation error for this file.

It seems that I missed the `OFFICIAL_BUILD` parameter, which would have include this file in the compilation process.

croscato: About the platform macros I'm not so sure it should be changed.

You're right `__APPLE__` and `__HAIKU__` are compiler defined macros, SDL "looks for" common compiler definitions and aggregate them in a SDL_PLATFORM_XXXX macro.

Excerpt from `SDL_platform_defines.h`:
```cpp
#if defined(__unix__) || defined(__unix) || defined(unix)
...
#define SDL_PLATFORM_UNIX 1
#endif

#ifdef __APPLE__
...
#define SDL_PLATFORM_APPLE 1
```

NyakoFox: Personally I think it'd be worth it to change; this abstraction seems to be done to make sure the definitions do exist no matter what compiler you're using.

croscato: Did the changes, only `__DragonFly__` doesn't have an SDL enum. So I kept as it was.

croscato: Compilation for Windows works using `Visual Studio 2022` if

- SDL3_DIR point to a development release of SDL3
- Removing `WIN32` from `CMakeLists.txt`, otherwise it will try to link with `WinMain`.

flibitijibibo: Ah, good catch on WinMain: We just need to add an explicit include to SDL3/SDL_main.h for that to be fixed.

flibitijibibo:
Next push should rebase against latest master, forgot to update the VS version in CI:

https://github.com/TerryCavanagh/VVVVVV/commit/dc2504e560ece1161bf3fd4c19187db5f8c84dda

desktop_version/src/KeyPoll.cpp
flibitijibibo: Man, what a mess... if one of you wants to file a ticket for this at SDL's tracker...

https://github.com/libsdl-org/SDL/issues

... CC me and I'll make sure it's in an upcoming milestone.

desktop_version/src/SteamNetwork.c
croscato: Reintroduced the `WIN32` option to `add_executable` in _CMakeLists.txt_ and conditionally including the `SDL_Main.h` header for the Windows platform solved the compilation problem.

I noticed that in `src/SDL_uikit_main.c` there is something similar for the IOS platform and made the changed according to the SDL3 migration guide. But I don't have the means to compile/test it.

flibitijibibo:
Threw together an untested diff that _should_ get Windows back on track:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index c87f612a..f719d5dd 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,7 +121,7 @@ jobs:
     runs-on: windows-latest
 
     env:
-      SDL_VERSION: 2.26.0
+      SDL_VERSION: 3.4.0
 
     steps:
     - uses: actions/checkout@v4
@@ -134,23 +134,23 @@ jobs:
       env:
         cache-name: cache-sdl
       with:
-        path: C:\SDL2-*
+        path: C:\SDL3-*
         key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ env.SDL_VERSION }}
 
     - if: ${{ steps.cache-windows-sdl.outputs.cache-hit != 'true' }}
       name: Download SDL if not cached
       run: |
-        Invoke-WebRequest "https://github.com/libsdl-org/SDL/releases/download/release-$env:SDL_VERSION/SDL2-devel-$env:SDL_VERSION-VC.zip" -OutFile C:\SDL.zip
+        Invoke-WebRequest "https://github.com/libsdl-org/SDL/releases/download/release-$env:SDL_VERSION/SDL3-devel-$env:SDL_VERSION-VC.zip" -OutFile C:\SDL.zip
         Expand-Archive C:\SDL.zip -DestinationPath C:\
 
     - name: CMake initial configure/generate
       run: |
         mkdir $env:SRC_DIR_PATH/build
         cd $env:SRC_DIR_PATH/build
-        $env:LDFLAGS =  "/LIBPATH:C:\SDL2-$env:SDL_VERSION\lib\x86 "
+        $env:LDFLAGS =  "/LIBPATH:C:\SDL3-$env:SDL_VERSION\lib\x86 "
         cmake -G "Visual Studio 18 2026" -A Win32 `
-              -DSDL2_INCLUDE_DIRS="C:\SDL2-$env:SDL_VERSION\include" `
-              -DSDL2_LIBRARIES="SDL2;SDL2main" ..
+              -DSDL3_INCLUDE_DIRS="C:\SDL3-$env:SDL_VERSION\include" `
+              -DSDL3_LIBRARIES="SDL3" ..
 
     - name: CMake configure (default version)
       run: |
```

desktop_version/src/KeyPoll.cpp
croscato: Before opening the issue, I tried to find how this really changed.

* The stated change in the migration guide, renaming _SDL_TEXTEDITING_EXT_ to _SDL_EVENT_TEXT_EDITING_EXT_ occurred in [7b50bae5](https://github.com/libsdl-org/SDL/commit/7b50bae52456430f199a89c99535d65dd9c297ea/src/events/SDL_keyboard.c) ([src/events/SDL_Keyboard.c:1062](https://github.com/libsdl-org/SDL/blob/7b50bae52456430f199a89c99535d65dd9c297ea/src/events/SDL_keyboard.c#L1062)).
Memory for the event text is always allocated and should be freed by the user.

* At [7e445da5](https://github.com/libsdl-org/SDL/commit/7e445da56990ce741e356ff7ed354abea129f288/src/events/SDL_keyboard.c) _SDL_EVENT_TEXT_EDITING_EXT_ functionality was combined with _SDL_TEXTEDITING_EXT_ ([src/events/SDL_keyboard.c:1108](https://github.com/libsdl-org/SDL/blob/7e445da56990ce741e356ff7ed354abea129f288/src/events/SDL_keyboard.c#L1108)).
Memory for the event text is allocated if the size is greater than the size of `event.edit.short_text` field. Also it's needed to call `SDL_CleanupEvent()` after receiving the event.

* At [70c149c8](https://github.com/libsdl-org/SDL/commit/70c149c88f05e469f07b7e5fff58b7e16e4d1bea/src/events/SDL_keyboard.c) ([src/events/SDL_keyboard.c:1107](https://github.com/libsdl-org/SDL/blob/70c149c88f05e469f07b7e5fff58b7e16e4d1bea/src/events/SDL_keyboard.c#L1107)) text for the event is always allocated using 
`SDL_AllocateEventMemory` and freed automatically by SDL.

* This behavior is further refined in [306c4164](https://github.com/libsdl-org/SDL/commit/306c4164bc5e9c8976ad54c747885fbc645d9084/src/events/SDL_keyboard.c) ([src/events/SDL_keyboard.c:1240](https://github.com/libsdl-org/SDL/blob/306c4164bc5e9c8976ad54c747885fbc645d9084/src/events/SDL_keyboard.c#L1240)) and [bb96320c](https://github.com/libsdl-org/SDL/commit/bb96320cc4380665180a146f3ba8cdf52554b998/src/events/SDL_keyboard.c) ([src/events/SDL_keyboard.c:754](https://github.com/libsdl-org/SDL/blob/bb96320cc4380665180a146f3ba8cdf52554b998/src/events/SDL_keyboard.c#L754)).

Documentation wise the only thing missing is the fact the _SDL_EVENT_TEXT_EDITING_EXT_ became _SDL_EVENT_TEXT_EDITING_ in the migration guide.

Everything works as @Daaaav mentioned, so the removal of the commented code is right thing to do.

croscato:
@flibitijibibo, pushed a new commit with the ci.yml patch.

Running the build commands present in the patch, it build correctly on my Windows machine. It should indeed fix the CI problems.

flibitijibibo:
Looks like it's just Android left, but that's for someone else to figure out... in the meantime, going to try and test this tomorrow.

flibitijibibo:
Basic OS-level functionality is okay, but it looks like the render targets used for screen transitions broke as part of the update - you can test this by opening the map in-game, the map should move in and out but in the current revision it just pops in at the end of the animation. Could be a messed up blend mode or target binding?

croscato:
> Basic OS-level functionality is okay, but it looks like the render targets used for screen transitions broke as part of the update - you can test this by opening the map in-game, the map should move in and out but in the current revision it just pops in at the end of the animation. Could be a messed up blend mode or target binding?

Had the same thought when I saw the bug, but turns out it was a logic problem.

Functions that in SDL2 returned integer errors, now returns `bool`. In my first update attempt I've changed all functions in `Graphics.h` to return `bool`, matching SDL3 results, but later decided to keep as `int` and return the appropriate value.

The problems with the map animation was in [src/Graphics.cpp:3185](https://github.com/TerryCavanagh/VVVVVV/blob/75ceaec54ea8664736d08146ec2f1ce1b84afec7/desktop_version/src/Graphics.cpp#L3185) where the result of `copy_texture` was being checked incorrectly. This is the only place where the result is checked, everywhere else it's just the plain function call.

Maybe, changing the `Graphics.h` functions signatures to match SDL results is not a bad idea?

NyakoFox:
Personally, I think using bools would be best; I originally kept return values as `int`s when porting to the renderer system so I could focus on getting it functional, but it'd be cleaner and easier to understand if things were actually booleans... and so things like this bug wouldn't happen :P

croscato:
Made the changes to the functions in `graphics` to match the return types from SDL.

Looking through the code, I found some places where SDL result where checked as `int` and fixed it too.

flibitijibibo:
All sounds good, will do another test round today!

flibitijibibo:
Initial test came out okay, if I don't find anything immediately obvious then I'll likely merge this tomorrow... is anyone available to look at the Android task? I'd like to get this up on 2.5-testing for people to try out over the weekend, so if we need to temporarily disable that task to give people more time I can do that post-merge.

flibitijibibo:
Not sure how I missed this the first round, but it looks like the big font is using the wrong texture filter - should be an easy fix but wanted to point it out:

image

image

desktop_version/src/Screen.cpp
```diff
@@ -63,15 +61,23 @@ void Screen::init(const struct ScreenSettings* settings)
...
+    m_renderer = SDL_CreateRenderer(m_window, NULL);
 
     if (m_renderer == NULL)
     {
         vlog_error("Could not create renderer: %s", SDL_GetError());
         VVV_exit(1);
     }
 
-    SDL_RenderSetVSync(m_renderer, (int) vsync);
+    SDL_SetDefaultTextureScaleMode(m_renderer, SDL_SCALEMODE_LINEAR);
```
flibitijibibo: I think this is it - I think the only time we need a linear filter is when it's enabled for the backbuffer surface, while the rest are nearest.

flibitijibibo: Yeah, this was all it needed AFAICT

```suggestion
    SDL_SetDefaultTextureScaleMode(m_renderer, SDL_SCALEMODE_NEAREST);
```

croscato: It indeed fixed, pushed the changes.

I tried make the Android CI work, but haven't had any success. Will give it another short.

flibitijibibo:
Will go ahead and merge once CI passes (minus Android, will temporarily disable for now). 2.5-testing on Steam will be  updated soon after.

